PR #1, opened 2015-09-11
Prevent endless loop if request extracting takes more than 30 seconds

woehrl01:
If `getExtractRequestsForNode` is taking longer than 30 seconds (if you have a lot of reloads queued up) you will get an endless loop here, because you will never get into the `for` loop

happens if `initial.load.use.extract.job.enabled=true`

leads to a lot of the following log outputs: (which never process)
```
Batch 1-2820 is not ready for delivery. It is currently scheduled for extraction
```


woehrl01:
sure :+1:

PR #3, opened 2015-09-14
Provide functionality to not require context table on sqlite for threadsafe data capture

woehrl01:
Uses the documented SQlite feature that you are able to override already defined functions with different output/behaviour (which is only visible to the current connection)

Recommended/tested settings:

```
sqlite.trigger.function.to.use=sqlite_version

db.init.sql=PRAGMA busy_timeout = 5000;
```

it's not required to use `jobs.synchronized.enable=true` anymore (if you don't care about `BUSY_TIMEOUT`)

uses the context table if no function name provided (so, no breaking change and opt-in) 

I think this is not working on android, because you need a `JdbcSqlTransaction`

You need to override an existing function, so you aren't required to provide a custom function implementation in the external applications. (which would fail the trigger if you wouldn't do that)


chenson42:
http://www.symmetricds.org/issues/view.php?id=2396

PR #8, opened 2016-01-07
Modified SymmetricLauncher to support running a single route, push, p…

hhowe29:
…ull sequence. Syntax is sym --engine foo --route --push --pull.


erilong:
I commit changes for symadmin command utility for version 3.8.  Here is the issue in the tracker:

http://www.symmetricds.org/issues/view.php?id=2477

Thanks for the feature idea.

PR #11, opened 2016-01-29
mssql trigger: changed from cursor to dml

woehrl01:
This is a bigger change, but it seems inevitable (to me).

I changed the MSSQL trigger to use DML via INSERT + SELECT instead of using a cursor. This has two reasons:
- I wanted to use a column value in my external_select statement, e.g.

```
$(curTriggerValue).$(curColumnPrefix)order_id

translated:

inserted.order_id
```

but I'm not able to get that value, because the cursor doesn't include the column values. Writing the select as 

```
select $(curTriggerValue).$(curColumnPrefix)order_id from $(curTriggerValue)

translated:

select inserted.order_id from inserted
```

even leads to returning multiple values or using just the first result (if you do a range insert/update/delete). 
I'm also not able to join via the pk, because this is also not included. 
- According to multiple sources, cursors are highly discouraged to use inside triggers. So this also has an impact on performance (should be faster).

### DIFF
--- a/symmetric-client/src/main/java/org/jumpmind/symmetric/db/mssql/MsSqlSymmetricDialect.java
+++ b/symmetric-client/src/main/java/org/jumpmind/symmetric/db/mssql/MsSqlSymmetricDialect.java
@@ -369,7 +369,8 @@ public class MsSqlSymmetricDialect extends AbstractSymmetricDialect implements I
     
     @Override
     protected String getDbSpecificDataHasChangedCondition(Trigger trigger) {
-        return "@OldDataRow is null or @DataRow != @OldDataRow";
+    	/* gets filled/replaced by trigger template as it will compare by each column */
+        return "$(anyNonBlobColumnChanged)";
     }
 
 }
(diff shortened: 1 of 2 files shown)


woehrl01:
I just saw that I missed, adjusting `getDbSpecificDataHasChangedCondition`. I'll have a look for that on monday.

woehrl01:
made the modifications. I changed the compare to use the columns, instead of @OldRowData, because now `trigger.update.capture.changed.data.only.enabled=true` also works if `use_capture_old_data` is disabled

woehrl01:
FIY. I just applied the latest changes (eg. adding the correct prefix)

chenson42:
This has been merged into 3.8 for testing

woehrl01:
Great @chenson42! I'm not sure if this is the latest code we use in production since half a year with great success. I'll have a look and let you know. I did not keep that up to date as I did knew it will be merged ever :) after some months. I'll open a new PR if there are additional changes I found useful if you want.

woehrl01:
I just had a look at our code, and we have no other change in our production code. Thanks for merging this in!

PR #12, opened 2016-01-31
Increased size of EXTERNAL_DATA column to varchar(4000)

woehrl01:
I increased the size of the EXTERNAL_DATA column to a bigger size, to allow storing of more informations in it.

The reason for this is that I want to store the target nodes comma separated inside this column (I know that there is the NODE_LIST column, but I can't fill this column inside the trigger.)

Additionally I use the Column Match Router with the expression `EXTERNAL_DATA=NULL OR EXTERNAL_DATA contains :NODE_ID`

The reason I don't want to use the Lookup Router is that in my scenario, there are a lot of different factors which define which nodes the data gets routed to. Which would require multiple different Lookup Routers with a lot of rows for some of them (>100.000).

If I create the correct routing inside EXTERNAL_SELECT, I can properly use the database indicies for reducing the rows in the lookup tables, the routing process is a lot simpler and faster because it doesn't require to load the lookup tables (for each routing process) and I can properly define the different factors via SQL queries, directly at the table trigger level and don't require separate viewes. 

I choosed VARCHAR(4000) to be able to define the routing to about 1000 different nodes, technically doesn't make a huge change storage/database wise and still can be indexed. (instead of LONGVARCHAR)

### DIFF
--- a/symmetric-core/src/main/resources/symmetric-schema.xml
+++ b/symmetric-core/src/main/resources/symmetric-schema.xml
@@ -81,7 +81,7 @@
         <column name="channel_id" type="VARCHAR" size="128"  description="The channel that this data belongs to, such as 'prices'" />
         <column name="transaction_id" type="VARCHAR" size="255"  description="An optional transaction identifier that links multiple data changes together as the same transaction." />
         <column name="source_node_id" type="VARCHAR" size="50"  description="If the data was inserted by a SymmetricDS data loader, then the id of the source node is record so that data is not re-routed back to it." />
-        <column name="external_data" type="VARCHAR" size="50"  description="A field that can be populated by a trigger that uses the EXTERNAL_SELECT" />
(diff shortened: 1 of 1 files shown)


woehrl01:
I thought about this PR for a while. And I think it would be better to have this configurable (as parameter) which applies the size inside [AbstractSymmetricDialect.readSymmetricSchemaFromXml()](https://github.com/JumpMind/symmetric-ds/blob/c439b5536e66057deeb329b9f9ef0ed380623df3/symmetric-core/src/main/java/org/jumpmind/symmetric/db/AbstractSymmetricDialect.java#L486) so that it won't introduce a change for present users.
what you do you think?

erilong:
I agree, it would be best to make a parameter so there are no upgrade issues, but there is an option to increase the size of the field.  To expand on this idea, it would be nice to have a parameter that lets the user configure the CLOBs on sym_data to be varchar with a user-specified size -- if table data fits into the smaller size, there is better performance.  I put in a request for that in the tracker (http://www.symmetricds.org/issues/view.php?id=2655)

PR #13, opened 2016-02-01
Add configuration option to always right trim CHAR fields

woehrl01:
This PR adds, that the already existing parameter `isCharColumnSpaceTrimmed` is used.

It also enables this parameter for Sqlite to trim char fields. 
If you have a MSSQL database with a CHAR(30) column in a table. If it syncs to a SQLite database it gets created as VARCHAR(30). Every data is padded with whitespaces. 
If you have a `=` compare on MSSQL Server it ignores the trailing spaces on SQLite it doesn't, so data is not found on mobile devices as it would on the central database.

Maybe it is better to make this a configurable options, let me know what you think, I'll add that if desired.


mmichalek:
Lukas, do you still need this change?  We have reviewed it and think the best course of action would be to a) remove the isCharColumnSpaceTrimmed() that isn't being used. b) Add a parameter for trimming the whitespace so it isn't so tied to the DB platform.  Feel free to submit an updated PR along those lines and we will review.  Thanks.

woehrl01:
Mark, I thanks for your reply. Yes, I am still relying heavily on this change. I am totally open in modifying it. I would change this as you suggest to a global configuration option. I'll update the PR. 

I'll add a second PR which removes the `isCharColumnSpaceTrimmed`

mmichalek:
Sounds good, thank you Lukas.  Mark

woehrl01:
Mark, I modified the PR, rebased it onto the latest 3.9 and squashed all the commits into a single one.

mmichalek:
Thank for the revised PR Lukas, it has been merged.

woehrl01:
Awesome! :clap:

PR #15, opened 2016-02-02
Enable using of encrypted sqlite databases

woehrl01:
This PR adds support for encrypted SQLite databases.

It addes the `pragma hexkey` to the init sqls. 

Encryption is (currently) not working with the [xerial/sqlite-jdbc](https://github.com/xerial/sqlite-jdbc) driver. 

[But with custom/commercial builds.](https://www.sqlite.org/see/doc/trunk/www/readme.wiki)

Of course you could just add this statement to the init sqls, but you would loose the bonus of the encrypted password feature.


woehrl01:
Moved that code to xerial/sqlite-jdbc#294 where I think this is better kept.

PR #20, opened 2016-03-16
quote character for strings in where clause should be single quote

ghost:


ghost:
@chenson42 
Any chance you can apply this change... it is a one-liner

PR #26, opened 2016-05-11
Add unique key violation detection for PostgreSQL

jdahlin:
Use savepoints to support unique key violtions for PostgreSQL/Redshift.

### DIFF
--- a/symmetric-db/src/main/java/org/jumpmind/db/platform/AbstractDdlBuilder.java
+++ b/symmetric-db/src/main/java/org/jumpmind/db/platform/AbstractDdlBuilder.java
@@ -2699,4 +2699,11 @@ public abstract class AbstractDdlBuilder implements IDdlBuilder {
         return caseSensitive;
     }
 
+    public String defineSavepoint(String savepointName) {
+        throw new DdlException("Platform does not support SQL savepoints.");
+    }
+
+    public String rollbackToSavepoint(String savepointName) {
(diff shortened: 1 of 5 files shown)


erilong:
Why do you want to use savepoints?  SymmetricDS did use savepoints like this in the past, and it was changed to a method of an insert statement that uses a select to make sure the row doesn't exist -- if you look in the PostgreSqlDmlStatement, you'll see how it builds inserts.  If the rows affected for the insert is 0, then it knows there was a conflict.  I heard there were performance problems with a savepoint on each row (see http://www.symmetricds.org/issues/view.php?id=523), but I don't have first hand experience with it.

If we added savepoints back in, we would want it as a parameter that can be toggled (see ParameterConstants and symmetric-default.properties), and default it to the current method until we could assess the performance.  This should really be a change on the 3.8 development branch (3.7 is the stable branch).  You only need savepoints on an insert, because that's where the database can raise an exception, which rolls back the row.  Updates and deletes don't need savepoints because they will just return 0 rows affected.  The PostgreSqlDmlStatement currently does the insert with select, which is overhead you don't need if you use savepoints, so you'd want to check the DatabaseInfo when the PostgreSqlDmlStatement is created and set a boolean internally on whether or not you need the special insert.  Setting the savepoint and rolling back to the savepoint should be on the ISymmetricDialect and PostgreSqlSymmetricDialect instead of the IDdlBuilder and PostgreSqlDdlBuilder (the ddl builders are supposed to be for creating tables).

PR #30, opened 2016-06-07
Clear NodeService cache if data changed

woehrl01:
This changed adds clearing of the `NodeService` cache if it detects a change via the `ConfigurationChangedDataRouter`.

This is required so that the server is able to register nodes if they are added via external statements. 

Without this change `nodeService.findNode` would return `null` until you restart the server.


woehrl01:
just saw that the cache has already been fully removed in the latest version in 3.7

PR #31, opened 2016-06-07
Load node by id and sql instead of client side filtering

woehrl01:
Changset 04a5761721 is not complete reverting the previous changes from 2bb50bfb2. This pull request changes `findNode` back to load the node directly via sql and not loading any available nodes into a map.

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/NodeService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/NodeService.java
@@ -127,7 +127,9 @@ public class NodeService extends AbstractService implements INodeService {
      * with it.
      */
     public Node findNode(String id) {
-        return findAllNodesAsMap().get(id);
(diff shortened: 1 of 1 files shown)


woehrl01:
Closing, as fixed in 3.7 via http://www.symmetricds.org/issues/view.php?id=2742 still relevant for 3.8 but I think this will be merged soon

chenson42:
The node list is cached in 3.8.  Do you see a problem with the cache implementation?

woehrl01:
I just had a look. In 3.8 it's still using findAllNodesAsMap which in return loads all nodes from database. The node list is cached but this cache is not used here.

woehrl01:
I'm talking about findNode without the boolean flag for cache usage. And findNode in case caching should not be used. I'm both cases all nodes are loaded.

chenson42:
Got it.  Thanks for the explanation.

woehrl01:
You're welcome!

PR #32, opened 2016-06-20
Allow to execute update only statements via restservice

woehrl01:
This change adds a parameter to the rest api to allow to execute non select statements.

It returns the number of applied changes via the already existing model.


chenson42:
http://www.symmetricds.org/issues/view.php?id=2654

PR #36, opened 2016-07-13
Improve routing performance if there are pending initial loads

woehrl01:
This PR reduces heavy database calls if there are nodes which have pending initial loads.

It replaces the call to  "load any active trigger_histories and compare the result > 0" with a direct call of count(*) on the database side and compare that result.

This is a partial fix to reduce: http://www.symmetricds.org/issues/view.php?id=2669


woehrl01:
has been superseded by dc0f364

PR #39, opened 2016-08-25
Delete staged resources if batch manually set to "NE" to recreate batch

woehrl01:
This PR leads to recreating the staged resources if a batch is manually set to "NE". This is required as according to the documentation you have to wait either until the resource times out or you have to clear the staged area, if you manually adjust the batch content on conflict or error.

> After modifying the batch you will have to clear the Staging Area manually or wait for the staged version of the batch to timeout and clear itself.

With this change you can fully control the resolving scenario via sql, without having to additionally log into the application server and manually delete the file or even have to restart the service.

Additionally it ensures that any new batch is really extracted and cannot be hijacked with a random file.

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
@@ -779,6 +779,13 @@ public class DataExtractorService extends AbstractService implements IDataExtrac
             long byteCount = 0l;
             long transformTimeInMs = 0l;
 
+            if (currentBatch.getStatus() == Status.NE) {
+                IStagedResource resource = getStagedResource(currentBatch);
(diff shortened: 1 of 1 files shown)


mmichalek (commented):
These changes have been committed and added to the issue tracker as: 0003275: Allow users to reset batch status to NE to force re-extraction of a batch.  Thank you for your PR!!!

chenson42:
This causes batch corruption issues for common batches.  This needs to be backed out for 3.8 and maybe a new flag can be added to force re-extraction for 3.9.

woehrl01:
If you are happy with a new flag/column, I'll create a new PR. Sorry for any inconvenience!

chenson42:
That would be great.  Can you do it on the 3.9 branch?  We are going to release that in the next several weeks.

woehrl01:
Absolutely! Will do tomorrow! :+1:

PR #42, opened 2016-09-19
Added option for immediate re-pull if data found

woehrl01:
This PR addes a parameter to do an immediate re-pull if the client got data. This is a useful option to speed up dataloading if you have a lot of unsynced big batches, which gets cutoff via `transport.max.bytes.to.sync`. If this is set to 1MB (default) and you have about 200 MB to sync, and the pull job runs every 10 seconds (default), it needs at least 30 Minutes to transfer all data.

Of course you can increate the max bytes to sync to a higher value, or increase the pull job frequency, but these parameters have other trade-offs, you need to consider.

I changed the log output a little bit so that you only get the "Immediate pull request" if you already ran a second cycle, to not clutter the log. Maybe this line should be changed to `log.debug`?


erilong:
Merged into 3.9 branch.  Thanks!

PR #44, opened 2016-09-27
Prevent reading out database password via rest api

woehrl01:
This PR prevents reading the database password via accessing the rest api. It simply returns an empty string instead.


erilong:
Merged into 3.9 branch.  Thanks!

PR #46, opened 2017-03-16
Update InterbaseTriggerTemplate.java

giltonic:
Now the Fieldis "name" not "id"

https://www.symmetricds.org/issues/view.php?id=2766


ILPlais (approved):
Triggers creation don't work without this changes.

chenson42:
We are working on fixing this.  Thanks for the patches!

chenson42:
This is fixed by this issue:  https://www.symmetricds.org/issues/view.php?id=2766

PR #49, opened 2017-05-09
Partial table reload, to not purge incoming and fix NPE

woehrl01:
This PR fixes two things:

1. If you make a partial initial load it purges the incoming events. In my opinion this is mostly not desired, as we could loose events from other tables, which are currently floating in.

2. If you create an entry in `TABLE_RELOAD_REQUEST` with an invalid `TRIGGER`/`ROUTER` combination, it failes with an NPE. This change adds a more usefull log message.


woehrl01:
Fixed by 9bb3123a

PR #51, opened 2017-10-12
Move loading of trigger_routers to more specific execution branch to reduce unneeded db calls

woehrl01:
This is a more superior fix to the problem discussed in #37, without needed to filter the nodes . This moves the call to ```getAllTriggerRoutersForReloadForCurrentNode``` to the execution branch where it is needed. This mostly reduces calls the database where it loads all the triggers and trigger_routers on any routing pass (e.g. every 5 seconds).


chenson42:
Checked in a fix so the commit is tied to an issue number.

0003285: Move loading of trigger_routers to more specific execution branch to reduce unneeded db calls

woehrl01:
@chenson42 whats the best way for further PRs? Should I add an issue at the symmetricds tracker and provide the issue number in the PR title?

chenson42:
The issue number really should be part of the commit.   Open an issue and commit your changes using the issue number.  We'll just accept the pull after a review in that case.  Thanks for all the submissions, btw!

woehrl01:
Perfect! Will do this with the next ones 👍

PR #53, opened 2017-11-13
0003310: Refactored auto.sync.config.at.startup and auto.sync.config.after.upgrade to be independent

woehrl01:
Currently `auto.sync.config.at.startup` only works in conjunction with `auto.sync.config.after.upgrade`. 

And there is no way to force an update of the configuration without requiring a version upgrade.

This PR changes that behaviour so that `auto.sync.config.at.startup` always syncs the config at startup, While `auto.sync.config.after.upgrade` only syncs the config after an upgrade. So both parameter do now what they are named as.


woehrl01:
Close as there has been no interest in the last 4 years

PR #54, opened 2017-11-13
0003311: MSSQL select varbinary from insert/deleted instead of origin table

woehrl01:
This PR changes the trigger so that on MSSQL the value of blob fields are selected from the inserted/deleted virtual table instead of selecting it from the origin table.

This only works if it is a `varbinary(max)` field, so we filter based on that. 

This reduces the access to the origin table and thus improves the trigger performance.


woehrl01:
Close as there has been no interest in the last 4 years

PR #57, opened 2017-11-16
0003275: Allow users to reset batch status to NE to force re-extraction of a batch

woehrl01:
This is a follow up PR after the discussion in #39 .

This adds a new column `reextract_flag` to `outgoing_batch` which forces a reextraction.


mmichalek:
Lukas, this PR in its current form looks very good.  The team though wants to hold off on non-critical schema changes for the 3.9.x series.  So when we kick off 3.10.x development I would like to rebase this PR and include it there.

woehrl01:
Mark, that's absolutely reasonable. Thanks for the heads up!

woehrl01:
Close as there has been no interest in the last 3 years

PR #58, opened 2017-11-24
Fix user-guide: node registration

albihasani94:
add the following command to the node registration section of the User Guide:
- open-registration

The open-registration keyword was missing from the command for registering a node:

image

The actual command in SymmetricDS:

```
bin/symadmin --engine <engine-name> open-registration <node group> <external id>
```


albihasani94:
Run the command as it is in the user guide(with the missing keyword);

image

Run the command with the added **open-registration** keyword:

image

albihasani94:
The actual command in the User Guide:

image

PR #59, opened 2017-11-27
0003322: Improve reload performance by caching rarely changing tables

woehrl01:
Some tables can be huge, with lots of binary columns and they change rarely. The fetching of the data on a reload could lead to a big stress on the db server if it has to convert all the binary fields. 

The following PR adds a feature to allow to define on the `trigger_router` level (new column `cache_reload`). 
If tables should be cached (`cache_reload=1`) they are cached in the staging area. The first time it fetches the table, it loads it from the database and writes all the data at the same time to a new staging area. The second time it fetches the data purley from that cache.

If new data flows through the data router it clears those caches, for the specific table.

We reduced our initial loads for a huge table, from 11 minutes to <1 minute and overall reducing the pressure on the db server.


svernyi:
@woehrl01 

So I'm actually trying this out right now. One thing I noticed is that it doesn't respect the MAX_BATCH_SIZE specified in sym_channel -- instead it just creates one particularly large file. Any chance you could break it up into smaller files akin to how symmetric normally does it?

svengeance:
@woehrl01 

So I'm actually trying this out right now. One thing I noticed is that it doesn't respect the MAX_BATCH_SIZE specified in sym_channel -- instead it just creates one particularly large file. Any chance you could break it up into smaller files akin to how symmetric normally does it? If this is unnecessary for some reason I don't understand let me know as well -- I'm always trying to learn how Sym works.

woehrl01:
@vernyis : thank you for trying this out. The issue with the batch size is interesting, as this change hooks into the reading from the database which should not have an effect on the batch distribution. Which version are you using?

woehrl01:
I see what you mean. the file is the caching file. Which use case do you see in splitting this cache file?

svengeance:
I'm on 3.8.33, for reference.

From what I understand it could make the process a bit more stable. I don't know how...'reliable', symmetric is in a sense, but I have about 14 large (>1mil row) tables that are capable of producing 500Mb files, and they take quite a bit to extract. We're still in the preparation stage, but I will eventually have 200+ clients that need this load. I'm worried about the high concurrent reading of one large file. Writing should not be a problem for me personally as I will be pre-extracting these caches, but I am unsure how the cachewriter would handle multiple clients asking for the same unfinished cache.

woehrl01:
We have such big tables too. What happens if you have unfinished caches the second reader still reads from the database but won't populate the cache. If the first one finishes all following reads are served from the cache until a change is detected via the router.

svengeance:
Well that sounds like it'll be just fine then. I assume you haven't ran into any concurrency issues?

I have an easy test setup using Docker for running any number of clients. I'll report back to you if I hit anything.

Cheers mate, danke schön!

woehrl01:
I hadn't yet any issues with this. I'm looking forward to your feedback! You're welcome!

svengeance:
Very surprised and saddened to have run into a similar issue I've posted about here in the forums: https://sourceforge.net/p/symmetricds/discussion/739236/thread/46a2ff05/

The difference is that the batch that is stuck in RQ is a batch from the cache'd reload table. It's not like there's a missing/deleted file..ahh. This isn't an issue with this feature though. 

It only happened to one of 10 initial loads that all ran simultaneously.

woehrl01:
After having a look at the provided link, I think this is a non-related bug, which you encounter without this change, too. My change doesn't change the logic of the extract/acknowledge logic it simply writes/reads a cache file instead of querying the database.

svengeance:
I thought you might appreciate [these stats](https://shifty.site/uploads/images/2018-01-29_19-09.png). I was able to run 20 simultaneous initial loads over about 7.5 hours, where I was bottlenecked mostly by the network speed of the server. Thanks so much for this PR, it has helped immensely. Now onto sqlite encryption..

abrougher:
The SymmetricDS team will be looking at ways to improve the reload performance early 2018. We definitely want to address this issue. Thanks for the contribution.

erilong:
I think multiple initial loads is a good problem to solve.  You put cache_reload on the sym_trigger_router.  Is there any reason to specify at the table level or would a parameter setting that affects all tables work as well?

woehrl01:
Thanks for your feedback @abrougher and @erilong.

The reason why I put cache_reload on the table level instead of a single parameter. Is that some table could be/are high volume tables, where the cache would mostly be invalidated instead of reused. Due to my implementation where the router invalides the cache as soon a change is detected on such table.

Additional there are some tables which are unique per node so that a cache isn't of a great use either.  

Last if you have a lot of data, sometimes there are only a small portion of the tables which highly affect the initial load speed, so if you can specify only those tables you'll end up with a good trade off between space and performance. 

What would be your suggestion? I'm looking forward to your ideas.

woehrl01:
@vernyis I pushed some fixes I encountered in the last days. 

Thanks a lot for your stats, interesting to see that they match with my experience, that after this change the reload is mostly limitied by the bandwidth.

woehrl01:
Hi @erilong . Was my explanation helpful to you? Thanks!

woehrl01:
Close as there has been no interest in the last 3 years

PR #60, opened 2017-12-11
The SymmetricDS Dashbord Does not show properly

unotech001:
Dear Sir,

We set up master to master profile replication for MS-SQL to Postgresql replication

The replicataion is done properly

Now we added new table to replicate but at master node Dashbord the "Manage","Configure" and "Explorer" tab 
not show.(We try with logout from node and login again but it not show)
symmds dashboard issue

Please help us

refer snap shot for details


joshahicks:
It appears you are looking to post a question instead of a pull request.   You could use the following link to post your question.

https://sourceforge.net/p/symmetricds/discussion/

PR #65, opened 2018-01-11
0003363: Do not extract files if unrelated table reload request is processed

woehrl01:
If you insert a table reload request and have a file sync enabled, these files get an extraction, too.

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
@@ -497,7 +497,7 @@ public class DataService extends AbstractService implements IDataService {
                                 mapReloadRequests, isFullLoad, symNodeSecurityReloadChannel);
 
                         insertFileSyncBatchForReload(targetNode, loadId, createBy, transactional,
-                                transaction, mapReloadRequests, processInfo);
+                                transaction, mapReloadRequests, isFullLoad, processInfo);
                                 
                         if (isFullLoad) {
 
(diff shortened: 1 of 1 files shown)


mmichalek:
Hi Lukas, 

We have reviewed this issue and PR.  2 things I wanted to ask about:
1) elong had a commit to DataService as part of https://www.symmetricds.org/issues/view.php?id=3470. This change causes files to only get extracted when doing a "full" initial load.  Would this change be sufficient for your needs?  (See https://github.com/JumpMind/symmetric-ds/commit/560e6f52dd02d273049a8c9331c7f42cfd5102e9)
2) There are cases when doing a reload in SymmetricDS where the trigger/router combination will be "ALL"/"ALL", indicating a full load ( driven by a row in sym_table_reload_request. )  I'm concerned this PR would break that functionality for file sync although I have not tested that.  What do you think?

Thanks,
Mark

woehrl01:
Hi Mark,

Thanks for your reply! I'd like to give my thoughts on this:

1. I think I missed that change but I don't think this is a good behavior, as I could imagine a scenario (for myself) where it could be useful to just resync files. So I think the change could be a show breaker for this.
2. Your right that this could be an issue, but I haven't tested this yet. As long as I remember correctly, if 'ALL', 'ALL'  is set it is handled like a normal full load.

If you like, I will have a look for both scenarios and update the PR to work with both. But this means that the change of @elong would be slightly changed. 

Best, 
Lukas

mmichalek:
Hi Lukas, thanks for the information.  I was curious if you were doing loads of just files.  I agree that #1 would prohibit the ability to perform a reload for just files. So, if you have a chance to test the ALL/ALL scenario with your approach we would be glad to merge the PR.  

Thanks again for your help.
Mark

woehrl01:
Hi Mark,

I'm currently not doing any file only loads, not because it wouldn't be useful, but because I haven't implemented it yet into our business logic 😅 .

I'll test both scenarios and will update the PR next week.

Best
Lukas

woehrl01:
Hi Mark, I updated and rebased the PR.

mmichalek (approved):

mmichalek:
Thank you Lukas, this has been approved and merged now.

PR #69, opened 2018-03-13
0003487: Executing ISqlTemplate.update() with multiple statements swallows errors

woehrl01:
If you execute a ISQLTemplate.update() with multiple statements at once, and one of the later statements are leading to an error (e.g. foreign key constraint violation) this exception is not thrown back to the user.

This PR reads all the result counts so that if an error is read it will be rethrown. This is also the documented behaviour of JDBC.


mmichalek:
Thank you for that PR - definitely don't want to hide errors.

PR #72, opened 2018-04-04
Supported geography type for MSSQL

lyamin626:
resolved 3327
see the same problem:
https://www.symmetricds.org/issues/view.php?id=3327


mmichalek:
Dmitry, thank you for your interest in the project and for your PR.  I am going to manually merge in your PR with a modified commit message.  Our commit messages typically follow the form of ISSUE_NUM: Description
So in this example it would be: 0003327: Support SQL Server geography type
So if you are able to open an issue with use such a commit message in the future, then I can more likely directly merge your PR.  
Thanks again for taking the same to share your contribution with the community.  Your change should be included in SymmetricDS 3.9.7.

PR #79, opened 2018-05-04
StagingManger.clean should also use createStagedResource

woehrl01:
I stumbled over this changeset b76a12ceb9e9e3297aa3c6f2e6527c9da654ab07 . And I think `StagingManger.clean(long ttlInMs)` should use the `createStagedResource` instead of handling the creation by itself, too.


erilong:
Change was made in changeset b76a12c for issue 3429.

PR #14, opened 2016-02-01
0003313: Propagate default GUID and GETDATE from MSSQL to SQLITE

woehrl01:
Currently it is not possible to sync a default Guid Column to Sqlite. This PR adds that mapping.

PR #80, opened 2018-05-04
0003545: Consider mapped default value on model compare

woehrl01:
A mapped default values lead to always detecting a model change. This PR changes the `ModelComperator` to take the mapped value into account.

### DIFF
--- a/symmetric-db/src/main/java/org/jumpmind/db/alter/ModelComparator.java
+++ b/symmetric-db/src/main/java/org/jumpmind/db/alter/ModelComparator.java
@@ -376,7 +376,7 @@ public class ModelComparator {
                 || (sourceDefaultValue != null && targetDefaultValue == null)
                 || (sourceDefaultValue != null && targetDefaultValue != null &&
                 ((isBigDecimal && ((BigDecimal) sourceDefaultValue).compareTo((BigDecimal) targetDefaultValue) != 0) ||
-                (!isBigDecimal && !sourceDefaultValue.toString().equals(targetDefaultValue.toString()))))) {
+                (!isBigDecimal && !ddlBuilder.mapDefaultValue(targetDefaultValue, desiredTypeCode).equals(sourceDefaultValue.toString()))))) {
             log.debug(
                     "The {} column on the {} table changed default value from {} to {} ",
                     new Object[] { sourceColumn.getName(), sourceTable.getName(),
--- a/symmetric-db/src/main/java/org/jumpmind/db/platform/AbstractDdlBuilder.java
+++ b/symmetric-db/src/main/java/org/jumpmind/db/platform/AbstractDdlBuilder.java
@@ -2032,7 +2032,7 @@ public abstract class AbstractDdlBuilder implements IDdlBuilder {
 		}
 	}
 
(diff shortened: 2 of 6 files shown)


mmichalek:
Lukas, we are concerned this PR will cause unneeded schema changes for existing SymmetricDS users. Also, this PR makes a decision about what the default value should be too deep in the logic.  Can you describe your desired behavior here?  Maybe there's another way we can solve it.

woehrl01:
Mark, I think I understand your concerns. In my opinion this PR reduces the number of schema changes for existing users (at least it reduced a behavior I experienced over the last few years)

The problem I encountered was that if a default value gets mapped across different database dialects (in my case sqlserver and sqlite). On the client side that the ModelComparator finds a difference as eg. `GETUTCDATE()` isn't equal to `current_timestamp` (see #14) . You should be able to easily reproduce this with an existing dialect mapping (e.g. MySql `SYSDATE` to Sqlite `current_date`)), and after sending a schema again, via SendSchema you will result in a unnecessary schema modification on the client.

Please let me know if you need a better explanation.

mmichalek:
Lukas, I think I see what you are saying.   You mentioned this is common between MSSQL and SQLite.  Can you provide sample schema XML I could use to reproduce between from MSSQL to SQLite?

Thanks,
Mark

woehrl01:
Sure, will do, thanks Mark!

rudiejd:
Closing as author has not responded for 3 years. Feel free to rebase and open a new PR

woehrl01:
Super disappoining, closing this for me not responding whereas you are the one who is not responding 👎🏼

rudiejd:
@woehrl01 we were waiting for you to respond to mark on this one with a sample. This is also 3 years old, so it's not really relevant to the current version of Symmetric

woehrl01:
Whatever. As there has been no interest from JumpMind in the last 3 years for having this or any of the other PR merged in any form or contributions brought to a mergable form, I consider closing this PR and the other ones a good decision. PR are not a one-way street. Simply see the plenty other PR created, closing this simply down and putting the blame on the author is pretty rude in my opinion.

rudiejd:
I didn't close any of your other PRs, just this one where you didn't respond to Mark's comment asking for a sample schema and another PR depending on that same question. You closed all the other ones yourself.

Sorry for the late response, and we don't mean to put the blame on you - just trying to get these older PRs cleaned up. We should have commented on many of them earlier. A lot of them, like yours, even precede our Azure tests, so it's hard to tell which older PRs will break things.

woehrl01:
I certainly understand cleaning older PR up. By closing it due to no response without further interaction, looks like basically putting the blame on the author. 

Of course, there is a lot more to the story and how this presents in this PR. There have been time-consumingly a lot of other PR prepared, made and issues filed in the bugtracker during that days by my team and me during the 2017-2019 years, where basically no real interest has been shown from JumpMind side in active collaboration and accepting contributions. I have certainly moved forward in the last few years, so I'm fine with leaving old things behind. That's why I also closed the other PRs.

As a hint, #14 and #80 are basically non-related to each other, except that it's about a MSSQL to SQLite compatibility contribution.

PR #82, opened 2018-05-04
FileSyncService should use the file_trigger table for cache invalidation

woehrl01:
`FileSyncService.refreshFromDatabase` should invalidate based on the `file_trigger` table and not the `trigger` table.


woehrl01:
Has been fixed in 48953ca

PR #83, opened 2018-05-04
0003547: Improve Sqlite DdlBuilder support

woehrl01:
This PR improves the DdlBuilder support on Sqlite.

Due to the fact that ddl support on Sqlite is very limited, most of the changes in schema lead to recreating the whole table which can take quite some time.

I implemented the following changes:

- A rename of the table is now executed with the `alter table ... rename` syntax
- Adding of columns is now supported via the `alter table ... add column` syntax
- Default values are surounded with parenthesis to prevent some syntax errors

This changes sum up in much less preasure on the database on schema changes.


mmichalek:
Thanks for this PR Lukas.

PR #90, opened 2018-05-15
0003563: improve performance of purging extract request table

woehrl01:
The query of the purging of extract requests can be improved by using `not exists` and doing improved queries if start- and endbatch ids are the same.

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/PurgeServiceSqlMap.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/PurgeServiceSqlMap.java
@@ -33,9 +33,12 @@ public class PurgeServiceSqlMap extends AbstractSqlMap {
         
         putSql("minDataGapStartId", "select min(start_id) from $(data_gap)");
         
-        putSql("deleteExtractRequestSql", "delete from $(extract_request) where status=? and last_update_time < ? and "
-                + "0 = (select count(1) from $(outgoing_batch) where status != 'OK' and batch_id between $(extract_request).start_batch_id and $(extract_request).end_batch_id)");
-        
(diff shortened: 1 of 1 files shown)


mmichalek:
Lukas, can you give a little more background about the problem you faced which led to this code change?

We had not heard of users having performance problems with this particular query.  And I am concerned the complicating the logic will not necessarily perform better on all database platforms.

mmichalek:
Ah - now I see you posted a timeout on the issue.  Sorry I missed that.  What conditions did that occur in (db platform? sym_outgoing_batch count?)  Thanks.

woehrl01:
Sure, I had about 200.000 entries inside the extract request table, and about 10.000.000 batches.

I can only describe the problem as I haven't saved the query plan of sqlserver 2016 standard. It did a full index scan on both tables, resulting in a very expensive hash join. The query performance comparison of both queries resulted in a 0% to 100% execution time.

I just did that change Asti didn't want to do a bigger change. I think this part should be changed in two separate queries, like the "fast data purge"

woehrl01:
Is this information helpful to you?

mmichalek:
Thanks, Lukas, that is helpful.  I also reviewed this change with the team now.  If we are understanding correctly, does this fix to the SQL only help in situations where there are lots of single batches extract requests?  
If so we might want to look into a more general solution that would help the extra purge SQL in general.  We were thinking either the query could be broken along the lines of what purgeByMinMax does or maybe even the "fast data purge" purge technique you mentioned.

woehrl01:
Haven't thought about using the `purgebyminmax`. Sounds like a good idea! I'll supply a diff. 
Thanks Mark!

rudiejd:
Closing as you haven't responded for 3 years with a diff. Again, feel free to rebase to 3.12 and reopen if this is still an issue.

PR #95, opened 2018-10-05
Excluded columns that are part of an index the index remains in table object

woehrl01:
This is similar to 208106c (0003689: Excluded columns that are part of a foreign key the foreign key remains in table object). If you exclude a column, and it is part of an index, the index still gets synced and thus creates an exception on the client.


els0r:
Hello. Is there a reason why this hasn't been merged yet? I'm running into the same issue, where I want to exclude column X but IX_X exists and is synced, thus complaining that column X doesn't exist in the table.

els0r:
Awesome. Thank you

PR #96, opened 2018-12-14
fix double check and synchronization

davecramer:
Unfortunately there's no way around not synchronizing. Even the first unsynchronized check could be wrong.


mmichalek (approved):

mmichalek (commented):
Dave, thanks for taking the time to review these initialization race conditions. There are some cases where we would rather take the approach of moving the initialization code into some sort of controlled "init" method, rather than synchronizing the whole "get" method.  This would be especially true for many of the methods on AbstractParameterService - methods which are called often by the application code, such as AbstractParameterService.getNodeGroupId()

davecramer:
@mmichalek synchronization isn't as bad as people think. Given the order of magnitude of other operations such as http and jdbc it's minimal.

That said how the multi-threaded issues are dealt with is irrelevant, as long as they are

jaredfrees:
Thanks for the good work @davecramer

PR #98, opened 2018-12-17
Make all inner classes static

davecramer:
from Joshua Bloch's Effective Java - Item 22*: Favor static member classes over nonstatic.

This does require the fix to H2DdlReader to pass all tests

### DIFF
--- a/symmetric-android/src/main/java/org/jumpmind/symmetric/android/AndroidSymmetricEngine.java
+++ b/symmetric-android/src/main/java/org/jumpmind/symmetric/android/AndroidSymmetricEngine.java
@@ -144,7 +144,7 @@ public class AndroidSymmetricEngine extends AbstractSymmetricEngine {
         return new AndroidClusterService(parameterService, symmetricDialect, nodeService);
     }
 
-    class AndroidRouterService extends RouterService {
+    static class AndroidRouterService extends RouterService {
 
         public AndroidRouterService(ISymmetricEngine engine) {
             super(engine);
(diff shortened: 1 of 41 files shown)


jaredfrees:
Why do you want to do this to the code?

davecramer:
https://softwareengineering.stackexchange.com/questions/238782/why-prefer-non-static-inner-classes-over-static-ones

jaredfrees:
I think I get the principle, but I don't think some of these classes should be static. For example, FileInfo in FtpDataWriter.java. We want to be able to create and pass around a FileInfo object on the fly basically having it work like a struct in c. I just don't see a great benefit in a situation like this. Feel free to explain otherwise.
Edit: I think I get what you are saying more now after reading this: https://stackoverflow.com/questions/24953370/bloch-effective-java-favor-static-classes-over-nonstatic-how-many-instances
The keyword static in front of class is confusing in Java..

jaredfrees:
So basically, a static nested class saves space because it does not include an instance of the enclosing class when it is created? While a non-static inner class can reference its outer class but we don't really need our inner classes to do this.

jaredfrees:
So anyway, I think you are correct and I am going to review these changes soon and try to merge the commits.

jaredfrees:
@davecramer Can you remove your first commit: fix double check and synchronization? This was already added.

davecramer:
sure
Dave Cramer

davecramer:
Thanks, I was getting to it, I kinda figured there must be some way to ignore a commit using github though

jaredfrees:
Thanks! @davecramer

jaredfrees:
Yeah I think it just ignored it itself because the commit was already there from yesterday

PR #99, opened 2018-12-17
fix possible null dereference issues

davecramer:


symmetric-io/src/main/java/org/jumpmind/symmetric/io/stage/StagingManager.java
```diff
@@ -112,8 +112,17 @@ protected void clean(Path path, long ttlInMs, StagingPurgeContext context) throw
                 clean(entry, ttlInMs, context);
             } else {     
                 try {
+                    String parentDirectory ="";
```
jaredfrees, line 115: I think I would rather do it like this:
         
      try {
            String stagingPath = null;
            IStagedResource resource = null;
             if (entry.getParent() != null && entry.getFileName() != null) {
                    stagingPath = StagedResource.toPath(directory, 
                            new File((entry.getParent().toString() + "/" + entry.getFileName().toString())));
                    resource = createStagedResource(stagingPath);  
             }
             if (stagingPath != null) {

So we can avoid trying to make a new file with parentDirectory and entryName as empty strings ("") so the new FIle path would just be new File("/")

davecramer: Your call

jaredfrees (changes requested):
I would like to go with the changes I said before, but the rest looks good. So:
```
try {
    String stagingPath = null;
    IStagedResource resource = null;
    if (entry.getParent() != null && entry.getFileName() != null) {
        stagingPath = StagedResource.toPath(directory, 
            new File((entry.getParent().toString() + "/" + entry.getFileName().toString())));
        resource = createStagedResource(stagingPath);  
    }
    if (stagingPath != null) {
```

symmetric-io/src/main/java/org/jumpmind/symmetric/io/stage/StagingManager.java
davecramer: Actually looking at this again. What happens in your solution if the parent is null ?


If you really want to avoid creating the empty string it is relatively simple:

String entryName=null;
Then when we create  the file:
new File(parentDirectory + "/" + entryName==null?'':entryName)

jaredfrees: In my solution, if the parent is null then nothing happens because both the if blocks are skipped and the code goes to the next entry in the for loop.
I guess I was assuming if entry.getParent() and entry.getFileName() do not return null then it has the correct parent and filename. The parent could be the empty string but I wouldn't want to assume that unless it is returned from entry.getParent(). 
Also, in your code, you might as well just put String entryName=""; because if it is null you're saying it should just return an empty string into the new FIle( ... path ... )

davecramer: Parent is null if you are at the root. So if the path name was /  then parent is null.

jaredfrees: Okay good point I did not realize that. Then I think if entry.getFileName() is null then nothing should happen and if entry.getParent() is null then it can be the empty string.
Edit: I am going to look at the methods a bit more now to make sure.

davecramer:
Pretty sure I was attempting to replicate the existing behaviour without
changing the semantics.


Dave Cramer

jaredfrees:
You're right it does keep the existing behavior, I am going to accept these commits then.

jaredfrees:
Sorry for interrogating and thanks for the changes.

davecramer:
no worries, thanks!

PR #100, opened 2019-01-30
0003868: quote table names used in SQLITE3 pragma statements.

gislik:
This pull request fixes issue [0003868](https://www.symmetricds.org/issues/view.php?id=3868).


jaredfrees:
Hi, thank you very much for pointing this out. I am going to make this change on the current branch, but with a little different code so it is more maintainable in the future. It will make it easier for use the accept your changes in the future if you sign this license agreement first: https://www.symmetricds.org/developer/contributor

PR #102, opened 2019-03-03
0003659: SqlAnywhere dialect crashes when DB name has special char

matthoward:


jaredfrees:
This looks pretty good, thanks for the bug fix Matt. I will look at this more next week when I have time and most likely approve it.

jaredfrees:
We realized that if you quote the DB name then it can become case sensitive so it could break other use cases. We need to evaluate this further, but thanks for pointing out the special character problem.

jaredfrees:
Everything looks good now. Thanks!

PR #103, opened 2019-04-12
3.9

deividnn:
how do i define a diferent schema to create tables like sym_data ,sym_data_event ,etc in postgresql?


erilong:
Hi deividnn, we generally don't merge newer releases into older releases.  We're you just trying to ask a question?  You can get help at the forums or chat room here:  https://www.symmetricds.org/get-help

On postgres, try giving SymmetricDS its own user/role and then set the search_patch to the schema you want, like this:

ALTER ROLE <your_login_role> SET search_path TO schema1;

PR #104, opened 2019-05-14
Added SQL that was removed in 3.9+ that generated SQLException

bradvillamtgov:
When upgrading from 3.8.43 to 3.9.x or 3.10.x there was an issue where the ConfigurationService class was trying to retrieve Node Channels. The ConfigurationService calls for the query "selectNodeChannelControlLastExtractTimeSql" which does not exist in the ConfigurationServiceSqlMap. I added the query from the 3.8 branch and was able to successfully upgrade from 3.8.43 to 3.10.2. 

[SQLException_message.txt](https://github.com/JumpMind/symmetric-ds/files/3179554/SQLException_message.txt)


erilong:
Thanks for the report and fix!  Tracking it here: https://www.symmetricds.org/issues/view.php?id=3957

PR #105, opened 2019-06-20
Volt template dirty

ssomagani:
When replicating tables into VoltDB from H2, the default JdbcSqlTemplate is picked sometimes which leads to VoltDB's JDBC driver erroring out setting the Transaction Isolation Level


erilong:
Thanks, we need to get our unit tests running on VoltDB, so we don't create another issue like that.

Can you sign our CLA to share copyright with us?

ssomagani:
Thanks Eric. 

I’ve signed the CLA.

PR #107, opened 2019-11-04
0004124: synchronize saveRegistrationRequest by named mutex

mkurz01:


erilong:
Shouldn't clients all have different IP addresses and not be using the loopback address?  It seems like the problem is from the data, not the source code.

pavel-jm:
The wait on author period had expired.

PR #108, opened 2019-12-10
0004189: Node online detector/ ping nodes

mkurz01:
See ticket 0004189


pavel-jm:
The Heartbeat batch on the system channel is accomplishing a similar task, arguably more thorough. This is because the ICMP (network protocol required for ping to work) is not part of the SymmetricDS firewall requirements and could be blocked.

PR #109, opened 2020-02-28
Don't take into account the order that the primary keys are in when choosing to drop and recreate the primary key

chenson42:


erilong (changes requested):
Should it be !caseSensitive && srcPk.getName().equalsIgnoreCase(tgtPk.getName())))

PR #116, opened 2020-10-15
android: Use SupportSQLite instead of the framework version

maxencehenneron:
This PR replaces the framework SQLite version with the new SupportSQLite from AndroidX. (https://developer.android.com/reference/androidx/sqlite/db/SupportSQLiteDatabase)

This new implementation allows to use more SQL implementations besides the framework one, like [room](https://developer.android.com/topic/libraries/architecture/room), [sqldelight](https://github.com/cashapp/sqldelight). I can create a PR to the android client demo once this gets merged but the changes required are minimal.

### DIFF
--- a/symmetric-android/src/main/java/org/jumpmind/symmetric/android/AndroidDatabasePlatform.java
+++ b/symmetric-android/src/main/java/org/jumpmind/symmetric/android/AndroidDatabasePlatform.java
@@ -34,17 +34,18 @@ import org.jumpmind.db.sql.ISqlTemplate;
 import org.jumpmind.db.sql.SqlTemplateSettings;
 
 import android.content.Context;
-import android.database.sqlite.SQLiteOpenHelper;
+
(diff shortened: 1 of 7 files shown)


maxencehenneron:
The build failed due to a gateway timeout... Will likely need to be restarted.

erilong:
Thanks for the PR.  We'll look at the build problem.  The Azure pipelines was a bit of an experiment.

Max, could you share copyright with the project steward by signing the following contributor agreement?  You can enter your information and submit it online:  https://www.symmetricds.org/developer/contributor

After that, we'll accept your code change.  Thanks!

maxencehenneron:
@erilong Thank you, I signed the agreement!

erilong:
Had to revert.  It won't build with our Gradle build.  Seems like we're missing a dependency, but AndroidX seems to use AAR dependency files and I don't know how to include those.  Any help appreciated.

PR #122, opened 2020-11-19
0004345: Postgres array types improperly cast into non-array types

boggard:
Support for array types and also json/jsonb types.


philipmarzullo64:
Will you agree to signing the Contributor License Agreement document so that we can accept this pull request?

The link to the form is located here:

Thanks.

boggard:
Yes, I signed agreement. PR can be accepted.

PR #140, opened 2021-02-04
0004800: Subselect router expression wrapped in parentheses

boggard:


erilong:
Fixed for 3.13.  https://www.symmetricds.org/issues/view.php?id=4800

PR #154, opened 2021-03-26
0004925: Add a new parameter that allows you to control the trigger prefix independent of the table prefix

chenson42:
Add a new parameter that allows you to control the trigger prefix independent of the table prefix


chenson42:
My use case is i have two SymmetricDS networks that need to source data from the same database.   I want to install two nodes pointing to the same tables with the same configuration.   If I do that they will create triggers with the same name.  Adding this property will allow me to change the prefix for one of the nodes.

PR #163, opened 2021-08-02
Fix: V14 label tweak

rudiejd:
Configure labels to use Vaadin components instead of innerHTML. This allows us to avoid re-rendering the entire span when right icon, left icon, or text is set. Also resolves the bugs with Label being refreshed and removed on the dashboards.  Label is now a Span with `display: flex` property set.

### DIFF
--- a/symmetric-sqlexplorer/src/main/java/org/jumpmind/vaadin/ui/common/Label.java
+++ b/symmetric-sqlexplorer/src/main/java/org/jumpmind/vaadin/ui/common/Label.java
@@ -28,81 +28,81 @@ import com.vaadin.flow.component.icon.VaadinIcon;
 
 public class Label extends Span {
 
-	private static final long serialVersionUID = 1L;
+    private static final long serialVersionUID = 1L;
 
-	private Html html;
+    private Html html;
 
-	private Component icon;
+    private Component icon;
 
-	public Label() {
-		super();
-	}
+    public Label() {
+        super();
+    }
 
-	public Label(String text) {
-		html = new Html("<div>" + text + "</div>");
-		getElement().getStyle().set("display", "flex");
-		getElement().getStyle().set("margin", "auto");
-		add(html);
-	}
+    public Label(String text) {
+        html = new Html("<div>" + text + "</div>");
+        getElement().getStyle().set("display", "flex");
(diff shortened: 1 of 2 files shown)


symmetric-sqlexplorer/src/main/java/org/jumpmind/vaadin/ui/common/Label.java
```diff
@@ -29,97 +30,79 @@
...
-    @Override
     public String getText() {
         return getElement().getProperty("innerHTML");
     }
     
-    @Override
     public void setText(String text) {
-        this.text = text;
-        updateLabel();
+        if (this.text != null) {
```
erilong: It confused me that "text" is String but "this.text" is Html.  Could you rename "this.text" to "this.html"?  Then you don't need to use "this" anymore since there is no ambiguousness.

symmetric-sqlexplorer/src/main/java/org/jumpmind/vaadin/ui/common/Label.java
```diff
@@ -29,97 +30,79 @@
...
-        icon.getElement().getStyle().set("margin-top", "-4px");
-        if (icon != null && icon instanceof Icon) {
-            ((Icon) icon).setSize("1em");
+    public void setLeftIcon(Component icon) {
+        if (this.icon != null) {
+            remove(this.icon);
         }
+        icon.getElement().getStyle().set("align-self", "flex-start");
+        this.icon = icon;
+		add(icon);
```
erilong: Fix indent

symmetric-sqlexplorer/src/main/java/org/jumpmind/vaadin/ui/common/Label.java
```diff
@@ -29,97 +30,79 @@
...
     }
 
+    public void setRightIcon(Component icon) {
+        if (this.icon != null) {
+            remove(this.icon);
+        }
+        icon.getElement().getStyle().set("align-self", "flex-end");
+        this.icon = icon;
+		add(icon);
```
erilong: fix indent

erilong (changes requested):

PR #174, opened 2021-08-19
fix: 0005074 CSV to database replication not working

rudiejd:
- a [user on the forum](https://sourceforge.net/p/symmetricds/discussion/1354726/thread/86e46e7027/) was having trouble with CSV to database replication
- josh was able to reproduce this issue and so was i
- the extractor service was not respecting the virtual CSV trigger because its triggerhistory was null, and was marking the data as having a missing trigger so that it could not be extracted
- i moved the check for whether the trigger is the  virtual one up higher and added a triggerhistory entry for it
- this resolves the issue and allows batches to be extracted

I also added to the docs to explain that users have to put the column names at the top of the CSV file for proper CSV routing

### DIFF
--- a/symmetric-assemble/src/asciidoc/examples/csv2db.ad
+++ b/symmetric-assemble/src/asciidoc/examples/csv2db.ad
@@ -85,7 +85,7 @@ NOTE: 'TARGET_BASE_DIR' should be changed to your target base directory.
 
 endif::pro[]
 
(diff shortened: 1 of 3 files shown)


erilong:
This is the right way to fix it, but it needed refactored to work in 3.16  See https://github.com/JumpMind/symmetric-ds/pull/328

PR #175, opened 2021-08-20
fix: 0004116 syncing postgres timestamp to time produces inaccurate vals

rudiejd:
- the [linked issue in the issue tracker] describes a problem a user had when going from timestamps in postgres to other databases
- i was able to reproduce this in 3.12
- i tracked the issue down to FormatUtils - the problem is that Java does not support time precision above milliseconds, so it was trying to use the nanosecond value provided by postgres as additional milliseconds, pushing the time up a couple of minutes
- i added a check for times with precision greater than milliseconds. i just truncate the value instead of rounding it since that's more accurate to what would actually be recorded 
- i added a regression test to format utils for detecting this bug


erilong:
Issue 4116 is fixed for 3.14.2.

PR #182, opened 2022-05-09
Fix that MS SQL initial load with VARCHAR > 8000 will succeed

SndR85:
I had an issue where I cannot import a database with tables where the VARCHAR size was bigger than 8000 bytes. With this change SymmetricDS will create the column with `VARCHAR(MAX)` on Microsoft SQL Server when the size of VARCHAR is bigger than 8000 bytes, which in my case succeed.


evan-miller-jumpmind:
Go ahead and sign out Contributor License Agreement and I'll accept the pull request:

SndR85:
I have signed the agreement.

PR #185, opened 2022-09-30
fix: 0005481 mssql text for symmetric tables to varchar

SndR85:
Fixes https://www.symmetricds.org/issues/view.php?id=5481. With this change the SymmetricDS tables can be created on MS SQL with _UTF8 of _SC collation by setting the `mssql.use.varchar.for.lob.in.sync` to `true`

### DIFF
--- a/symmetric-client/src/main/java/org/jumpmind/symmetric/db/mssql/MsSqlSymmetricDialect.java
+++ b/symmetric-client/src/main/java/org/jumpmind/symmetric/db/mssql/MsSqlSymmetricDialect.java
@@ -25,6 +25,7 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.sql.Types;
 
 import org.apache.commons.lang3.StringUtils;
 import org.jumpmind.db.model.Column;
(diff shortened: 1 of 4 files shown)


symmetric-client/src/main/java/org/jumpmind/symmetric/db/mssql/MsSqlSymmetricDialect.java
```diff
@@ -73,11 +74,20 @@ public MsSqlSymmetricDialect(IParameterService parameterService, IDatabasePlatfo
     public Database readSymmetricSchemaFromXml() {
         Database db = super.readSymmetricSchemaFromXml();
         if (parameterService.is(ParameterConstants.MSSQL_USE_NTYPES_FOR_SYNC)) {
-            Table table = db.findTable(TableConstants.getTableName(getTablePrefix(),
+            Table sym_table = db.findTable(TableConstants.getTableName(getTablePrefix(),
```
erilong: This should be symTable to follow our lower camel case naming convention.

symmetric-core/src/main/resources/symmetric-default.properties
```diff
@@ -3155,7 +3163,7 @@ cloud.bulk.row.terminator=\r\n
 #
 # DatabaseOverridable: true
 # Tags: cloud, snowflake, azure
-cloud.bulk.field.quote=�
```
erilong: Why was the cloud.bulk.field.quote parameter changed?  That seems unrelated to the issue.

erilong (changes requested):

symmetric-core/src/main/resources/symmetric-default.properties
SndR85: Yes, that should not happen, sorry for that, I will try to undo that.

SndR85: I have removed this line change.

symmetric-client/src/main/java/org/jumpmind/symmetric/db/mssql/MsSqlSymmetricDialect.java
SndR85: I have changed it in https://github.com/JumpMind/symmetric-ds/pull/185/commits/69760f23a9004b20e29ec14d17ded246948b6335

PR #186, opened 2022-09-30
0005483: Toggle to convert TEXT to VARCHAR in MS SQL

SndR85:
This PR will fix https://www.symmetricds.org/issues/view.php?id=5483.

When a column with the type of `long(n)varchar` is created it will be created in MS SQL as `(n)varchar(max)`. As this should not be the default action, I have created an environment variable which can activate this conversion. Maybe is there a better place to do this kind of type conversions?

### DIFF
--- a/symmetric-db/src/main/java/org/jumpmind/db/platform/mssql/MsSql2005DdlBuilder.java
+++ b/symmetric-db/src/main/java/org/jumpmind/db/platform/mssql/MsSql2005DdlBuilder.java
@@ -98,10 +98,7 @@ public class MsSql2005DdlBuilder extends MsSql2000DdlBuilder {
             sqlType = String.format("DECIMAL(38,%d)", column.getScale());
         } else if (useVarcharForText && (column.getMappedTypeCode() == Types.LONGVARCHAR || column.getMappedTypeCode() == Types.LONGNVARCHAR || column
                 .getMappedTypeCode() == Types.CLOB)) {
-            sqlType = "";
-            if (column.getMappedTypeCode() == Types.LONGNVARCHAR) {
-                sqlType = "N";
-            }
+            sqlType = (column.getMappedTypeCode() == Types.LONGNVARCHAR) ? "N" : "";
             sqlType += "VARCHAR(MAX)";
         }
         return sqlType;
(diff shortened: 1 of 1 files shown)


symmetric-db/src/main/java/org/jumpmind/db/platform/mssql/MsSql2005DdlBuilder.java
```diff
@@ -87,13 +88,24 @@ protected void dropColumnChangeDefaults(Table sourceTable, Column sourceColumn,
 
     @Override
     public String getSqlType(Column column) {
+        Map<String, String> env = System.getenv();
+        boolean useVarcharForText = "y".equalsIgnoreCase(env.get("USE_VARCHAR_FOR_TEXT"));
```
erilong: This should be a system property using our dot naming convention instead of an environment variable.  For example, use System.getProperty("mssql.use.varchar.for.lob", "true").equalsIgnoreCase("true").

symmetric-db/src/main/java/org/jumpmind/db/platform/mssql/MsSql2005DdlBuilder.java
```diff
@@ -87,13 +88,24 @@ protected void dropColumnChangeDefaults(Table sourceTable, Column sourceColumn,
...
             sqlType = "VARCHAR(MAX)";
         } else if (column.getMappedTypeCode() == Types.DECIMAL && column.getSizeAsInt() > 38) {
             sqlType = String.format("DECIMAL(38,%d)", column.getScale());
+        } else if (useVarcharForText && (column.getMappedTypeCode() == Types.LONGVARCHAR || column.getMappedTypeCode() == Types.LONGNVARCHAR || column
+                .getMappedTypeCode() == Types.CLOB)) {
+            sqlType = "";
+            if (column.getMappedTypeCode() == Types.LONGNVARCHAR) {
+                sqlType = "N";
+                final String newSize = Integer.toString(column.getSizeAsInt() * 2);
+                column.setSize(newSize);
```
erilong: What is the reason for modifying the column size here?  The getSqlType() method is a get method that normally wouldn't modify what's passed in.  It already returns VARCHAR(MAX) with a maximum size, so there doesn't seem to be a reason to modify the column size also.

erilong (changes requested):

symmetric-db/src/main/java/org/jumpmind/db/platform/mssql/MsSql2005DdlBuilder.java
SndR85: Sorry, that was incorrect to add. I have removed the lines for newSize.

symmetric-db/src/main/java/org/jumpmind/db/platform/mssql/MsSql2005DdlBuilder.java
SndR85: I have replaced it by getting the property as you suggested. The default value should be false as it should not be enabled by default.

PR #187, opened 2022-10-03
0005496 : Optionally cast varchar to nvarchar

SndR85:
I have the challenge to cast varchar to nvarchar in synchronisation. I have created a PR for that.


symmetric-db/src/main/java/org/jumpmind/db/platform/mssql/MsSql2000DdlBuilder.java
```diff
@@ -169,6 +170,10 @@ protected void dropTable(Table table, StringBuilder ddl, boolean temporary, bool
         println("  CLOSE refcursor", ddl);
         println("  DEALLOCATE refcursor", ddl);
         ddl.append("  DROP TABLE ");
+        if (schema != null) {
```
erilong: Can you change this to if StringUtils.isNotBlank(schema) instead?

SndR85: I've changed it as you suggested

PR #200, opened 2024-10-15
6616: Save column references as numeric values for faster look-ups in AbstractDatabaseWriter.getRowData

pavel-jm:
6616: Save column references as numeric values for faster look-ups in AbstractDatabaseWriter.getRowData(). New TableColumnSourceReferences class to store column lookups plus a unit test.


pavel-jm:
Targeting 3.16 instead

PR #203, opened 2024-10-22
0006621: Triggers to include NVARCHAR(MAX) and VARCHAR(MAX) columns (MS SQL Server-specific) 

pavel-jm:
0006621: Triggers to include NVARCHAR(MAX) and VARCHAR(MAX) columns (MS SQL Server-specific)


symmetric-core/src/main/java/org/jumpmind/symmetric/db/AbstractTriggerTemplate.java
```diff
@@ -786,11 +786,18 @@ protected boolean requiresEmptyLobTemplateForDeletes() {
         return false;
     }
 
+    /***
+     * Helps detect Large Object columns. Some LOBs are inaccessible to triggers or require specialized code.
+     */
+    public boolean isLob(Column column) {
```
erilong: This could be "protected" to emphasize its use for overriding and prevent someone from thinking its part of a public API.

PR #206, opened 2024-10-29
6620 The UNLOGGED table mode during initial load into an empty PostgreSQL table

pavel-jm:
0006620: Turn on UNLOGGED table mode during initial load into an empty PostgreSQL table


pavel-jm:
Wrong target branch!

PR #208, opened 2024-11-12
0006647: Pause/Resume Nodes

joshahicks:
Ability to leverage the sym_node_channel_ctl table to pause and resume node synchronization without losing changes.


erilong (commented):
The data sync is using ConfigurationService to get the node channels, so isn't this information already available for use?  There is already getNodeChannels() and saveNodeChannelControl(), although you might need to expose deleteNodeChannelControl() that is buried in deleteChannel().

erilong:
Fixed in https://github.com/JumpMind/symmetric-ds/pull/209 instead

PR #211, opened 2024-11-20
0006655: Add API to clear node host entries

erilong:
Add a REST API call to clear node host entries by node_id or by node_id and instance_id.

### DIFF
--- a/symmetric-core/src/test/java/org/jumpmind/symmetric/service/impl/MockNodeService.java
+++ b/symmetric-core/src/test/java/org/jumpmind/symmetric/service/impl/MockNodeService.java
@@ -397,4 +397,8 @@ public class MockNodeService implements INodeService {
     public boolean setInitialLoadEnded(ISqlTransaction transaction, String nodeId) {
         return false;
     }
+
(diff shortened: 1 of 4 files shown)


symmetric-core/src/test/java/org/jumpmind/symmetric/service/impl/MockNodeService.java
```diff
@@ -397,4 +397,8 @@ public NodeStatus getNodeStatus(String nodeId) {
     public boolean setInitialLoadEnded(ISqlTransaction transaction, String nodeId) {
         return false;
     }
+
+    @Override
+    public void deleteNodeHostInstance(String nodeId, String instanceId) {
+    }
 }
```
pavel-jm, line 404: Just curios - is the MockNodeService class part of some unit test, which is not checked in to this branch?

erilong: It's used in unit tests like DefaultNodeIdCreatorTest.

PR #213, opened 2024-12-13
0006676: Comments on stored procedures do not get migrated with DDL

catherinequamme:
Ticket: https://issues.symmetricds.org/view.php?id=6676

This fixes the issue, not sure if we need a new variable instead of stripOutBlockComments.

### DIFF
--- a/symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/writer/DefaultDatabaseWriter.java
+++ b/symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/writer/DefaultDatabaseWriter.java
@@ -1075,7 +1075,7 @@ public class DefaultDatabaseWriter extends AbstractDatabaseWriter {
     protected List<String> getSqlStatements(String script) {
         List<String> sqlStatements = new ArrayList<String>();
         SqlScriptReader scriptReader = new SqlScriptReader(new StringReader(script));
-        scriptReader.setStripOutBlockComments(writerSettings.isStripOutBlockCommentsInScripts());
+        scriptReader.setStripOutComments(writerSettings.isStripOutCommentsInScripts());
         try {
             String sql = scriptReader.readSqlStatement();
             while (sql != null) {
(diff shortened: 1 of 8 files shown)


symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/writer/DefaultDatabaseWriter.java
```diff
@@ -1075,6 +1075,7 @@ protected void bindVariables(Map<String, Object> variables) {
     protected List<String> getSqlStatements(String script) {
         List<String> sqlStatements = new ArrayList<String>();
         SqlScriptReader scriptReader = new SqlScriptReader(new StringReader(script));
+        scriptReader.setStripOutBlockComments(false);
```
pavel-jm: In order to support backwards compatibility I recommend adding a new parameter with default value of "false" in the symmetric-default.properties file (as well as in ParameterConstants.java)

Sample moniker: ddl.sync.routine.comments
Sample constant: public final static String DDL_SYNC_ROUTINE_COMMENTS="ddl.sync.routine.comments"

erilong: The DefatulDatabaseWriter can access DatabaseWriterSettings (variable is writerSettings), so add a property like stripOutComments to it.  You can set it to the parameter value in AbstractDataLoaderFactory class buildParameterDatabaseWriterSettings() method.  Those parameters are prefixed with "dataloader", so maybe call it "dataloader.sql.event.strip.comments".

erilong (changes requested):

symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/writer/DefaultDatabaseWriter.java
pavel-jm: Added new setting dataloader.sql.event.strip.comments with a default of true

erilong (approved):

PR #214, opened 2024-12-27
0006684 ISO format for TransactionId in DataService.reCaptureData

pavel-jm:
0006684 ISO format for TransactionId in DataService.reCaptureData
Plus refactor of a magic number 30000 to the PROGRESS_LOG_UPDATE_DELAY_MS constant for regular log entries about progress


joshahicks (approved):
Looks good

PR #216, opened 2024-12-27
0006685 Fix recapture of stale data when row is missing and refactor reCaptureData

pavel-jm:
0006685 Fix recapture of stale data when row is missing and refactor DataService.reCaptureData to process individual stale Data objects

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
@@ -3645,12 +3645,8 @@ public class DataService extends AbstractService implements IDataService {
                 insertList.add(recapturedData);
                 if (System.currentTimeMillis() - lastUpdateTimestamp > PROGRESS_LOG_UPDATE_DELAY_MS) {
                     lastUpdateTimestamp = System.currentTimeMillis();
-                    log.info("Recaptured stale data_id={}; table={} (from a list of {})", recapturedData.getDataId(), recapturedData.getTableName(), dataList
-                            .size());
-                    // TODO: Consider batching records for same table and trigger router together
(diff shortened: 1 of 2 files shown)


philipmarzullo64 (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
```diff
@@ -3596,94 +3597,152 @@ protected void convertDataToReload(Data data, Table table, TriggerHistory hist,
...
-                Set<TriggerRouter> triggerRouters = engine.getTriggerRouterService().getTriggerRouterForTableForCurrentNode(
-                        hist.getSourceCatalogName(), hist.getSourceSchemaName(), hist.getSourceTableName(), false);
-                table = platform.getTableFromCache(hist.getSourceCatalogName(), hist.getSourceSchemaName(), hist.getSourceTableName(), false);
-                if (triggerRouters != null && triggerRouters.size() > 0 && table != null) {
-                    Trigger trigger = triggerRouters.iterator().next().getTrigger();
-                    table = table.copyAndFilterColumns(hist.getParsedColumnNames(), hist.getParsedPkColumnNames(), true, false);
-                    if (data.getDataEventType() == DataEventType.INSERT) {
-                        keys = data.toParsedRowData();
-                        if (keys != null && keys.length >= table.getPrimaryKeyColumnCount()) {
-                            keys = ArrayUtils.subarray(keys, 0, table.getPrimaryKeyColumnCount());
```
pavel-jm, line 3633: Are primary keys ALWAYS in the front in sym_data.row_data?

erilong: Yes, it orders the columns so that primary keys are at the front of row_data.

PR #217, opened 2025-01-02
0006688: Dynamic reload queue to scale initial load

erilong:


symmetric-core/src/main/java/org/jumpmind/symmetric/cache/OutgoingBatchCache.java
philipmarzullo64, line 1: checkReadyQueuesCache() method should put the semaphore acquire and release in a try/finally because a database query is being executed and it could fail which will throw an exception and skip the release of the semaphore.

philipmarzullo64 (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/cache/OutgoingBatchCache.java
erilong: Fixed it.  Thanks.

symmetric-core/src/main/java/org/jumpmind/symmetric/util/QueueThread.java
```diff
@@ -0,0 +1,65 @@
...
+    public QueueThread(String queue) {
+        if (queue != null) {
+            int index = queue.indexOf(Constants.DELIMITER_QUEUE_THREAD);
+            if (index != -1 && queue.length() >= index + 1) {
+                name = queue.substring(0, index);
+                try {
+                    threadId = Integer.parseInt(queue.substring(index + 1));
+                    usingThreading = true;
+                } catch (NumberFormatException e) {
+                }
```
pavel-jm, line 43: The threadId member remains unassigned when the NumberFormatException occurs. 
Since such exception might be an indicator of a change in database outside SymmetricDS control, it would be helpful to log it - in case there is a support ticket.

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataLoaderService.java
```diff
@@ -647,6 +652,18 @@ protected IDataWriter chooseDataWriter(Batch batch) {
         return batchesProcessed;
     }
 
+    private void notifyQueuesReady(RemoteNodeStatus status, IIncomingTransport transport) {
+        if (status != null && Constants.QUEUE_DEFAULT.equals(status.getQueue())) {
+            String queues = null;
+            Map<String, String> headers = transport.getHeaders();
+            if (headers != null) {
+                queues = headers.get(WebConstants.HEADER_READY_QUEUES);
```
pavel-jm: could the value of "queues" be null when headers originate from an older SymmetricDS version?
And IF so, will that generate a NullPointer on queues.split(...) below

symmetric-core/src/main/java/org/jumpmind/symmetric/util/QueueThread.java
erilong: Added logging.  Thanks.

pavel-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataLoaderService.java
erilong: Yes, I think you're right.  Fixed it.  Thanks!

PR #221, opened 2025-02-02
Support for boolean data types on Apache Derby

RomeoPapaFoxtrot:
Fixes for issues 0006637 & 0006700.

### DIFF
--- a/symmetric-client/src/main/java/org/jumpmind/symmetric/db/derby/DerbyTriggerTemplate.java
+++ b/symmetric-client/src/main/java/org/jumpmind/symmetric/db/derby/DerbyTriggerTemplate.java
@@ -33,6 +33,13 @@ public class DerbyTriggerTemplate extends AbstractTriggerTemplate {
     public DerbyTriggerTemplate(ISymmetricDialect symmetricDialect) {
         super(symmetricDialect);
         //@formatter:off
+
+        // NOTE.
+        // Column type templates are not actually used in the database triggers. However, the presence of a type template
+        // causes a generic trigger (as per trigger templates below) to be created for the relevant column. When this
+        // trigger fires (on insert/update/delete), DerbyFunctions:insertData is eventually called with the JDBC result
+        // set, and the CSV data is then formatted using DerbyFunctions:escape for all types (except blob/clob).
+
         emptyColumnTemplate = "''" ;
         stringColumnTemplate = "sym_escape($(tableAlias).\"$(columnName)\")" ;
         xmlColumnTemplate = null;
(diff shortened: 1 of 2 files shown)


symmetric-client/src/main/java/org/jumpmind/symmetric/db/derby/DerbyTriggerTemplate.java
```diff
@@ -44,7 +51,8 @@ public DerbyTriggerTemplate(ISymmetricDialect symmetricDialect) {
         clobColumnTemplate = "sym_clob_to_string('\"$(columnName)\"', '$(schemaName)$(tableName)', $(primaryKeyWhereString) )" ;
         blobColumnTemplate = "sym_blob_to_string('\"$(columnName)\"', '$(schemaName)$(tableName)', $(primaryKeyWhereString) )" ;
         wrappedBlobColumnTemplate = null;
-        booleanColumnTemplate = null;
+        booleanColumnTemplate = "sym_escape($(tableAlias).\"$(columnName)\")" ;
+
```
pavel-jm, line 49: This escape function should is unnecessary, because boolean value should not have quotes or other symbols to escape.
This template is used to build trigger code, and should only need to convert boolean value to a string (VARCHAR).
May we suggest this edit (as always it's prudent to confirm it work on a real DB):
` booleanColumnTemplate = "case when $(columnName) is null then '' when $(columnName) then '1' else '0' end";`

symmetric-client/src/main/java/org/jumpmind/symmetric/db/derby/DerbyTriggerTemplate.java
```diff
@@ -33,6 +33,13 @@ public class DerbyTriggerTemplate extends AbstractTriggerTemplate {
     public DerbyTriggerTemplate(ISymmetricDialect symmetricDialect) {
         super(symmetricDialect);
         //@formatter:off
+
```
pavel-jm, line 36: Let's move this note to the Pull Request summary, please.

symmetric-db/src/main/java/org/jumpmind/db/platform/AbstractDatabasePlatform.java
```diff
@@ -578,6 +578,11 @@ protected Object parseInteger(String value) {
         }
     }
 
+    protected Object parseBoolean(String value) {
+        value = cleanNumber(value);
+        return value.equals("1") ? Boolean.TRUE : Boolean.FALSE;
+    }
```
pavel-jm, line 584: Looks great!

pavel-jm (changes requested):
Hi,
We've reviewed your PR and requested a few edits. Let me know if you have follow up questions!

symmetric-client/src/main/java/org/jumpmind/symmetric/db/derby/DerbyTriggerTemplate.java
RomeoPapaFoxtrot: I discovered that, for Derby, these trigger templates are not used. Any non-null value will result in a generic trigger being created for the column. Essentially we could set the value to "", and it would work. For clarity, I have amended the trigger template text as per your suggestion.

symmetric-client/src/main/java/org/jumpmind/symmetric/db/derby/DerbyTriggerTemplate.java
RomeoPapaFoxtrot: Comments removed.

symmetric-client/src/main/java/org/jumpmind/symmetric/db/derby/DerbyTriggerTemplate.java
pavel-jm: Thank you.
I'm setting up Derby now. Time permitting I'll test this soon.

pavel-jm:
Already fixed under pull request: https://github.com/JumpMind/symmetric-ds/pull/225

PR #222, opened 2025-02-05
Apply datarows locking scheme to key SymmetricDS tables on Sybase

pavel-jm:
0006716 Apply datarows locking scheme to key SymmetricDS tables on Sybase

### DIFF
--- a/symmetric-db/src/main/java/org/jumpmind/db/platform/ase/AseDdlBuilder.java
+++ b/symmetric-db/src/main/java/org/jumpmind/db/platform/ase/AseDdlBuilder.java
@@ -146,6 +146,15 @@ public class AseDdlBuilder extends AbstractDdlBuilder {
         super.createTable(table, ddl, temporary, recreate);
     }
 
+    /**
+     * Specifies "datarows" for table locking scheme, in order to overcome overly restrictive defaults ("allpages" or "datapages") leading to deadlocks
+     */
+    @Override
+    protected void writeTableCreationStmtEnding(Table table, StringBuilder ddl) {
+        ddl.append(" lock datarows");
+        super.writeTableCreationStmtEnding(table, ddl);
+    }
+
     @Override
     protected void writeColumn(Table table, Column column, StringBuilder ddl) {
         if (column.isGenerated()) {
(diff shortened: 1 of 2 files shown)


symmetric-db/src/main/java/org/jumpmind/db/platform/ase/AseDdlBuilder.java
```diff
@@ -146,6 +146,15 @@ protected void createTable(Table table, StringBuilder ddl, boolean temporary, bo
...
 
+    /**
+     * Specifies "datarows" for table locking scheme, in order to overcome overly restrictive defaults ("allpages" or "datapages") leading to deadlocks
+     */
+    @Override
+    protected void writeTableCreationStmtEnding(Table table, StringBuilder ddl) {
+        ddl.append(" lock datarows");
+        super.writeTableCreationStmtEnding(table, ddl);
+    }
+
```
pavel-jm: _Example_
-- Before: 
CREATE TABLE sym_file_incoming(
    relative_dir VARCHAR(255) NOT NULL,
    file_name VARCHAR(260) NOT NULL,
    last_event_type CHAR(1) NOT NULL,
    node_id VARCHAR(50) NOT NULL,
    file_modified_time NUMERIC(18,0) NULL,
    PRIMARY KEY (relative_dir, file_name)
)

-- AFTER: 
CREATE TABLE sym_file_incoming(
    relative_dir VARCHAR(255) NOT NULL,
    file_name VARCHAR(260) NOT NULL,
    last_event_type CHAR(1) NOT NULL,
    node_id VARCHAR(50) NOT NULL,
    file_modified_time NUMERIC(18,0) NULL,
    PRIMARY KEY (relative_dir, file_name)
) **lock datarows**

Per spec https://infocenter.sybase.com/help/index.jsp?topic=/com.sybase.dc20021_1251/html/locking/X12029.htm

erilong: Instead of affecting all tables (including user tables), we could alter only the SYM runtime tables that need it.  Look in AseSymmetricDialect.createOrAlterTablesIfNecessary() where it currently alters sym_data, sym_data_event, sym_outgoing_batch, and sym_monitor_event to use row-level locks.  Maybe we just need to add some tables to that routine, like sym_extract_request.

pavel-jm: Thanks - The AseSymmetricDialect is the class to change!
With a little refactoring it is now ready for another look.

erilong (approved):

PR #224, opened 2025-02-06
0006721: Data loader incorrect SQL error when no columns on target table

erilong:
If source and target table names match but the column names do not, the dataloader will execute invalid SQL that causes a syntax error. Instead, we should log an illegal state exception explaining that no column names matched. This happened to a user whose intention was to transform column names, but the transform was not setup correctly, and the resulting error message was confusing.  Insert and update needed a check, but delete already had a check to complain about no primary key columns.

### DIFF
--- a/symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/writer/DefaultDatabaseWriter.java
+++ b/symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/writer/DefaultDatabaseWriter.java
@@ -249,7 +249,7 @@ public class DefaultDatabaseWriter extends AbstractDatabaseWriter {
         if (isCteExpression()) {
             retSql = getPlatform().getDdlBuilder().getDatabaseInfo().getCteExpression()
                     + " " + sql;
-            retSql = currentDmlStatement.updateCteExpression(retSql, batch.getSourceNodeId());
+            retSql = DmlStatement.updateCteExpression(retSql, batch.getSourceNodeId());
         }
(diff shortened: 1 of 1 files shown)


symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/writer/DefaultDatabaseWriter.java
```diff
@@ -249,7 +249,7 @@ protected String replaceCteExpression(String sql) {
         if (isCteExpression()) {
             retSql = getPlatform().getDdlBuilder().getDatabaseInfo().getCteExpression()
                     + " " + sql;
-            retSql = currentDmlStatement.updateCteExpression(retSql, batch.getSourceNodeId());
+            retSql = DmlStatement.updateCteExpression(retSql, batch.getSourceNodeId());
```
pavel-jm, line 252: My copy of the DmlStatement.updateCteExpression is not static.
Is a file missing from this PR?

erilong: The method on DmlStatement is available since 2024-10-30.  It is an overloaded method, one is static and the other is not.

pavel-jm: Yep, they are right next there.
False alarm!

pavel-jm (approved):

PR #226, opened 2025-02-10
6722: Purge of batches before data causes appearance of stranded data

erilong:
See 3 issues fixed:  https://issues.symmetricds.org/view.php?id=6722

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/PurgeService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/PurgeService.java
@@ -29,7 +29,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.function.LongConsumer;
 
-import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.time.DateUtils;
 import org.apache.commons.lang3.time.FastDateFormat;
 import org.jumpmind.db.platform.DatabaseNamesConstants;
@@ -234,7 +233,7 @@ public class PurgeService extends AbstractService implements IPurgeService {
         if (rangeMinMax[1] == minMax[1]) {
             minMax[1] = -1;
         } else {
-            minMax[0] = notOkBatchId + 1;
+            minMax[0] = notOkBatchId;
         }
         return minMax;
     }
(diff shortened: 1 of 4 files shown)


evan-miller-jumpmind (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/PurgeService.java
```diff
@@ -926,22 +925,25 @@ private OutgoingContext buildOutgoingContext(Calendar retentionCutoff) {
         context.setMinEventBatchId(startEventBatchId);
         // Leave 1 batch and its data around so MySQL auto increment doesn't reset
         long endBatchId = sequenceService.currVal(Constants.SEQUENCE_OUTGOING_BATCH) - 1;
-        List<Long> batchIds = sqlTemplateDirty.query(getSql("maxBatchIdByChannel"), new LongMapper(),
+        List<Long> batchIds = sqlTemplateDirty.query(getSql("maxBatchIdForOldBatches"), new LongMapper(),
                 new Object[] { startBatchId, endBatchId, new Timestamp(context.getRetentionCutoff().getTime().getTime()) },
                 new int[] { symmetricDialect.getSqlTypeForIds(), symmetricDialect.getSqlTypeForIds(), Types.TIMESTAMP });
         if (batchIds != null && batchIds.size() > 0) {
```
pavel-jm, line 931: If this query returns no rows, do we need to set the Max Batch Id?
like: 
context.setMaxBatchId(startBatchId); ?
log.info("No batch rows found older than {}. New max eligible batch ID: {}"
, context.getRetentionCutoff().getTime().getTime(), startBatchId);

pavel-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/PurgeService.java
erilong: That does happen when there are no batches old enough to purge.  In that case, the max batch ID stays as zero, and you'll see it in the log, where it has a starting batch ID (the last batch ID it purged plus one) and the ending batch ID will be zero.  That intentionally stops it from deleting anything.  Using the starting batch ID for max/ending batch ID would cause it to delete that batch, but it may not be old enough yet, so I think it's correct for the end of the range to come from the query or otherwise be zero.

philipmarzullo64 (approved):

PR #227, opened 2025-02-17
6714: tools like bulk loaders that are installed or downloaded as modules

erilong:


symmetric-core/src/main/java/org/jumpmind/symmetric/util/ModuleManager.java
```diff
@@ -192,6 +206,86 @@ public void installAll() throws ModuleException {
...
+        return contentLength;
+    }
+
+    protected HttpURLConnection getHttpUrlConnection(String urlString, String method) throws IOException {
+        URL url = new URL(urlString);
+        HttpURLConnection conn = (HttpURLConnection) url.openConnection();
+        conn.setRequestMethod(method);
+        conn.setInstanceFollowRedirects(true);
+        conn.setConnectTimeout(10000);
+        conn.setReadTimeout(30000);
```
pavel-jm, line 285: I'm concerned that for some customers 30 seconds may not be enough to download the payload...
Should this be configurable or specified by the caller?

symmetric-core/src/main/java/org/jumpmind/symmetric/util/ModuleManager.java
```diff
@@ -192,6 +206,86 @@ public void installAll() throws ModuleException {
...
+                pb.redirectErrorStream(true);
+                Process proc = pb.start();
+                try (BufferedReader reader = new BufferedReader(new InputStreamReader(proc.getInputStream()))) {
+                    String line = null;
+                    while ((line = reader.readLine()) != null) {
+                        log.info("Install script: {}", line);
+                    }
+                }
+                log.info("Install script RC={}", proc.waitFor());
+            }
```
pavel-jm, line 245: If the does not exist here it might indicate a system problem worth a separate error message?
Example: some permissions are missing, there is not enough free disk space to unpack module, antivirus blocks file, etc.

symmetric-core/src/main/java/org/jumpmind/symmetric/util/ModuleManager.java
erilong: It's a timeout when waiting for data to arrive.  They should have called it idle timeout.  When I tested it, my download took about 180 seconds.

symmetric-core/src/main/java/org/jumpmind/symmetric/util/ModuleManager.java
erilong: Some modules need an install script and others don't, so I was just trying to handle both cases.

symmetric-core/src/main/java/org/jumpmind/symmetric/util/ModuleManager.java
```diff
@@ -119,35 +114,51 @@ public static ModuleManager getInstance() throws ModuleException {
...
+                    String urlString = buildUrl(repo, artifact);
                     try {
-                        URL url = new URL(urlString);
-                        HttpURLConnection conn = (HttpURLConnection) url.openConnection();
-                        conn.setRequestMethod("GET");
-                        conn.setInstanceFollowRedirects(true);
-                        conn.setConnectTimeout(10000);
-                        conn.setReadTimeout(30000);
+                        HttpURLConnection conn = getHttpUrlConnection(urlString, WebConstants.METHOD_GET);
                         if (conn.getResponseCode() == 200) {
```
pavel-jm, line 157: I'm concerned that HTTP redirect codes are not being processed (not sure if they are common)?
These might change target URL in the response: 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections
301 = Website Moved Permanently
307 = Temporary Redirect
308 = Permanent Redirect

evan-miller-jumpmind (approved):

pavel-jm (approved):

PR #229, opened 2025-02-24
0006444 Bulk Load Percentage, and load detail changes

mdrouard-jumpmind:
0006444 Bulk Load Percentage, and load detail changes. 
Various changes made for load detail panel. 
Changes added to extract_request 
and table_reload_status
to add caching

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
@@ -1460,9 +1460,8 @@ public class DataExtractorService extends AbstractService implements IDataExtrac
         if (platform.supportsParametersInSelect()) {
             transaction.prepareAndExecute(getSql("updateExtractRequestLoadTime"), outgoingBatch.getBatchId(), new Date(),
                     outgoingBatch.getReloadRowCount() > 0 ? outgoingBatch.getDataRowCount() : 0,
-                    outgoingBatch.getLoadMillis(), outgoingBatch.getBatchId(), new Date(), outgoingBatch.isBulkLoaderFlag(),
-                    outgoingBatch.getReloadRowCount() > 0 ? outgoingBatch.getDataRowCount() : 0,
-                    outgoingBatch.getReloadRowCount() > 0 ? outgoingBatch.getDataRowCount() : 0, outgoingBatch.getBatchId(),
+                    outgoingBatch.getLoadMillis(), outgoingBatch.getBatchId(), new Date(), outgoingBatch.isBulkLoaderFlag() == true ? outgoingBatch
+                            .getLoadRowCount() : 0, outgoingBatch.getBatchId(),
                     outgoingBatch.getBatchId(), outgoingBatch.getNodeId(), outgoingBatch.getLoadId(), engine.getNodeId());
         } else {
             String sql = getSql("updateExtractRequestLoadTimeNoParamsInSelect");
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
@@ -1469,6 +1469,9 @@ public class DataExtractorService extends AbstractService implements IDataExtrac
             sql = FormatUtils.replace("rowCount",
                     String.valueOf(outgoingBatch.getReloadRowCount() > 0 ? outgoingBatch.getDataRowCount() : 0), sql);
             sql = FormatUtils.replace("loadMillis", String.valueOf(outgoingBatch.getLoadMillis()), sql);
+            sql = FormatUtils.replace("bulkRowsLoaded",
+                    String.valueOf(outgoingBatch.isBulkLoaderFlag() == true ? (outgoingBatch.getReloadRowCount() > 0 ? outgoingBatch.getDataRowCount() : 0)
+                            : 0), sql);
             transaction.prepareAndExecute(sql, outgoingBatch.getBatchId(), new Date(), outgoingBatch.getBatchId(),
                     outgoingBatch.getBatchId(), outgoingBatch.getNodeId(), outgoingBatch.getLoadId(), engine.getNodeId());
         }
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
@@ -1460,7 +1460,9 @@ public class DataExtractorService extends AbstractService implements IDataExtrac
         if (platform.supportsParametersInSelect()) {
             transaction.prepareAndExecute(getSql("updateExtractRequestLoadTime"), outgoingBatch.getBatchId(), new Date(),
                     outgoingBatch.getReloadRowCount() > 0 ? outgoingBatch.getDataRowCount() : 0,
-                    outgoingBatch.getLoadMillis(), outgoingBatch.getBatchId(), new Date(), outgoingBatch.getBatchId(),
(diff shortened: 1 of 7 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorServiceSqlMap.java
```diff
@@ -64,7 +64,12 @@ public DataExtractorServiceSqlMap(IDatabasePlatform platform,
         
         putSql("updateExtractRequestLoadTime", "update $(extract_request) set loaded_time = (case when end_batch_id = ? then ? when 1 = 0 then last_update_time else null end), "
                 + " loaded_rows = loaded_rows + ?, loaded_millis = loaded_millis + ?, last_loaded_batch_id = ?, "
-                + " last_update_time=? where start_batch_id <= ? and end_batch_id >= ? and node_id=? and load_id=? and source_node_id = ? and loaded_time is null");
+                + " last_update_time=?, "
+                + " bulk_rows_loaded = bulk_rows_loaded + "
+                + " (case when 0 = ? then 0"
+                + " when ? is null then 0"
+                + " else ? end)"
```
erilong: Could this case logic be done in Java instead?  So the SQL would just be:
bulk_rows_loaded = bulk_rows_loaded + ?

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
```diff
@@ -1460,7 +1460,9 @@ public void updateExtractRequestLoadTime(ISqlTransaction transaction, Date loadT
         if (platform.supportsParametersInSelect()) {
             transaction.prepareAndExecute(getSql("updateExtractRequestLoadTime"), outgoingBatch.getBatchId(), new Date(),
                     outgoingBatch.getReloadRowCount() > 0 ? outgoingBatch.getDataRowCount() : 0,
-                    outgoingBatch.getLoadMillis(), outgoingBatch.getBatchId(), new Date(), outgoingBatch.getBatchId(),
+                    outgoingBatch.getLoadMillis(), outgoingBatch.getBatchId(), new Date(), outgoingBatch.isBulkLoaderFlag(),
+                    outgoingBatch.getReloadRowCount() > 0 ? outgoingBatch.getDataRowCount() : 0,
+                    outgoingBatch.getReloadRowCount() > 0 ? outgoingBatch.getDataRowCount() : 0, outgoingBatch.getBatchId(),
```
erilong: Is it necessary to check that getReloadRowCount() > 0?  Isn't that true for every extract request?

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataServiceSqlMap.java
```diff
@@ -156,7 +160,11 @@ public DataServiceSqlMap(IDatabasePlatform platform, Map<String, String> replace
                 + " last_update_time = ?, "
                 + " batch_bulk_load_count = case when ? between start_data_batch_id and end_data_batch_id then batch_bulk_load_count + ? else batch_bulk_load_count end, "
                 + " error_flag = case when error_batch_id = ? then 0 else error_flag end, "
-                + " error_batch_id = case when error_batch_id = ? then null else error_batch_id end "
+                + " error_batch_id = case when error_batch_id = ? then null else error_batch_id end, "
+                + " bulk_row_load_count = (select case"
+                + " when 0 = ? then $(table_reload_status).bulk_row_load_count"
+                + " when sum(loaded_rows) is null then $(table_reload_status).bulk_row_load_count"
+                + " else sum(loaded_rows) + $(table_reload_status).bulk_row_load_count end from $(extract_request) where end_batch_id = ? and source_node_id = ?) "
```
erilong: It looks like it increments bulk_row_load_count only for the last batch of the extract request.  I think that means you wouldn't know if bulk was working until it's done loading all the batches for that table.  Could it work like rows_loaded instead?  It would set bulk_row_load_count to a sum(lbulk_rows_loaded) from extract request.

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataServiceSqlMap.java
```diff
@@ -177,7 +185,11 @@ public DataServiceSqlMap(IDatabasePlatform platform, Map<String, String> replace
                 + " last_update_time = current_timestamp, "
                 + " batch_bulk_load_count = case when $(batchId) between start_data_batch_id and end_data_batch_id then batch_bulk_load_count + $(isBulkLoaded) else batch_bulk_load_count end, "
                 + " error_flag = case when error_batch_id = $(batchId) then 0 else error_flag end, "
-                + " error_batch_id = case when error_batch_id = $(batchId) then null else error_batch_id end "
+                + " error_batch_id = case when error_batch_id = $(batchId) then null else error_batch_id end, "
+                + " (bulk_row_load_count) = (select case"
+                + " when $(isBulkLoaded) = 0 then $(table_reload_status).bulk_row_load_count"
+                + " when sum(loaded_rows) is null then $(table_reload_status).bulk_row_load_count"
+                + " else $(table_reload_status).bulk_row_load_count + sum(loaded_rows) end from $(extract_request) where end_batch_id = $(batchId) and source_node_id = '$(nodeId)') "
```
erilong: Same as previous comment.

erilong (changes requested):

symmetric-core/src/main/resources/symmetric-schema.xml
```diff
@@ -838,8 +839,8 @@
         <column name="sql_message" type="LONGVARCHAR" description="For a request in error, this is the error message that describes the error." />
         <column name="last_update_by" type="VARCHAR" size="50"  description="The user who last updated this entry." />
         <column name="last_update_time" type="TIMESTAMP" required="true" description="Timestamp when a user or the system last updated this entry." />
-        <column name="batch_bulk_load_count" type="BIGINT" required="true" default="0" description="The number of batches that were loaded with the bulk loader."/>
-        
+        <column name="batch_bulk_load_count" type="BIGINT" required="true" default="0" description="The number of batches that were loaded with the bulk loader." />
+        <column name="bulk_row_load_count" type="BIGINT" required="true" default="0" description="The number of rows that were loaded with the bulk loader."/>
```
evan-miller-jumpmind: The column name `bulk_row_load_count` doesn't make much sense to me. Maybe this could be changed to something like `bulk_load_row_count`, `row_bulk_load_count` (to be consistent with `batch_bulk_load_count`), or `rows_bulk_loaded` (to be consistent with `rows_loaded`).

evan-miller-jumpmind (approved):

symmetric-core/src/main/resources/symmetric-schema.xml
mdrouard-jumpmind: Changed to row_bulk_load_count

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataServiceSqlMap.java
mdrouard-jumpmind: Changed the logic in the sql, will need review of latest change to see if it makes sense for all cases. This was tested with postgres -> postgres and it worked as expected.

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorServiceSqlMap.java
mdrouard-jumpmind: This was moved to DataExtractorService

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
mdrouard-jumpmind: This was changed with the SQL logic move to DataExtractorService

erilong (approved):

PR #233, opened 2025-03-12
0006712: Handled variables and BeanShell scripts when searching for an engine file using an engine name

evan-miller-jumpmind:


pavel-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/AbstractParameterService.java
```diff
@@ -249,11 +245,11 @@ public String getEngineName() {
...
-        replacementValues.put("externalId", getExternalId());
-        replacementValues.put("engineName", getEngineName());
-        replacementValues.put("syncUrl", getSyncUrl());
-        replacementValues.put("registrationUrl", getRegistrationUrl());
+        replacementValues.put("nodeGroupId", nodeGroupId);
+        replacementValues.put("externalId", externalId);
+        replacementValues.put("engineName", engineName);
+        replacementValues.put("syncUrl", syncUrl);
+        replacementValues.put("registrationUrl", registrationUrl);
         return replacementValues;
```
erilong, line 253: Can you change these variables back to the getters instead? I think that was intentional because the variables are set lazily.

symmetric-core/src/main/java/org/jumpmind/symmetric/util/PropertiesUtil.java
```diff
@@ -41,14 +43,27 @@ public static String getEnginesDir() {
         return enginesDir;
     }
 
-    public static File findPropertiesFileForEngineWithName(String engineName) {
+    public static File findPropertiesFileForEngineWithName(String engineName, Map<String, String> replacementValues) {
+        if (replacementValues == null) {
```
erilong, line 51: Can you add back the old method signature that takes just an engineName?  That preserves the old API for compatibility and makes calling code a little cleaner.

symmetric-client/src/main/java/org/jumpmind/symmetric/AbstractCommandLauncher.java
```diff
@@ -207,7 +207,7 @@ protected void configurePropertiesFile(CommandLine line) throws ParseException {
                         line.getOptionValue(OPTION_PROPERTIES_FILE));
             }
         } else if (line.hasOption(OPTION_ENGINE)) {
-            propertiesFile = PropertiesUtil.findPropertiesFileForEngineWithName(line.getOptionValue(OPTION_ENGINE));
+            propertiesFile = PropertiesUtil.findPropertiesFileForEngineWithName(line.getOptionValue(OPTION_ENGINE), null);
             if (propertiesFile == null || (propertiesFile != null && !propertiesFile.exists())) {
```
erilong: This line could remove the null argument and use the old API.  See my comment on preserving the old API.

symmetric-core/src/main/java/org/jumpmind/symmetric/util/PropertiesUtil.java
```diff
@@ -41,14 +43,27 @@ public static String getEnginesDir() {
...
+            replacementValues = new HashMap<String, String>();
+        }
+        boolean replacementValuesProvided = !replacementValues.isEmpty();
         File[] files = findEnginePropertiesFiles();
```
erilong, line 55: Avoid the double negative tests and just call this variable something like isEmptyReplacementValues = replacementValues.isEmpty().

erilong (changes requested):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/AbstractParameterService.java
evan-miller-jumpmind: This would cause an infinite loop because the getters call `getReplacementVariables()`.

Before my changes, `substituteScripts()` referenced the variables rather than the getters, so I chose to use the variables so that the behavior of the getters remains the same. Also, the `getReplacementVariables()` method was not being used anywhere before I made my changes, so this change does not affect any other code.

erilong: Yeh, it has to be variables then.  It looks like AbstractSymmetricEngine.init() is calling the getters to initialize them in a predictable way at least.

erilong (approved):

PR #237, opened 2025-03-21
0006777: Add usable disk space properties to runtime stats.

w3bbd3v-jm:


symmetric-client/src/main/java/org/jumpmind/symmetric/util/SnapshotUtil.java
```diff
@@ -445,16 +447,6 @@ public static File createSnapshot(ISymmetricEngine engine, IProgressListener lis
         }
         checkpoint(engine, listener, stepNumber++, totalSteps);
         writeDirectoryStaging(engine, tmpDir);
-        File logDir = LogSummaryAppenderUtils.getLogDir();
```
w3bbd3v-jm, line 448: Pulled the logic to determine the `logDir` into its own static method, and pulled up. This way we determine the `logDir` before we write the runtime stats, which includes its usable space on disk. 

I don't think we need to change any of the `checkpoint`s as the real work seemed to be in `writeDirectoryStaging`.

symmetric-client/src/main/java/org/jumpmind/symmetric/util/SnapshotUtil.java
```diff
@@ -709,6 +702,29 @@ protected static void writeRuntimeStats(ISymmetricEngine engine, File tmpDir) {
...
+        if (!logDir.exists()) {
+            logDir = new File("../logs");
+        }
+        if (!logDir.exists()) {
+            logDir = new File("target");
+        }
+        return logDir;
+    }
+
+    protected static void addUsableDiskSpaceProperties(Properties properties, ISymmetricEngine engine, File tmpDir, File logDir) {
```
w3bbd3v-jm, line 718: Here's an example of the new stats in the `runtime.properties` file. I removed some of the stats so they are together, as they are sorted in alphabetical order normally. 
Screenshot 2025-03-21 at 12 25 10 PM

symmetric-util/src/main/java/org/jumpmind/util/FormatUtils.java
```diff
@@ -158,6 +162,17 @@ public static String formatString(String format, String arg) {
         }
     }
 
+    public static String formatSize(long bytes) {
```
w3bbd3v-jm: Let me know if you know something like this already exists, as we wouldn't want to duplicate. I didn't see anything like it in the code base though.

symmetric-client/src/main/java/org/jumpmind/symmetric/util/SnapshotUtil.java
```diff
@@ -709,6 +702,29 @@ protected static void writeRuntimeStats(ISymmetricEngine engine, File tmpDir) {
         }
     }
 
+    protected static File getOrCreateLogDir() {
```
w3bbd3v-jm, line 704: Copied from removed implementation above.

symmetric-util/src/main/java/org/jumpmind/util/FormatUtils.java
evan-miller-jumpmind: It looks like we're already duplicating code. I found some other methods that are similar to this one, but none of them display 2 digits after the decimal point like your method:

- `UiUtils.formatFileSize()`
- `AbstractDashboardPanel.humanReadableByteCount()`
- `BatchWindow.byteCountToDisplaySize()`
- `NodeWindow.byteCountToDisplaySize()`
- `StagingManager` calls `FileUtils.byteCountToDisplaySize()`

evan-miller-jumpmind (approved):

symmetric-util/src/main/java/org/jumpmind/util/FormatUtils.java
erilong: Sounds like a great refactor opportunity for 3.17 (more time for testing in that release).

erilong (approved):

symmetric-client/src/main/java/org/jumpmind/symmetric/util/SnapshotUtil.java
pavel-jm: Nice!

symmetric-util/src/main/java/org/jumpmind/util/FormatUtils.java
w3bbd3v-jm: Apache `FileUtils` seems to provide something we can use. There is some trivial differences, like it drops the decimal and "B" for bytes becomes just "bytes". I removed the `FormatUtils` changes and am now using that. Thanks for pointing that out! In general, I think less code is better code!

PR #239, opened 2025-03-28
Feature/0006678 support include indexes pg

w3bbd3v-jm:


symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDdlReader.java
```diff
@@ -446,4 +451,109 @@ public PlatformTrigger getPlatformTrigger(IDatabasePlatform platform, Trigger tr
         }
         return platformTrigger;
     }
+
+    @Override
+    protected Collection<IIndex> readIndices(Connection connection, DatabaseMetaDataWrapper metaData,
+            String tableName) throws SQLException {
+        Collection<IIndex> indices = super.readIndices(connection, metaData, tableName);
+        if (metaData.getMetaData().getDatabaseMajorVersion() < 13
```
w3bbd3v-jm, line 490: An area of improvement would be to create new types to support different PG versions. There is a hard-coded version check here and also in the `getPlatformTrigger` method, which existed prior.

evan-miller-jumpmind (approved):

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDdlReader.java
pavel-jm: I agree! Let's rename this class as PostgreSqlDdlReader9 (or PostgreSqlDdlReader95 in a nod to Postgres 9.5).
And add a new descendant class PostgreSqlDdlReader13.
Most of the code will stay in the version 9

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDdlReader.java
```diff
@@ -446,4 +451,109 @@ public PlatformTrigger getPlatformTrigger(IDatabasePlatform platform, Trigger tr
...
+                || indices == null || indices.isEmpty()) {
+            return indices;
+        }
+        String sql = "SELECT\n"
```
pavel-jm: This SQL text is large. To improve readability it can be moved to either class-level constant, data member or even member function.
I'm up for suggestions to hep readability and make this method open to extension to help support changes in future versions of PostgreSQL.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDdlReader.java
```diff
@@ -446,4 +451,109 @@ public PlatformTrigger getPlatformTrigger(IDatabasePlatform platform, Trigger tr
...
+                + "    AND array_length(ix.indkey, 1) > ix.indnkeyatts\n"
+                + "ORDER BY\n"
+                + "    i.relname";
+        try (PreparedStatement ps = connection.prepareStatement(sql)) {
+            ps.setString(1, tableName);
+            try (ResultSet rs = ps.executeQuery()) {
+                while (rs.next()) {
+                    readIndex(indices, rs);
+                }
+            }
```
pavel-jm, line 500: Is there a missing catch{} block?
Customers would likely want to know _why_ something had failed.

symmetric-jdbc/src/test/java/org/jumpmind/db/platform/postgresql/PostgreSqlDdlReaderTest.java
```diff
@@ -1176,6 +1180,78 @@ void testReadTableWithVarcharAndSizeToGetDecimal() throws Exception {
         assertEquals(expectedTable, testTable);
     }
 
+    @Test
```
pavel-jm, line 1183: IF, in a PR review meeting we decide to split PostgreSqlDdlReader into Reader9 and Reader13, let's rename test classes too?

symmetric-db/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDdlBuilder.java
```diff
@@ -656,4 +656,19 @@ public String getSqlType(Column column) {
         }
         return type;
     }
+
+    @Override
```
pavel-jm, line 660: Let's discuss splitting PostgreSqlDdlBuilder into version-specific classes (9 and 13 also) ?

pavel-jm (commented):
Let's schedule a live PR review meeting once Eric is back in office. The fact that some PostgreSQL versions do not support Covering Index is worth discussing.

erilong (approved):

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDdlReader.java
pavel-jm: At the team meeting Eric had brought up the point that more classes is not the best solution at this time.

philipmarzullo64 (approved):

PR #242, opened 2025-03-31
0006781: Prevent database deadlock on sym_sequence table (Router and Initial Load)

pavel-jm:
Database deadlocks on sym_sequence table (MySQL)
Router Service and Initial Load deadlocked

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/SequenceService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/SequenceService.java
@@ -49,6 +49,7 @@ public class SequenceService extends AbstractService implements ISequenceService
                 createSqlReplacementTokens()));
     }
 
+    @Override
     public void init() {
         Map<String, Sequence> sequences = getAll();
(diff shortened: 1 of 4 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/SequenceService.java
```diff
@@ -133,20 +137,30 @@ public Long execute(ISqlTransaction transaction) {
...
+        }
+        long sequenceTimeoutInMs = parameterService.getLong(ParameterConstants.SEQUENCE_TIMEOUT_MS, 5000);
+        long ts = System.currentTimeMillis();
+        int attemptNo = 0;
+        do {
+            attemptNo++;
+            try {
                 long nextVal = tryToGetNextVal(transaction, name, size);
                 if (nextVal > 0) {
+                    if (log.isDebugEnabled()) {
```
w3bbd3v-jm, line 190: Do you want to combine these two conditions in a single if statement?

pavel-jm: This controls logging vs.  "return nextVal;" so they are different.

PR #243, opened 2025-04-01
0006799: Old heartbeat batches syncing old heartbeat time

erilong:
We noticed this on support.  I tracked it back to the issue that caused the regression.  I added a simple unit test to prevent the code mistake from being made again.


pavel-jm (approved):
Wow! Great catch: EXTRACT transform was generated by the getConfigLoadTransforms()

PR #245, opened 2025-04-09
0006757: ExtractDataReader support from target platform

joshahicks:
Log based implementations that use the stream lob setting will fail to extract data during CDC as the ExtractDataReader always uses the platform (which in the case of a log miner setup is H2).   So it fails to find the user table in H2 when it should be looking in the user database.

### DIFF
--- a/symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/reader/ExtractDataReader.java
+++ b/symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/reader/ExtractDataReader.java
@@ -85,7 +85,7 @@ public class ExtractDataReader implements IDataReader {
     public ExtractDataReader(IDatabasePlatform platform, List<IExtractDataReaderSource> sources, IDatabasePlatform targetPlatform) {
         this.sourcesToUse = new ArrayList<IExtractDataReaderSource>(sources);
         this.platform = platform;
-        this.targetPlatform = platform;
+        this.targetPlatform = targetPlatform;
         isSybaseASE = platform.getName().equals(DatabaseNamesConstants.ASE);
     }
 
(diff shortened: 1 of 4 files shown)


symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/reader/ExtractDataReader.java
```diff
@@ -62,25 +63,29 @@ public class ExtractDataReader implements IDataReader {
...
         this.filters = filters;
         this.isUsingUnitypes = isUsingUnitypes;
         this.isSybaseASE = platform.getName().equals(DatabaseNamesConstants.ASE);
     }
 
-    public ExtractDataReader(IDatabasePlatform platform, List<IExtractDataReaderSource> sources) {
+    public ExtractDataReader(IDatabasePlatform platform, List<IExtractDataReaderSource> sources, IDatabasePlatform targetPlatform) {
         this.sourcesToUse = new ArrayList<IExtractDataReaderSource>(sources);
         this.platform = platform;
+        this.targetPlatform = platform;
```
pavel-jm: Should this be 
this.targetPlatform = targetPlatform ;
?

erilong (approved):

symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/reader/ExtractDataReader.java
pavel-jm: Could you re-check the line 88, please?
It looks like it needs the same tweak as line 79.

pavel-jm (changes requested):
Great improvement! See my comment on a 3rd constructor...

pavel-jm (approved):

PR #249, opened 2025-04-14
Feature/0006795 ms sql datetime insight monitor

w3bbd3v-jm:


symmetric-core/src/main/java/org/jumpmind/symmetric/common/MonitorConstants.java
```diff
@@ -72,7 +73,7 @@ public static Map<String, String> getMonitorTypesByVersion() {
         }
         for (String name : new String[] { JVM_64_BIT, JVM_CRASH, JVM_OOM, JVM_THREADS, BLOCK, MYSQL_MODE, NEXT_DATA_IN_GAP, CHANNELS_DISABLED, MAX_BATCH_SIZE,
                 MAX_DATA_TO_ROUTE, MAX_BATCH_TO_SEND, MAX_CHANNELS, CHANNEL_SUSPEND, MISSING_PRIMARY_KEY, CHANNELS_FOREIGN_KEY, JOB_TRENDING, JOB_ERROR,
-                CONNECTION_POOL, CONNECTION_RESET, LOB, STRANDED_OR_EXPIRED_DATA, UNKNOWN_CA }) {
+                CONNECTION_POOL, CONNECTION_RESET, LOB, STRANDED_OR_EXPIRED_DATA, UNKNOWN_CA, INCOMPATIBLE_DATETIME }) {
```
evan-miller-jumpmind: I think this new constant should get added to the map with a version of 3.16.2, not 3.15.0.

evan-miller-jumpmind (changes requested):

symmetric-core/src/main/java/org/jumpmind/symmetric/common/MonitorConstants.java
w3bbd3v-jm: Sure, I'm wondering about the approach. I will add a new map for 3.16.2 for now.

evan-miller-jumpmind: It looks like Eric added this version mapping for [issue 6247](https://issues.symmetricds.org/view.php?id=6247) in order to prevent nodes running newer versions of SymmetricDS from syncing monitors to other nodes running older versions of SymmetricDS which don't support them.

evan-miller-jumpmind (approved):

PR #252, opened 2025-04-15
0006820: Run H2 SHUTDOWN COMPACT script during shutdown hook.

w3bbd3v-jm:


symmetric-jdbc/src/main/java/org/jumpmind/db/platform/h2/H2DatabasePlatform.java
```diff
@@ -134,4 +142,13 @@ public String massageForLimitOffset(String sql, int limit, int offset) {
         }
         return sql + " limit " + limit + " offset " + offset + ";";
     }
+
+    private void shutdownAndCompact() {
+        try {
+            log.info("Shutting down H2 database with COMPACT option");
+            getSqlTemplate().update("SHUTDOWN COMPACT");
+        } catch (SqlException exception) {
+            log.info("Caught an exception during H2 shutdown and compact, with message: " + exception.getMessage());
```
evan-miller-jumpmind: More information (including the exception class name and stack trace) gets written to the log when you pass the exception as the second argument to the `info()` method, like this:

`log.info("Caught an exception during H2 shutdown and compact", exception);`

Or maybe you intentionally called the single-argument `info()` method in order to prevent the log from getting too cluttered.

evan-miller-jumpmind (approved):
This PR includes some changes that aren't related to issue 6820 (additional `@Override` annotations, removal of `(String)` casts), but it's probably fine to keep these.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/h2/H2DatabasePlatform.java
erilong: I could go either way on it, either logging the exception with stack trace, or log the exception class name and its message (sometimes the message is empty).  It's probably going to be an error returned from the driver and database, so the stack trace may not have much value.

log.info("Caught {} during H2 shutdown and compact, with message: {}", exception.getClass().getName(), exception.getMessage());

pavel-jm: Plus, if the error message inside the exception is worth user's attention it is beneficial to user (and/or support team) the WARN log level.
log.warn(...

pavel-jm (approved):

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/h2/H2DatabasePlatform.java
w3bbd3v-jm: Yeah, it was intentional. It is indeed mostly noise from the driver, leading me to see marginal value in the stack trace. I'll push a change to include the class name. My sense is that WARN seems to alarming for that.

w3bbd3v-jm: Here is a common example: 
```INFO [] [H2DatabasePlatform] [Thread-12] Caught org.jumpmind.db.sql.SqlException during H2 shutdown and compact, with message: Database is already closed (to disable automatic closing at VM shutdown, add ";DB_CLOSE_ON_EXIT=FALSE" to the db URL)```

philipmarzullo64 (approved):

PR #254, opened 2025-04-16
0006826: Add a batch.outgoing.tosend.offline.count to runtime-stats in the Support Snapshot

catherinequamme:
Issue: https://issues.symmetricds.org/view.php?id=6826

Note: This PR is related to issue 6831, they should be merged in at the same time. (https://github.com/JumpMind/symmetric-pro/pull/278)

Questions:

1. Is ParameterConstants.OFFLINE_NODE_DETECTION_PERIOD_MINUTES the right parameter to use in this case to detect an offline node?


evan-miller-jumpmind (approved):
I think you're using the correct parameter since `NodeService` already uses that same parameter to check for offline nodes. It is different to the parameter that the web console uses, though, so when pro users take a snapshot, it may not line up with the Manage Nodes screen or the offline node monitors.

erilong (approved):

PR #259, opened 2025-04-18
0006835: Show cancelled load error message to user on load detail screen

erilong:


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
```diff
@@ -768,15 +768,9 @@ public int updateTableReloadRequestsCancelled(long loadId, String sourceNodeId)
...
-        if (e.getCause() instanceof SQLException) {
-            SQLException ex = (SQLException) e.getCause();
-            return sqlTemplate.update(getSql("updateTableReloadStatusError"), ex.getErrorCode(), ex.getSQLState(),
-                    ex.getMessage(), loadId, sourceNodeId);
-        } else {
-            return sqlTemplate.update(getSql("updateTableReloadStatusError"), e.getErrorCode(), null, e.getMessage(),
-                    loadId, sourceNodeId);
-        }
+    protected int updateTableReloadRequestsError(long loadId, String sourceNodeId, int errorCode, String sqlState, String message) {
+        return sqlTemplate.update(getSql("updateTableReloadStatusError"), errorCode, sqlState, message,
```
pavel-jm: IF the length of sqlState exceeds 10 chars, the target column (sql_state) will be unable to accommodate it and throw an error. Actual value comes from vendor-specific JDBC implementation and has no official limit on length of the string.
Since this method is already in the critical path to recovery from a failure, it is worth adding extra code to check sqlState.length() <= 10 and trim it (and log original?)

pavel-jm (approved):
Nice improvement to error reporting. See note about an unchecked length of sqlState ...

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
erilong: Yeh, that makes sense.  I checked in a change to use left 10 characters.

pavel-jm (approved):

PR #261, opened 2025-04-25
0006850: Extract error should be recorded on outgoing batch

erilong:


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
```diff
@@ -953,6 +960,7 @@ protected OutgoingBatch extractOutgoingBatch(ProcessInfo extractInfo, Node targe
                         }
                     }
                 } catch (RuntimeException ex) {
+                    // TODO: update outgoing batch with error
```
w3bbd3v-jm: Do we still need this TODO?

w3bbd3v-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
erilong: Nope, it shouldn't be there.  I had TODO in two places trying to decide which place was better.  Thanks for catching that!

pavel-jm (approved):

PR #264, opened 2025-04-30
0006865: Run sync-triggers on SQL events for create and alter DDL changes.

w3bbd3v-jm:


symmetric-core/src/main/java/org/jumpmind/symmetric/load/ConfigurationChangedDatabaseWriterFilter.java
```diff
@@ -64,6 +66,7 @@ public class ConfigurationChangedDatabaseWriterFilter extends DatabaseWriterFilt
     private static final String CTX_KEY_MY_NODE_SECURITY = "MyNodeSecurity." + SUFFIX;
     private static final String CTX_KEY_CANCEL_LOAD = "CancelLoad." + SUFFIX;
     private static final String CTX_KEY_INITAL_LOAD_ID = "InitialLoadId." + SUFFIX;
+    private static final String CREATE_OR_ALTER_TABLE_PATTERN = " *(alter|create) +.*";
```
evan-miller-jumpmind: It looks like this `CREATE_OR_ALTER_TABLE_PATTERN` will detect when any database object is created or altered. I think you should either narrow this down to only detect when tables are created or altered, or rename it to `CREATE_OR_ALTER_PATTERN` if that isn't the intention.

evan-miller-jumpmind (changes requested):

symmetric-core/src/main/java/org/jumpmind/symmetric/load/ConfigurationChangedDatabaseWriterFilter.java
w3bbd3v-jm: Good catch! I agree -- I'll change the regex to include tables only.

evan-miller-jumpmind (approved):

erilong (approved):

PR #268, opened 2025-05-05
0006867: Made it possible to track engine uninstallation progress

evan-miller-jumpmind:
Roughly 15-20% of the time spent uninstalling an engine is spent dropping triggers. I weighted the progress accordingly, but I can't update the progress as each trigger is dropped without adding an additional database query. I decided that it would be best to avoid running an additional query, so the progress for dropping triggers only updates when all triggers have been dropped.


symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
```diff
@@ -770,52 +773,97 @@ msg, getParameterService().getExternalId(), getParameterService().getNodeGroupId
...
         log.info("Removing SymmetricDS database objects");
-        symmetricDialect.dropTablesAndDatabaseObjects();
+        Database symSchema = ((AbstractSymmetricDialect) symmetricDialect).readSymmetricSchemaFromDatabase();
+        String dropTablesSql = platform.getDdlBuilder().dropTables(symSchema);
+        DatabaseInfo databaseInfo = platform.getDatabaseInfo();
+        int dropStatementsToRunCount = SqlScript.calculateTotalStatements(dropTablesSql,
+                databaseInfo.getSqlCommandDelimiter(), databaseInfo.isTriggersContainJava());
+        SqlScript dropTablesScript = new SqlScript(dropTablesSql, getSqlTemplate(), false, null);
+        dropTablesScript.setListener(new ISqlResultsListener() {
+            @Override
```
pavel-jm, line 866: To minimize method size and make listener reusable, it would be great to move this ( new ISqlResultsListener ...) code into a separate method (or a class?)

Just as an example a static method might make this code easier to follow:
private static ISqlResultsListener generateDropTableListener()

pavel-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/ISymmetricEngine.java
```diff
@@ -77,6 +78,8 @@ public interface ISymmetricEngine {
 
     public void uninstall();
 
+    public void uninstall(ProcessInfo processInfo);
```
w3bbd3v-jm: Do we want to add to the public API or is the existing API sufficient and we delegate to the new method?

evan-miller-jumpmind: I think I could use the existing API to pass the `ProcessInfo` through the `StatisticManager` instead if that would be better.

erilong: I'm still wondering about if IProcessInfoListener is the right way for UI to hook into things, but it's there since 3.13.  The pattern is to pass an IProcessInfoListener.  The process doing work, in this case uninstall() method, would ask the StatisticsManager for a new process info, and set the listener on it.  I guess you could create a new ProcessInfo() without it coming from StatisticsManager.  But it could be useful to see a process entry when uninstall is running.

erilong (changes requested):

erilong (approved):

PR #270, opened 2025-05-06
0006871: Made the Purge Incoming job only purge inactive trigger histories that are older than the oldest create_time in sym_data

evan-miller-jumpmind:


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/PurgeService.java
```diff
@@ -552,8 +553,13 @@ private long purgeRegistrationRequests() {
     private long purgeTriggerHist() {
         Calendar retentionCutoff = Calendar.getInstance();
         retentionCutoff.add(Calendar.MINUTE, -parameterService.getInt(ParameterConstants.PURGE_TRIGGER_HIST_RETENTION_MINUTES));
-        log.info("Purging trigger histories that are inactive and older than {}", fastFormat.format(retentionCutoff.getTime()));
-        long count = sqlTemplate.update(getSql("deleteInactiveTriggerHistSql"), retentionCutoff.getTime());
+        Date retentionCutoffDate = retentionCutoff.getTime();
+        Timestamp minDataCreateTime = sqlTemplateDirty.queryForObject(getSql("minDataCreateTime"), Timestamp.class);
+        if (minDataCreateTime != null && minDataCreateTime.before(retentionCutoffDate)) {
+            retentionCutoffDate = minDataCreateTime;
```
pavel-jm, line 561: An optional suggestion - aiming to reduce potential support calls for purge "Not working properly" .

Log a WARN message in this block to let customers know that purge is skipping trigger history data and this might be a symptom of a bigger problem with unresolved backlog.

I'm also curious what other team members think about the severity of this?

pavel-jm (commented):
This is certainly a nice improvement, despite performance costs involved! 
Just one follow up question regarding logging requirements.

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/PurgeService.java
evan-miller-jumpmind: I agree, it would make sense to add a log message here. I think WARN level would be okay.

w3bbd3v-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/PurgeService.java
evan-miller-jumpmind: I added the log message. If its wording or severity level needs changed, let me know.

PR #272, opened 2025-05-08
0006877: Prevented SQL Server DDL reader from attempting to unescape single quotes within default values

evan-miller-jumpmind:
Calling `unescape()` here can cause certain default values to become invalid. I could not find a reason why `unescape()` should be called here, since the default values come straight from the database. The unit tests for `MsSqlDdlReader` still pass with this change in place. Let me know if there's something I'm missing here.

This is likely a wider issue because `unescape()` is called by 13 other `IDdlReader`s with the same `"'"` and `"''"` arguments. Let me know if I should look into this further.

### DIFF
--- a/symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSqlDdlReader.java
+++ b/symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSqlDdlReader.java
@@ -364,10 +364,10 @@ public class MsSqlDdlReader extends AbstractJdbcDdlReader {
                 if ((column.getScale() == 0) && defaultValue.endsWith(".")) {
                     defaultValue = defaultValue.substring(0, defaultValue.length() - 1);
                 }
-            } else if (TypeMap.isTextType( column.getMappedTypeCode())) {
+            } else if (TypeMap.isTextType(column.getMappedTypeCode())) {
                 String unescapedValue = unescapeTextValue(defaultValue, column);
                 if (log.isTraceEnabled()) {
-                    log.trace("Unescaped default value fro column={}, Original={}, Result={}", column.getName(), defaultValue, unescapedValue);
+                    log.trace("Unescaped default value for column={}, Original={}, Result={}", column.getName(), defaultValue, unescapedValue);
                 }
                 defaultValue = unescapedValue;
             }
--- a/symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSqlDdlReader.java
+++ b/symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSqlDdlReader.java
@@ -364,10 +364,32 @@ public class MsSqlDdlReader extends AbstractJdbcDdlReader {
                 if ((column.getScale() == 0) && defaultValue.endsWith(".")) {
                     defaultValue = defaultValue.substring(0, defaultValue.length() - 1);
                 }
-            } else if (TypeMap.isTextType(column.getMappedTypeCode()) && defaultValue.startsWith("N'") && defaultValue.endsWith("'")) {
-                defaultValue = defaultValue.substring(2, defaultValue.length() - 1);
+            } else if (TypeMap.isTextType( column.getMappedTypeCode())) {
+                String unescapedValue = unescapeTextValue(defaultValue, column);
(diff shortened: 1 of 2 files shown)


symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSqlDdlReader.java
```diff
@@ -364,11 +364,8 @@ protected void parseDefaultValue(Column column) {
                 if ((column.getScale() == 0) && defaultValue.endsWith(".")) {
                     defaultValue = defaultValue.substring(0, defaultValue.length() - 1);
                 }
-            } else if (TypeMap.isTextType(column.getMappedTypeCode())) {
-                if (defaultValue.startsWith("N'") && defaultValue.endsWith("'")) {
-                    defaultValue = defaultValue.substring(2, defaultValue.length() - 1);
-                }
-                defaultValue = unescape(defaultValue, "'", "''");
+            } else if (TypeMap.isTextType(column.getMappedTypeCode()) && defaultValue.startsWith("N'") && defaultValue.endsWith("'")) {
+                defaultValue = defaultValue.substring(2, defaultValue.length() - 1);
```
pavel-jm: My concern this will fail to strip quotes from the non-N types of text data. In SQL Server VARCHAR and NVARCHAR are handled differently:
'some_string' for VARCHAR
N'some_string' for NVARCHAR

My guess is that it is helpful to check for the closing quote:
    defaultValue.endsWith("'")

However, a separate IF is needed to check for the presence of the Unicode prefix "N" to remove it:
  defaultValue.startsWith("N'")

evan-miller-jumpmind: The original behavior was to strip the outer quotes and `N` character from N types, but to leave the outer quotes there for non-N types. I don't see how my change would affect this behavior. The only effect is that `unescape()` no longer gets called, so instances of `''` no longer get replaced with `'`.

If there's a problem with how it handles stripping the outer quotes, I could create a separate issue and take a look.

erilong (changes requested):
Could you add a unit test to prove the issue that fails is now fixed?  Maybe add a couple unit tests around '' and N'' to show that it's working as expected also.

erilong (approved):

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSqlDdlReader.java
pavel-jm: Yes, and removing unescape() call is introducing a problem. 
Here's a proposed test (it fails on my laptop):

    @Test
    @DisplayName("Parse Complex Default Value1")
    void testParseDefaultValue1() throws Exception {
        MockDbDataSource mockDataSource = new MockDbDataSource(MsSqlDatabasePlatform_VERSION10);
        String originalDefault = "'John''s'";
        String expectedDefault = "John's";
        int mappedTypeCode = Types.VARCHAR;
        MsSqlDdlReader testReader = createMsSqlDdlReader(mockDataSource);
        Column column = new Column();
        column.setDefaultValue(originalDefault);
        column.setMappedTypeCode(mappedTypeCode);
        testReader.parseDefaultValue(column);
        assertEquals(expectedDefault, column.getDefaultValue());
    }

pavel-jm (changes requested):
I'm available to meet and test this out together, when you have some time.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSqlDdlReader.java
evan-miller-jumpmind: I don't think my change introduced this problem. Without my change, that test still fails with a result of `'John's'`.

pavel-jm: I've added one change - to break out the text unwrapping into a separate method. Also added few tests for it. Please review.

evan-miller-jumpmind: Your change looks good to me.

PR #277, opened 2025-05-13
0006872: Added a JSON transform type

evan-miller-jumpmind:


symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/transform/JsonColumnTransform.java
```diff
@@ -0,0 +1,90 @@
...
+import org.jumpmind.db.platform.IDatabasePlatform;
+import org.jumpmind.extension.IBuiltInExtensionPoint;
+import org.jumpmind.symmetric.io.data.CsvData;
+import org.jumpmind.symmetric.io.data.CsvUtils;
+import org.jumpmind.symmetric.io.data.DataContext;
+
+import com.google.gson.Gson;
+
+public class JsonColumnTransform implements ISingleNewAndOldValueColumnTransform, IBuiltInExtensionPoint {
+    public final static String NAME = "json";
```
w3bbd3v-jm, line 17: Since we return this in the public API, can/should we make this private? Are the other constants public because of the unit tests?

evan-miller-jumpmind: All of our `ColumnTransform` classes have a `public final static String NAME` which is referenced by other classes such as `TransformService` and our pro UI classes for editing transforms. I was just following this pattern. I agree that it would make sense to make `NAME` private and make `getName()` static. If we do that, it would probably be best to create a separate issue for this and to make this change to all of our `ColumnTransform` classes.

The other constants are public because they are referenced by `JsonColumnTransformTest` and one of our Pro UI classes (see the pro PR).

erilong (approved):

pavel-jm (approved):

PR #282, opened 2025-05-20
0006862: Remove trailing slash from sync URL

catherinequamme:
Issue: https://issues.symmetricds.org/view.php?id=6862

This change prevents issues with starting up when a user changes their sync URL in the engine file to have a trailing '/' or '\'. 

If the user adds a trailing slash in the deploy wizard however, this does not resolve that. I can add a way to resolve that issue though by adjusting the code in EndpointScreen.java though. Let me know if I should do so!


pavel-jm (approved):
Thank you for improving this and helping customers be successful!

catherinequamme:
If the user adds a trailing slash in the deploy wizard however, this does not resolve that. I can add a way to resolve that issue though by adjusting the code in EndpointScreen.java though. Should I do that as well?

PR #285, opened 2025-05-28
0006903: Made the log miner honor triggers' sync_on_insert/update/delete_conditions using similar expressions to the column match router

evan-miller-jumpmind:
There are a couple of places where I moved some code to allow it to be shared between the column match router and the log miner:
- `ColumnMatchDataRouter.Expression` (and some methods) -> `org.jumpmind.symmetric.model.ColumnMatchExpression`
  - This resulted in some of the tests in `ColumnMatchDataRouterTest` depending on `ColumnMatchExpression` instead of `ColumnMatchDataRouter`. Let me know if I should create a new `ColumnMatchExpressionTest` class for these tests.
- `AbstractDataRouter.getDataMap()`, `getDataObjectMap()`, and dependent methods -> `SymmetricUtils`
  - It wasn't completely necessary to move `getDataObjectMap()` and its dependent methods, but I thought it would be best for these to be in the same place as `getDataMap()`.

Let me know if any of that needs changed.

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/model/ColumnMatchExpression.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/model/ColumnMatchExpression.java
@@ -84,8 +84,8 @@ public class ColumnMatchExpression {
     public static List<ColumnMatchExpression> parse(String expression) throws SyntaxParsingException {
         List<ColumnMatchExpression> expressions = new ArrayList<ColumnMatchExpression>();
         if (!StringUtils.isBlank(expression)) {
-            String[] operators = { NOT_EQUALS, EQUALS, NOT_CONTAINS, CONTAINS, HAS, NOT_HAS, STARTS_WITH,
-                    NOT_STARTS_WITH, ENDS_WITH, NOT_ENDS_WITH };
+            String[] operators = { NOT_EQUALS, EQUALS, NOT_CONTAINS, CONTAINS, NOT_HAS, HAS, NOT_STARTS_WITH,
+                    STARTS_WITH, NOT_ENDS_WITH, ENDS_WITH };
             String[] expTokens = expression.split("\\s*(\\s+or|\\s+OR)?(\r\n|\r|\n)(or\\s+|OR\\s+)?\\s*" +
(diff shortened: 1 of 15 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/model/ColumnMatchExpression.java
```diff
@@ -0,0 +1,252 @@
...
+        } else if (operator.equals(NOT_STARTS_WITH)) {
+            hasNotStartsWith = true;
+        } else if (operator.equals(ENDS_WITH)) {
+            hasEndsWith = true;
+        } else if (operator.equals(NOT_ENDS_WITH)) {
+            hasNotEndsWith = true;
+        }
+    }
+
+    public static List<ColumnMatchExpression> parse(String expression) throws SyntaxParsingException {
```
w3bbd3v-jm, line 84: FWIW, the new methods seem straight forward enough, with simple I/O (and not much mocking, if any), to be good  candidates for unit tests.

evan-miller-jumpmind: The methods in `ColumnMatchExpression` aren't new. I just moved them over from `ColumnMatchDataRouter` and renamed a couple of them. They are already covered by `ColumnMatchDataRouterTest`.

w3bbd3v-jm: I'd vote to move them over. That way we don't need to worry about it later when `ColumnMatchDataRouter` changes.

evan-miller-jumpmind: That makes sense, I'll create a `ColumnMatchExpressionTest` and have it test the `parse()` and `run()` methods.

w3bbd3v-jm: Thanks Evan

erilong (approved):

PR #288, opened 2025-06-02
0006116: Create node import config, channels overwritten

erilong:


erilong:
I moved the creation of the node's identity before configuration, cluster, and sequence services are initialized.  I reviewed their init methods and it looks like there is no dependency on node identity (except for configuration service, of course).

I put target of 3.16.4, so don't merge this one in yet until after 3.16.3 is released, so we have more time to test and use it.

philipmarzullo64 (approved):

pavel-jm (approved):

PR #290, opened 2025-06-03
0006915: Added a parameter for the maximum outgoing form post limit

evan-miller-jumpmind:


symmetric-core/src/main/java/org/jumpmind/symmetric/transport/http/HttpTransportManager.java
```diff
@@ -113,19 +115,48 @@ public int sendAcknowledgement(Node remote, List<IncomingBatch> list, Node local
...
+                maxFormKeys = 100000;
+            }
+            for (int i = 0; i < backOffPostCount && maxFormKeys > 1; i++) {
+                maxFormKeys /= 2;
+            }
+            int maxByteSize = engine.getParameterService().getInt(ParameterConstants.TRANSPORT_MAX_BYTES_TO_SYNC);
+            for (String data : getAcknowledgementData(remote.requires13Compatiblity(), local.getNodeId(), list, maxFormKeys, maxByteSize)) {
+                log.debug("Sending ack: {}", data);
+                statusCode = sendMessage("ack", remote, local, data, securityToken, requestProperties, registrationUrl);
+                if (statusCode != HttpConnection.HTTP_OK) {
```
evan-miller-jumpmind, line 131: Eric suggested removing this check now that I removed the early `return`, but I think this check is still necessary in order to break out of the loop when there's an error.

symmetric-core/src/main/resources/symmetric-default.properties
```diff
@@ -588,6 +588,14 @@ transport.type=http
 # Type: integer
 transport.max.bytes.to.sync=104857600
 
+# This is the maximum number of form keys to synchronize in one connect.
+# A value of 0 or below won't limit the number of form keys.
+#
+# DatabaseOverridable: true
+# Tags: transport
+# Type: integer
+transport.max.form.keys=99900
```
erilong: Shouldn't this be 100,000 to match the spring boot setting?

erilong (approved):

symmetric-core/src/main/resources/symmetric-default.properties
evan-miller-jumpmind: When we discussed this issue, we decided that the default for this new parameter should be a bit lower than the Spring Boot setting. I could increase it to 100,000 if that would be better.

erilong: It will stay within that limit, right?  Yeh, then the settings can match.

pavel-jm (approved):

PR #293, opened 2025-06-09
0006918: In Configure -> Parameters show that parameter changed in engine file

catherinequamme:
Issue: https://issues.symmetricds.org/view.php?id=6918

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/model/DatabaseParameter.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/model/DatabaseParameter.java
@@ -86,7 +86,7 @@ public class DatabaseParameter implements Serializable {
         this.nodeGroupId = nodeGroupId;
     }
 
-    public boolean getInEngineFile() {
+    public boolean isInEngineFile() {
         return inEngineFile;
     }
 
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/model/DatabaseParameter.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/model/DatabaseParameter.java
@@ -30,6 +30,7 @@ public class DatabaseParameter implements Serializable {
     private String value;
     private String externalId = ParameterConstants.ALL;
     private String nodeGroupId = ParameterConstants.ALL;
+    private boolean inEngineFile = false;
 
     public DatabaseParameter() {
     }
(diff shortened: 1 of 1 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/model/DatabaseParameter.java
```diff
@@ -76,4 +85,12 @@ public String getNodeGroupId() {
     public void setNodeGroupId(String nodeGroupId) {
         this.nodeGroupId = nodeGroupId;
     }
+
+    public boolean getInEngineFile() {
```
evan-miller-jumpmind: It would make the code a little easier to read if this method was called `isInEngineFile()` instead.

evan-miller-jumpmind (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/model/DatabaseParameter.java
catherinequamme: Agreed, just wanted to follow the pattern in the class. I can change it though!

evan-miller-jumpmind: That's because the other variables are all `String`s. Usually `boolean` getters should start with "is". You might already know this, but you can have Eclipse generate getters and setters for you if you right click the class name and then click Source -> Generate Getters and Setters.

symmetric-core/src/main/java/org/jumpmind/symmetric/model/DatabaseParameter.java
```diff
@@ -45,6 +46,14 @@ public DatabaseParameter(String key, String value, String externalId, String nod
         this.nodeGroupId = nodeGroupId;
     }
 
+    public DatabaseParameter(String key, String value, String externalId, String nodeGroupId, boolean inEngineFile) {
+        this.key = key;
+        this.value = value;
+        this.externalId = externalId;
+        this.nodeGroupId = nodeGroupId;
+        this.inEngineFile = inEngineFile;
```
erilong, line 54: Since the ParameterService doesn't need this inEngineFile property, let's create a new model object for use by the ParameterEditPanel.  The class could be on the Pro side and it could extend DatabaseParameter.  If it's only going to be used by the screen, it could be an inner class.

erilong (changes requested):

catherinequamme:
Closed, created inner class on Pro side.

PR #294, opened 2025-06-09
0006928: Guard against null value for FileModifiedTime in FileSyncService.acknowledgeFiles

pavel-jm:
0006928: Guard against null value for FileModifiedTime in FileSyncService.acknowledgeFiles


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/FileSyncService.java
```diff
@@ -121,6 +121,7 @@ public FileSyncService(ISymmetricEngine engine) {
         setSqlMap(new FileSyncServiceSqlMap(platform, createSqlReplacementTokens()));
```
pavel-jm, line 121: The main change is in the acknowledgeFiles() method. The rest is due to automatic formatting.

erilong (approved):

w3bbd3v-jm (approved):

PR #298, opened 2025-06-13
0006939: Changed log messages from info level to warn level and documented Too Many Form Keys insight

evan-miller-jumpmind:
I had to change these log messages to `warn` level because `LogSummaryAppender` doesn't track `info`-level log messages.

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/transport/http/HttpTransportManager.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/transport/http/HttpTransportManager.java
@@ -136,10 +136,10 @@ public class HttpTransportManager extends AbstractTransportManager implements IT
                         } else {
                             backOffPostCount++;
                             if (maxFormKeys > FORM_KEYS_PER_BATCH) {
-                                log.info("Ack received a {} response from node {}. The form key limit will be reduced from {} to {} during the next attempt.",
+                                log.warn("Ack received a {} response from node {}. The form key limit will be reduced from {} to {} during the next attempt.",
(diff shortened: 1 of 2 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/transport/http/HttpTransportManager.java
```diff
@@ -136,10 +136,10 @@ public int sendAcknowledgement(Node remote, List<IncomingBatch> list, Node local
                         } else {
                             backOffPostCount++;
                             if (maxFormKeys > FORM_KEYS_PER_BATCH) {
-                                log.info("Ack received a {} response from node {}. The form key limit will be reduced from {} to {} during the next attempt.",
+                                log.warn("Ack received a {} response from node {}. The form key limit will be reduced from {} to {} during the next attempt.",
                                         statusCode, remote.getNodeId(), maxFormKeys, Math.max(maxFormKeys / 2, FORM_KEYS_PER_BATCH));
                             } else {
-                                log.info("Ack received a {} response from node {}. A form key limit of 50000 will take effect during the next attempt.",
+                                log.warn("Ack received a {} response from node {}. A form key limit of 50000 will take effect during the next attempt.",
```
pavel-jm: This functionality is a great addition and messaging helps system administrator.
To simplify code maintenance I recommend replacing the magic number (50000) with the actual value of a constant (or parameter), which controls this IF block.
You know - one less thing to update whenever defaults get modified in a future release some years from now.

pavel-jm (changes requested):

pavel-jm (approved):

PR #301, opened 2025-06-16
0006946: Add overrides for isBlob and isClob for AseDatabasePlatform

catherinequamme:
Issue: https://issues.symmetricds.org/view.php?id=6946

Fix for 3.15

### DIFF
--- a/symmetric-jdbc/src/main/java/org/jumpmind/db/platform/ase/AseDatabasePlatform.java
+++ b/symmetric-jdbc/src/main/java/org/jumpmind/db/platform/ase/AseDatabasePlatform.java
@@ -54,6 +54,7 @@ import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.codec.binary.Hex;
 import org.apache.commons.lang3.StringUtils;
 import org.jumpmind.db.model.Column;
+import org.jumpmind.db.model.ColumnTypes;
 import org.jumpmind.db.platform.AbstractJdbcDatabasePlatform;
 import org.jumpmind.db.platform.DatabaseNamesConstants;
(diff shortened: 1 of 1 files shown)


symmetric-jdbc/src/main/java/org/jumpmind/db/platform/ase/AseDatabasePlatform.java
```diff
@@ -143,6 +144,21 @@ public boolean canColumnBeUsedInWhereClause(Column column) {
...
+    @Override
+    public boolean isBlob(Column column) {
+        int type = column.getJdbcTypeCode();
+        if (settings.isTreatBinaryAsLob()) {
+            return type == Types.BLOB || type == Types.BINARY || type == Types.VARBINARY || type == Types.LONGVARBINARY || type == -10;
+        }
+        return type == Types.BLOB || type == Types.LONGVARBINARY || type == -10;
+    }
+
     @Override
```
pavel-jm, line 162: Thank you for spending time to reproduce this issue locally!
Let's share an "urgent" patch JAR with the customer (who is running the 3.15.16 version) and 
then spend time to add a unit test class (AseDatabasePlatformTest) and a unit test targeting the canColumnBeUsedInWhereClause method.

The objective is to reproduce the Production issue in unit test (with the fix commented-out) and then certify the fix is working as expected.
You can use the H2DatabasePlatformTest as a guide for both file location and implementation.

Feel free to direct any questions to me or Andy (author of the H2DatabasePlatformTest)!

pavel-jm (changes requested):

pavel-jm (approved):

PR #306, opened 2025-06-20
0006953: Support capture for SAP Hana

w3bbd3v-jm:


symmetric-client/src/main/java/org/jumpmind/symmetric/db/hana/HanaSymmetricDialect.java
evan-miller-jumpmind, line 1: I don't understand these changes - can you explain them? It looks like `HanaSymmetricDialect.getTransactionTriggerExpression()` no longer overrides `AbstractSymmetricDialect.getTransactionTriggerExpression()` and one of its parameters now uses a different `Trigger` model object. I don't see where `HanaSymmetricDialect.getTransactionTriggerExpression()` is called apart from the unit test.

w3bbd3v-jm: I haven't explored the diff yet, but the Trigger object should not change. I'll address this and review the diffs.

evan-miller-jumpmind (approved):

evan-miller-jumpmind (approved):

erilong (approved):

PR #310, opened 2025-06-30
0006971: Prevent aggressive purging of the sym_console_table_stats and sym_node_host_stats table

catherinequamme:
Issue: https://issues.symmetricds.org/view.php?id=6971

Pro PR: https://github.com/JumpMind/symmetric-pro/pull/324

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/ParameterService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/ParameterService.java
@@ -49,7 +49,6 @@ public class ParameterService extends AbstractParameterService implements IParam
     private ISqlTemplate sqlTemplate;
     private Date lastUpdateTime;
     private List<DatabaseParameter> offlineParameters;
-    private final int MINS_IN_ONE_WEEK = 10080;
 
     public ParameterService(IDatabasePlatform platform, ITypedPropertiesFactory factory, String tablePrefix) {
         this.tablePrefix = tablePrefix;
@@ -86,9 +85,6 @@ public class ParameterService extends AbstractParameterService implements IParam
 
     public void saveParameter(String externalId, String nodeGroupId, String key, Object paramValue, String lastUpdateBy) {
         paramValue = paramValue != null ? paramValue.toString() : null;
-        if (key.equals(ParameterConstants.PURGE_STATS_RETENTION_MINUTES) && Integer.parseInt((String) paramValue) < MINS_IN_ONE_WEEK) {
-            paramValue = MINS_IN_ONE_WEEK;
-        }
         if (extensionService != null) {
             for (IParameterSaveFilter filter : extensionService.getExtensionPointList(IParameterSaveFilter.class)) {
                 paramValue = filter.filterSaveParameter(key, (String) paramValue);
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/ParameterService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/ParameterService.java
@@ -49,6 +49,7 @@ public class ParameterService extends AbstractParameterService implements IParam
     private ISqlTemplate sqlTemplate;
     private Date lastUpdateTime;
     private List<DatabaseParameter> offlineParameters;
+    private final int MINS_IN_ONE_WEEK = 10080;
 
     public ParameterService(IDatabasePlatform platform, ITypedPropertiesFactory factory, String tablePrefix) {
         this.tablePrefix = tablePrefix;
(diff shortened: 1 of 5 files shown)


evan-miller-jumpmind (commented):
I was expecting the code changes to be in `PurgeService` and `ConsolePurgeListener` where the value of `purge.stats.retention.minutes` is retrieved, but I think I prefer your approach. When the user tries to set the parameter below 10,080 via the web console, this approach will make it clearer that it has a minimum value. Also, if we start using this parameter in more places, then we won't have to remember to write code to enforce the minimum value.

One thing I noticed is that there are two other `reload()` methods in `TypedPropertiesFactory` that are not affected by this change. Should they call `limitPurgeStatsRetentionMinutesMinimum()` too?

I also noticed that in `PurgeServiceTest.setUp()`, the parameter is set to 1,440, which is below the minimum value. Should this be increased to 10,080?

catherinequamme:
> I was expecting the code changes to be in `PurgeService` and `ConsolePurgeListener` where the value of `purge.stats.retention.minutes` is retrieved, but I think I prefer your approach. When the user tries to set the parameter below 10,080 via the web console, this approach will make it clearer that it has a minimum value. Also, if we start using this parameter in more places, then we won't have to remember to write code to enforce the minimum value.
> 
> One thing I noticed is that there are two other `reload()` methods in `TypedPropertiesFactory` that are not affected by this change. Should they call `limitPurgeStatsRetentionMinutesMinimum()` too?
> 
> I also noticed that in `PurgeServiceTest.setUp()`, the parameter is set to 1,440, which is below the minimum value. Should this be increased to 10,080?



> I was expecting the code changes to be in `PurgeService` and `ConsolePurgeListener` where the value of `purge.stats.retention.minutes` is retrieved, but I think I prefer your approach. When the user tries to set the parameter below 10,080 via the web console, this approach will make it clearer that it has a minimum value. Also, if we start using this parameter in more places, then we won't have to remember to write code to enforce the minimum value.
> 
> One thing I noticed is that there are two other `reload()` methods in `TypedPropertiesFactory` that are not affected by this change. Should they call `limitPurgeStatsRetentionMinutesMinimum()` too?
> 
> I also noticed that in `PurgeServiceTest.setUp()`, the parameter is set to 1,440, which is below the minimum value. Should this be increased to 10,080?

I saw the test class too and wasn't sure if we should change it there as well. Couldn't hurt if this will be the new minimum. As for the other 2 reload methods in TypedPropertiesFactory, I will add the change to both to be safe.

evan-miller-jumpmind (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/ParameterService.java
```diff
@@ -85,6 +86,9 @@ public void saveParameter(String key, Object paramValue, String lastUpdateBy) {
 
     public void saveParameter(String externalId, String nodeGroupId, String key, Object paramValue, String lastUpdateBy) {
         paramValue = paramValue != null ? paramValue.toString() : null;
+        if (key.equals(ParameterConstants.PURGE_STATS_RETENTION_MINUTES) && Integer.parseInt((String) paramValue) < MINS_IN_ONE_WEEK) {
+            paramValue = MINS_IN_ONE_WEEK;
+        }
```
erilong: User can still update parameters through the database or properties, so I don't think there should be code in this class trying to filter values.

symmetric-core/src/main/java/org/jumpmind/symmetric/util/TypedPropertiesFactory.java
```diff
@@ -56,22 +59,32 @@ public TypedProperties reload() {
...
             throw new IoException(e);
         }
     }
 
+    private void limitPurgeStatsRetentionMinutesMinimum(TypedProperties p) {
+        String purgeStatsMinutes = p.get(ParameterConstants.PURGE_STATS_RETENTION_MINUTES);
+        if (purgeStatsMinutes != null && Integer.valueOf(purgeStatsMinutes) < MINS_IN_ONE_WEEK) {
+            p.setProperty(ParameterConstants.PURGE_STATS_RETENTION_MINUTES, MINS_IN_ONE_WEEK);
+        }
+    }
```
erilong: The properties factory is meant to be generic and unaware of the actual properties and values.  I think this is the wrong place for this code and it should be removed.

erilong (changes requested):
I think it would more straight-forward to modify PurgeService and enforce the minimum value there:

`retentionCutoff.add(Calendar.MINUTE, -Integer.max(parameterService.getInt(ParameterConstants.PURGE_STATS_RETENTION_MINUTES), 10080));`

That is similar to the pattern where code is in-line with a default value for a parameter.

catherinequamme:
> I think it would more straight-forward to modify PurgeService and enforce the minimum value there:
> 
> `retentionCutoff.add(Calendar.MINUTE, -Integer.max(parameterService.getInt(ParameterConstants.PURGE_STATS_RETENTION_MINUTES), 10080));`
> 
> That is similar to the pattern where code is in-line with a default value for a parameter.

Should I also enforce the value in purgeTableStats() in ConsolePurgeListener.java?

erilong (approved):

PR #317, opened 2025-07-02
0006985: Sybase ASE Unitype Fixes

catherinequamme:
The 3.16 version of this PR: https://github.com/JumpMind/symmetric-ds/pull/316#event-18419755334

Issue: https://issues.symmetricds.org/view.php?id=6985

### DIFF
--- a/symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/reader/ExtractDataReader.java
+++ b/symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/reader/ExtractDataReader.java
@@ -47,6 +47,8 @@ import org.jumpmind.symmetric.io.data.IDataReader;
 import org.jumpmind.util.CollectionUtils;
 import org.jumpmind.util.FormatUtils;
 import org.jumpmind.util.Statistics;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ExtractDataReader implements IDataReader {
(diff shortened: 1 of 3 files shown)


symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/reader/ExtractDataReader.java
```diff
@@ -281,40 +284,30 @@ protected CsvData convertUtf16toUTF8(Table table, CsvData data) {
...
-                            rowData[index] = utf8String;
+                            if (!baseString.startsWith("fffe")) {
+                                baseString = "fffe" + baseString;
+                            }
+                            byte[] utf16Bytes = Hex.decodeHex(baseString);
+                            String utf16Str = new String(utf16Bytes, "UTF-16");
+                            String utf8Str = new String(utf16Str.getBytes("UTF-8"), "UTF-8");
+                            rowData[index] = utf8Str;
+                        } catch (UnsupportedEncodingException | DecoderException e) {
+                            log.warn("Failed to decode UTF-16 to UTF-8 for column '{}' with value '{}': {}",
```
pavel-jm: A warning message is a great addition! It will help identify data issues more quickly.
Let's add full name of the table to make this message more detailed?

pavel-jm (changes requested):
Thank you for figuring this out!
I also like the new WARN message. It will be more helpful with the full table name in it though.

pavel-jm (approved):

PR #322, opened 2025-07-11
0006990: Set the Postgres DatabaseInfo's triggersCreateOrReplaceSupported property to true

evan-miller-jumpmind:
The issue description says to set the property in the DDL builder. I tried that but ran into some issues because the `databaseMajorVersion` comes from the SQL template, which is initialized after the DDL builder in the `AbstractJdbcDatabasePlatform` constructor.

I noticed that `PostgreSqlSymmetricDialect` already has a `versionSupportsReplaceTriggers` variable, so I set the property in its constructor instead. I had to move `versionSupportsReplaceTriggers` above the trigger template because it needs to be set before the `PostgreSqlTriggerTemplate` constructor calls `getCreateTriggerCommandBeginning()`.


philipmarzullo64 (approved):

pavel-jm (approved):
Well done!

PR #323, opened 2025-07-11
0006922: Removed DDL trigger when uninstalling a node

evan-miller-jumpmind:
Even though `trigger.capture.ddl.changes` is a startup parameter, setting it to false in `sym_parameter` still takes effect and overrides the engine file. Then, when syncing triggers, `TriggerRouterService.updateOrCreateDdlTriggers()` removes the DDL trigger.

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
@@ -848,10 +848,8 @@ abstract public class AbstractSymmetricEngine implements ISymmetricEngine {
             }
             processInfo.incrementCurrentDataCount();
             if (platform.readTableFromDatabase(null, null, TableConstants.getTableName(prefix, TableConstants.SYM_LOCK)) != null) {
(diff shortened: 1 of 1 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
```diff
@@ -848,6 +848,10 @@ public synchronized void uninstall(IProcessInfoListener listener) {
             }
             processInfo.incrementCurrentDataCount();
             if (platform.readTableFromDatabase(null, null, TableConstants.getTableName(prefix, TableConstants.SYM_LOCK)) != null) {
+                if (parameterService.is(ParameterConstants.TRIGGER_CAPTURE_DDL_CHANGES)) {
```
pavel-jm: Excellent catch! Could you add one more parameter to this action, please:
postgres.trigger.capture.truncate.event
This risks leaving a shared function behind (on PostgreSQL only) which is used for truncate-level triggers.

pavel-jm (changes requested):
Excellent catch! Could you add one more parameter to this action, please:
postgres.trigger.capture.truncate.event

pavel-jm (approved):

PR #331, opened 2025-07-22
0007000: Cached the lists of table names in each catalog/schema in the DDL reader

evan-miller-jumpmind:
I created new `org.jumpmind.cache` packages in `symmetric-jdbc` for `ObjectDefinitionCache` and `ObjectDefinitionCacheTest`. Let me know if there's a better place for these classes.

### DIFF
--- a/symmetric-jdbc/src/main/java/org/jumpmind/cache/ObjectDefinitionCache.java
+++ b/symmetric-jdbc/src/main/java/org/jumpmind/cache/ObjectDefinitionCache.java
@@ -75,8 +75,10 @@ public class ObjectDefinitionCache {
     }
 
     public void clearTableNameCache() {
-        tableNameCache.clear();
-        tableNameCacheTime = System.currentTimeMillis();
+        synchronized (tableNameCacheLock) {
+            tableNameCache.clear();
+            tableNameCacheTime = System.currentTimeMillis();
+        }
     }
 
     private class TableNameCacheKey {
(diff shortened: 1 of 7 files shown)


philipmarzullo64 (approved):

symmetric-jdbc/src/main/java/org/jumpmind/cache/ObjectDefinitionCache.java
```diff
@@ -0,0 +1,134 @@
...
+    private AbstractJdbcDdlReader ddlReader;
+    private Object tableNameCacheLock = new Object();
+    volatile private Map<TableNameCacheKey, List<String>> tableNameCache = new HashMap<TableNameCacheKey, List<String>>();
+    volatile private long tableNameCacheTime;
+
+    public ObjectDefinitionCache(AbstractJdbcDdlReader ddlReader) {
+        this.ddlReader = ddlReader;
+    }
+
+    public List<String> getTableNames(String catalog, String schema, String[] tableTypes) {
```
w3bbd3v-jm: I see the use of `String` catalog and schema in a lot of places. Wondering if this can be improved by using the `CatalogSchema` model. That way clients of the API won't need to remember the order of the Strings in the API signature. Just an idea.

evan-miller-jumpmind: I didn't know about the `CatalogSchema` model. That makes sense, I'll use that instead of `String`s.

PR #334, opened 2025-07-31
0007008: Allowed other threads to access the node channel cache while it's being refreshed

evan-miller-jumpmind:
I used `OutgoingBatchCache` as an example when making this change. One difference is that if the cache is null, `acquire()` is called in order to wait for the lock and avoid returning a null or empty result. This should only happen shortly after startup.

Because every cache within `ConfigurationCache` shares the `configurationCacheLock`, I also made similar changes to the channel, queue, node group link, and node group channel window caches.

Once this PR is approved, I'll backport the changes to 3.15 to resolve issue 7007.

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ConfigurationCache.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ConfigurationCache.java
@@ -26,6 +26,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.Semaphore;
 
 import org.jumpmind.symmetric.ISymmetricEngine;
 import org.jumpmind.symmetric.common.ParameterConstants;
@@ -35,11 +36,14 @@ import org.jumpmind.symmetric.model.NodeGroupChannelWindow;
 import org.jumpmind.symmetric.model.NodeGroupLink;
 import org.jumpmind.symmetric.service.IConfigurationService;
 import org.jumpmind.symmetric.service.IParameterService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ConfigurationCache {
+    private final Logger log = LoggerFactory.getLogger(ConfigurationCache.class);
     private IParameterService parameterService;
     private IConfigurationService configurationService;
-    private Object configurationCacheLock = new Object();
+    private Semaphore configurationCacheLock = new Semaphore(1);
     volatile private Map<String, List<NodeChannel>> nodeChannelCache;
     volatile private Map<String, Channel> channelsCache;
     volatile private Collection<String> queuesCache;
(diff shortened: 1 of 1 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ConfigurationCache.java
```diff
@@ -56,24 +60,59 @@ public ConfigurationCache(ISymmetricEngine engine) {
...
                 }
-                if (nodeId != null) {
-                    nodeChannels = configurationService.getNodeChannelsFromDb(nodeId);
-                    nodeChannelCache.put(nodeId, nodeChannels);
-                } else {
-                    nodeChannels = new ArrayList<NodeChannel>(0);
+            } else if (configurationCacheLock.tryAcquire()) {
+                try {
+                    populateNodeChannelCache(nodeId, true, channelCacheTimeoutInMs);
+                } finally {
```
pavel-jm, line 101: populateNodeChannelCache calls getNodeChannelsFromDb, therefore is vulnerable to a SqlException.
Since nodeChannelCache != null we have an option to catch (and log!) this exception and use current value. This should make product more resilient.

symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ConfigurationCache.java
```diff
@@ -82,81 +121,180 @@ public long getNodeChannelCacheTime() {
...
+                    } finally {
+                        configurationCacheLock.release();
+                    }
+                } catch (InterruptedException e) {
+                    log.error("Failed to retrieve node group channel windows", e);
+                }
+            } else if (configurationCacheLock.tryAcquire()) {
+                try {
+                    populateNodeGroupChannelWindowCache(channelCacheTimeoutInMs);
+                } finally {
```
pavel-jm, line 272: The populateNodeGroupChannelWindowCache method calls getNodeGroupChannelWindowsFromDb, therefore it is vulnerable to a SqlException.
Since channelWindowsByChannelCache != null here, we have an option to catch (and log!) this exception and return the current, albeit older value. This should improve overall product resiliency.

symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ConfigurationCache.java
```diff
@@ -82,81 +121,180 @@ public long getNodeChannelCacheTime() {
...
+                    } finally {
+                        configurationCacheLock.release();
+                    }
+                } catch (InterruptedException e) {
+                    log.error("Failed to retrieve node group links", e);
+                }
+            } else if (configurationCacheLock.tryAcquire()) {
+                try {
+                    populateNodeGroupLinkCache(cacheTimeoutInMs);
+                } finally {
```
pavel-jm, line 222: The populateNodeGroupLinkCache method calls getNodeGroupLinksFromDb, therefore it is vulnerable to a SqlException.
Since nodeGroupLinksCache != null here, we have an option to catch (and log!) this exception and return the current, albeit older value?

symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ConfigurationCache.java
```diff
@@ -82,81 +121,180 @@ public long getNodeChannelCacheTime() {
...
                 .getLong(ParameterConstants.CACHE_TIMEOUT_NODE_GROUP_LINK_IN_MS);
-        if (System.currentTimeMillis() - nodeGroupLinkCacheTime >= cacheTimeoutInMs
-                || nodeGroupLinksCache == null || refreshCache) {
-            synchronized (configurationCacheLock) {
-                if (System.currentTimeMillis() - nodeGroupLinkCacheTime >= cacheTimeoutInMs
-                        || nodeGroupLinksCache == null || refreshCache) {
-                    nodeGroupLinksCache = configurationService.getNodeGroupLinksFromDb();
-                    nodeGroupLinkCacheTime = System.currentTimeMillis();
+        if (System.currentTimeMillis() - nodeGroupLinkCacheTime >= cacheTimeoutInMs || nodeGroupLinksCache == null || refreshCache) {
+            if (nodeGroupLinksCache == null) {
```
pavel-jm: I believe the refreshCache parameter is intended to force database call and should be included in the IF statement along with
nodeGroupLinksCache == null || refreshCache

symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ConfigurationCache.java
```diff
@@ -82,81 +121,180 @@ public long getNodeChannelCacheTime() {
...
                     }
-                    queuesCache = queues;
-                    channelCacheTime = System.currentTimeMillis();
+                } catch (InterruptedException e) {
+                    log.error("Failed to retrieve channels", e);
+                }
+            } else if (configurationCacheLock.tryAcquire()) {
+                try {
+                    populateChannelCache(channelCacheTimeoutInMs);
+                } finally {
```
pavel-jm, line 166: The populateChannelCache method calls getChannelsFromDb, therefore it is vulnerable to a SqlException.
Since channelsCache != null we have an option to catch (and log!) this exception and return the current, albeit older value. This should make product more resilient.

pavel-jm (changes requested):
Excellent work! 
A few catch blocks will make product more resilient to database errors/timeouts.

pavel-jm (approved):

philipmarzullo64 (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ConfigurationCache.java
```diff
@@ -35,11 +37,14 @@
 import org.jumpmind.symmetric.model.NodeGroupLink;
 import org.jumpmind.symmetric.service.IConfigurationService;
 import org.jumpmind.symmetric.service.IParameterService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ConfigurationCache {
+    private final Logger log = LoggerFactory.getLogger(ConfigurationCache.class);
```
w3bbd3v-jm: Is there a good reason why the  LOGGER is not static?

symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ConfigurationCache.java
```diff
@@ -82,81 +126,192 @@ public long getNodeChannelCacheTime() {
...
+                    log.error("Failed to retrieve channels", e);
+                } finally {
+                    configurationCacheLock.release();
                 }
             }
         }
     }
 
+    protected void populateChannelCache(long channelCacheTimeoutInMs) throws SqlException {
+        long ts = System.currentTimeMillis();
```
w3bbd3v-jm: Can we rename `ts` to something like `startTime`?

w3bbd3v-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ConfigurationCache.java
evan-miller-jumpmind: I used `OutgoingBatchCache` as an example and its logger isn't static. I don't think making it static would hurt anything though.

pavel-jm (approved):

PR #338, opened 2025-08-12
0007018 ParameterService refactor (unique per engine) and new AbstractColumnTransform

pavel-jm:
dependency

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/io/data/transform/BshColumnTransform.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/io/data/transform/BshColumnTransform.java
@@ -35,7 +35,6 @@ import java.util.Map;
 
 import org.apache.commons.lang3.StringUtils;
 import org.jumpmind.db.platform.IDatabasePlatform;
-import org.jumpmind.extension.IBuiltInExtensionPoint;
 import org.jumpmind.symmetric.SymmetricException;
 import org.jumpmind.symmetric.common.Constants;
 import org.jumpmind.symmetric.common.ParameterConstants;
@@ -50,7 +49,8 @@ import org.slf4j.LoggerFactory;
 import bsh.Interpreter;
 import bsh.TargetError;
 
-public class BshColumnTransform implements ISingleNewAndOldValueColumnTransform, IBuiltInExtensionPoint {
+public class BshColumnTransform extends AbstractColumnTransform implements ISingleNewAndOldValueColumnTransform 
+{
     protected final Logger log = LoggerFactory.getLogger(getClass());
     final String INTERPRETER_KEY = String.format("%d.BshInterpreter", hashCode());
     public static final String NAME = "bsh";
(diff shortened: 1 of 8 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/io/data/transform/JavaColumnTransform.java
```diff
@@ -51,18 +51,27 @@ public void setExtensionService(IExtensionService extensionService) {
...
 
+    @Override
     public String getName() {
         return NAME;
     }
 
+    @Override
+    public boolean isParameterServiceRequired() {
+        return true;
+    }
```
evan-miller-jumpmind: `JavaColumnTransform` doesn't use `ParameterService`, so it doesn't need to override `isParameterServiceRequired()`.

symmetric-core/src/main/java/org/jumpmind/symmetric/io/data/transform/LookupColumnTransform.java
```diff
@@ -38,23 +37,32 @@
...
 
+    @Override
     public String getName() {
         return NAME;
     }
 
+    @Override
+    public boolean isParameterServiceRequired() {
+        return false;
+    }
```
evan-miller-jumpmind: Can you remove this override of `isParameterServiceRequired()` since it's just a copy of the parent implementation?

symmetric-core/src/main/java/org/jumpmind/symmetric/io/data/transform/VariableColumnTransform.java
```diff
@@ -58,14 +57,22 @@ public class VariableColumnTransform implements ISingleNewAndOldValueColumnTrans
...
 
+    @Override
     public String getName() {
         return NAME;
     }
 
+    @Override
+    public boolean isParameterServiceRequired() {
+        return false;
+    }
```
evan-miller-jumpmind: Can you remove this override of `isParameterServiceRequired()` since it's just a copy of the parent implementation?

evan-miller-jumpmind (changes requested):
What made you decide to have `JavaColumnTransform`, `LookupColumnTransform`, and `VariableColumnTransform` extend `AbstractColumnTransform`? If these 3 classes extend `AbstractColumnTransform`, then shouldn't every other `*ColumnTransform` class extend `AbstractColumnTransform` too?

symmetric-core/src/main/java/org/jumpmind/symmetric/io/data/transform/BshColumnTransform.java
```diff
@@ -50,7 +49,8 @@
 import bsh.Interpreter;
 import bsh.TargetError;
 
-public class BshColumnTransform implements ISingleNewAndOldValueColumnTransform, IBuiltInExtensionPoint {
+public class BshColumnTransform extends AbstractColumnTransform implements ISingleNewAndOldValueColumnTransform 
+{
```
erilong, line 53: I think the IBuiltInExtensionPoint interface is needed on all these transforms because they are built-in extensions, not ones written by user.  Without that interface, it will start logging about these extension points and they show up on the About -> Extensions screen as provided by the user.

erilong (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/io/data/transform/LookupColumnTransform.java
pavel-jm: Thanks - Updated!

symmetric-core/src/main/java/org/jumpmind/symmetric/io/data/transform/VariableColumnTransform.java
pavel-jm: Thanks - Updated!

evan-miller-jumpmind (approved):
I now understand why not all `*ColumnTransform` classes extend `AbstractColumnTransform`. I didn't notice that the rest of them were in the `symmetric-io` package.

PR #340, opened 2025-08-20
0007024: fix enum for Turkey locale.

w3bbd3v-jm:


pavel-jm (approved):

erilong (approved):

evan-miller-jumpmind (changes requested):
Prior to your change, the result of `properties.get(ParameterConstants.DBDIALECT_ORACLE_JDBC_LOB_HANDLING, SqlTemplateSettings.JdbcLobHandling.PLAIN.name())` was converted to uppercase. Now only the result of `SqlTemplateSettings.JdbcLobHandling.PLAIN.name()` gets converted to uppercase.

1. This will make the `oracle.jdbc.lob.handling` parameter case sensitive when it was previously case insensitive. If any users have this parameter set to a lowercase value (which is what this parameter's description in the User Guide suggests), then this change will cause them to get an `IllegalArgumentException`. Is that an intended effect?
2. If the above effect is intentional, then is there any reason to call `FormatUtils.upper()` here? Doesn't `SqlTemplateSettings.JdbcLobHandling.PLAIN.name()` already return `"PLAIN"` in uppercase?

w3bbd3v-jm:
> Prior to your change, the result of `properties.get(ParameterConstants.DBDIALECT_ORACLE_JDBC_LOB_HANDLING, SqlTemplateSettings.JdbcLobHandling.PLAIN.name())` was converted to uppercase. Now only the result of `SqlTemplateSettings.JdbcLobHandling.PLAIN.name()` gets converted to uppercase.
> 
>     1. This will make the `oracle.jdbc.lob.handling` parameter case sensitive when it was previously case insensitive. If any users have this parameter set to a lowercase value (which is what this parameter's description in the User Guide suggests), then this change will cause them to get an `IllegalArgumentException`. Is that an intended effect?
> 
>     2. If the above effect is intentional, then is there any reason to call `FormatUtils.upper()` here? Doesn't `SqlTemplateSettings.JdbcLobHandling.PLAIN.name()` already return `"PLAIN"` in uppercase?

You're right Evan. Thanks for catching that. I think we can do even better by wrapping what is returned from the properties.get call with FormatUtils.upper(). That way if someone sets the parameter to "plain" the issue does not reappear. Working on a fix...

evan-miller-jumpmind (approved):

PR #341, opened 2025-08-20
Retrofit NVARCHAR(MAX) fix 0006621 into 3.16

pavel-jm:


w3bbd3v-jm (approved):

catherinequamme (approved):

symmetric-client/src/main/java/org/jumpmind/symmetric/db/mssql/MsSql2005TriggerTemplate.java
```diff
@@ -20,10 +20,30 @@
...
 import org.jumpmind.symmetric.db.ISymmetricDialect;
 
 public class MsSql2005TriggerTemplate extends MsSqlTriggerTemplate {
     public MsSql2005TriggerTemplate(ISymmetricDialect symmetricDialect) {
         super(symmetricDialect);
     }
-}
+
+    /***
+     * Treats NVARCHAR(MAX), VARCHAR(MAX) columns as regular string type, not as "large objects".
```
w3bbd3v-jm, line 35: Should we reference the superclass method's Javadoc? So something like this? 
```
    /**
     * {@inheritDoc} 
     *
     * Treats NVARCHAR(MAX), VARCHAR(MAX) columns as regular string type, not as "large objects". 
     */
```

symmetric-client/src/main/java/org/jumpmind/symmetric/db/mssql/MsSql2005TriggerTemplate.java
```diff
@@ -20,10 +20,30 @@
...
+     */
+    @Override
+    protected boolean isLob(Column column) {
+        int mappedTypeCode = column.getMappedTypeCode();
+        int size = column.getSizeAsInt();
+        String jdbcTypeName = column.getJdbcTypeName();
+        if (mappedTypeCode == Types.LONGVARCHAR
+                && (jdbcTypeName.equalsIgnoreCase("NVARCHAR") && size > MsSql2005DdlBuilder.NVARCHARMAX_LIMIT
```
w3bbd3v-jm, line 43: Should we reverse this to be `"NVARCHAR".equalsIgnoreCase(jdbcTypeName)` and `"VARCHAR".equalsIgnoreCase(jdbcTypeName)` just in case `jdbcTypeName` is `NULL`?

philipmarzullo64 (approved):

PR #342, opened 2025-08-21
0007028: Apply transforms to oldValue to help remote system find the record to update or delete

pavel-jm:
Do NOT approve PR just yet - more changes are coming!

### DIFF
--- a/symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/transform/IsBlankTransform.java
+++ b/symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/transform/IsBlankTransform.java
@@ -62,8 +62,6 @@ public class IsBlankTransform implements ISingleNewAndOldValueColumnTransform, I
         if (StringUtils.isBlank(oldValue)) {
             result.setOldValue(expression);
         }
-        log.warn("Done. column={}, old={}, transformedOld={}, new={}, transformedNew={}", column.getSourceColumnName(), oldValue, result.oldValue, newValue,
-                result.newValue);
         return result;
     }
 }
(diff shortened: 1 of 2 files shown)


symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/transform/IsBlankTransform.java
```diff
@@ -49,13 +51,16 @@ public boolean isLoadColumnTransform() {
...
         if (StringUtils.isBlank(newValue)) {
-            String expression = column.getTransformExpression();
-            if (StringUtils.isEmpty(expression)) {
-                expression = null;
-            }
-            result = new NewAndOldValue(expression, oldValue);
+            result.setNewValue(expression);
+        }
+        if (StringUtils.isBlank(oldValue)) {
+            result.setOldValue(expression);
```
erilong: By default, the data loader only loads columns that are changing (dataloader.apply.changes.only=true), so by setting the old and new values to the same value, the column will be ignored.  The introduction of NewAndOldValue was supposed to be backwards compatible, which is why only newer transforms are using it.

pavel-jm: I agree with compatibility, but differ on one point:
The only way Old value and New value are the same if 
StringUtils.isBlank(newValue) == StringUtils.isBlank(oldValue)

Otherwise values will be different and this is an improvement over existing code, which sends OldValue of NULL (bypassing this transform) to the system, which cannot accept NULLs due to table structure being different from the source database.

philipmarzullo64 (approved):

pavel-jm:
A approach was discussed in a team meeting. A new branch will be created to implement it.

PR #344, opened 2025-08-26
0007033: Prevented leading ZWNBSP character from getting lost when replicating unitype data from Sybase ASE

evan-miller-jumpmind:
Once this is merged, I'll make the same change on the 3.16 branch to fix issue 7034.

### DIFF
--- a/symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/reader/ExtractDataReader.java
+++ b/symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/reader/ExtractDataReader.java
@@ -294,9 +294,7 @@ public class ExtractDataReader implements IDataReader {
                     if (index >= 0 && rowData[index] != null) {
                         try {
                             String baseString = rowData[index];
-                            if (!baseString.startsWith("fffe")) {
-                                baseString = "fffe" + baseString;
-                            }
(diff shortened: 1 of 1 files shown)


catherinequamme (approved):

symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/reader/ExtractDataReader.java
```diff
@@ -294,9 +294,7 @@ protected CsvData convertUtf16toUTF8(Table table, CsvData data) {
                     if (index >= 0 && rowData[index] != null) {
                         try {
                             String baseString = rowData[index];
-                            if (!baseString.startsWith("fffe")) {
-                                baseString = "fffe" + baseString;
-                            }
+                            baseString = "fffe" + baseString;
```
pavel-jm, line 297: Just curious - what happens when the baseString is null or empty?

evan-miller-jumpmind: There's a null check on line 294 above. If it's empty, it looks like it still needs `"fffe"` added as a prefix before it can be decoded.

pavel-jm (approved):

PR #345, opened 2025-08-27
0007028: add transform unit tests.

w3bbd3v-jm:


symmetric-io/src/test/java/org/jumpmind/symmetric/io/data/transform/IsNullTransformTest.java
```diff
@@ -0,0 +1,154 @@
...
+    @Test
+    void testTransform_withDelete_returnsExpressionWhenOldValueIsNull() throws IgnoreColumnException, IgnoreRowException {
+        String expression = "expression";
+        testTransformDelete(Expected.of(NewValue.of(expression), null), NewValue.of("new"), OldValue.of(null), TransformExpression.of(expression));
+    }
+
+    @Test
+    void testTransform_withDelete_returnsEmptyWhenOldValueAndExpressionAreNull() throws IgnoreColumnException, IgnoreRowException {
+        testTransformDelete(Expected.of(NewValue.of(""), null), NewValue.of("new"), OldValue.of(null), TransformExpression.of(null));
+    }
```
evan-miller-jumpmind, line 65: Why do these delete tests expect `Expected.of(NewValue.of(*), null)` rather than `Expected.of(null, OldValue.of(*))`?

symmetric-io/src/test/java/org/jumpmind/symmetric/io/data/transform/IsNullTransformTest.java
```diff
@@ -0,0 +1,154 @@
...
+    @Test
+    void testTransform_withUpdate_returnsExpressionWhenNewValueIsNull() throws IgnoreColumnException, IgnoreRowException {
+        String expression = "expression";
+        testTransformUpdate(Expected.of(NewValue.of(expression), null), NewValue.of(null), OldValue.of("old"), TransformExpression.of(expression));
+    }
+
+    @Test
+    void testTransform_withUpdate_returnsEmptyWhenNewValueAndExpressionAreNull() throws IgnoreColumnException, IgnoreRowException {
+        testTransformUpdate(Expected.of(NewValue.of(""), null), NewValue.of(null), OldValue.of("old"), TransformExpression.of(null));
+    }
```
evan-miller-jumpmind, line 82: Why do these update tests expect the old value to be null when it was "old" before the transform was applied?

evan-miller-jumpmind (changes requested):
Most of the delete tests expect the new value to not be null. Deletes only have an old value, so shouldn't a delete transform always return null for the new value?

In `IsBlankTransformTest` and `IsEmptyTransformTest`, the tests for deletes seem to only cover the cases when the new value is blank/empty/null. Should they cover the cases when the old value is blank/empty/null instead?

The tests for updates seem to only cover the cases when the new value is blank/empty/null. Should they also cover the cases when the old value is blank/empty/null?

The insert tests in `IsBlankTransformTest` and `IsEmptyTransformTest` expect the old value to not be null. Inserts only have a new value, so shouldn't an insert transform always return null for the old value?

As far as I know, SymmetricDS doesn't support transforms for create events (see `TransformWriter.isTransformable()`). Are the create tests necessary?

symmetric-io/src/test/java/org/jumpmind/symmetric/io/data/transform/IsNullTransformTest.java
w3bbd3v-jm: This is because the expected `NewAndOldValue` return object uses the oldValue on a DELETE event type as the newValue. The oldValue is not always set on the return object. That logic lives in the `NewAndOldValue` class though so I'm not testing that logic here. Let me know if you want to discuss. To your point, I think I should add another test for DELETE though that passes `null` as the `newValue` input. 
I'll also add the Expected object is only for asserting the result, not for `transform` method input. If that is not clear, I can improve the readability there if you have ideas.

evan-miller-jumpmind (approved):

pavel-jm (approved):

PR #346, opened 2025-08-28
0007035: Caught IllegalArgumentException when sorting process info

evan-miller-jumpmind:
See the [issue tracker](https://issues.symmetricds.org/view.php?id=7035) for the `IllegalArgumentException` and its stack trace.

I didn't notice anything wrong with `ProcessInfo.compareTo()`, so I wrote a quick test to `sort()` a `List` containing `ProcessInfo`s with various `status`es and `startTime`s and found that I couldn't reproduce the error. I concluded that the `IllegalArgumentException` was occurring because one or more of the `ProcessInfo`s had their `status` modified during the `sort()`.

I looked through the callers of `StatisticManager.getProcessInfos()` and I didn't notice any that would experience any major issues if the `List` was out of order, so I decided to just catch the `IllegalArgumentException` and return the unsorted `List`.

Some other options for the `catch` block would be to return a sorted deep copy of `processInfos.values()` instead of the unsorted shallow copy, or to try calling `sort()` again. Let me know if either of those would be better.

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/statistic/StatisticManager.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/statistic/StatisticManager.java
@@ -125,7 +125,11 @@ public class StatisticManager implements IStatisticManager {
 
     public List<ProcessInfo> getProcessInfos() {
         List<ProcessInfo> list = new ArrayList<ProcessInfo>(processInfos.values());
-        Collections.sort(list);
+        try {
+            Collections.sort(list);
+        } catch (IllegalArgumentException e) {
+            log.warn("Failed to sort process infos", e);
+        }
         return list;
     }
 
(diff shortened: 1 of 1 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/statistic/StatisticManager.java
```diff
@@ -125,7 +125,11 @@ public Set<String> getNodesWithProcessesInError() {
 
     public List<ProcessInfo> getProcessInfos() {
         List<ProcessInfo> list = new ArrayList<ProcessInfo>(processInfos.values());
-        Collections.sort(list);
+        try {
+            Collections.sort(list);
+        } catch (IllegalArgumentException e) {
+            log.warn("Failed to sort process infos", e);
```
pavel-jm, line 131: Returning unsorted list without caller being aware violates method's contract. 
A deep copy would be preferred.
Another option is to identify how why list items are changing and make a deep copy somewhere "upstream".

pavel-jm (changes requested):
Let's brainstorm on how to avoid returning an unsorted list?

symmetric-core/src/main/java/org/jumpmind/symmetric/statistic/StatisticManager.java
evan-miller-jumpmind: The `List` items are changing because other processes update their `ProcessInfo` status as they make progress. I made a change to create a deep copy of the `List` when the `IllegalArgumentException` occurs, which makes the `List` safe to sort.

pavel-jm (approved):

PR #351, opened 2025-09-05
0007032: Added a getMaxDataIdAlreadyRouted() method to RouterService to support the Activity Dashboard panel

evan-miller-jumpmind:


pavel-jm (approved):

erilong (approved):

w3bbd3v-jm (approved):

evan-miller-jumpmind:
This change probably won't be necessary. The purpose of this change was to add a new method to `RouterService` so that the Activity dashboard panel could call it and use its result to query `sym_data`. However, we determined that the `sym_data` query is too slow and that it needs replaced.

PR #354, opened 2025-09-08
7040: SQL Server sql_variant data type support

philipmarzullo64:


erilong (approved):
Looks good.  Should we add sql_variant as a supported type in mssqlserver.ad?

PR #362, opened 2025-09-09
0007044: Added data_received and data_bytes_received columns to sym_node_host_channel_stats

evan-miller-jumpmind:


pavel-jm (approved):

catherinequamme (changes requested):
The changes overall look good to me, but I think we should fix the 2 broken tests to make sure this code change doesn't cause any other tests to fail.

evan-miller-jumpmind:
I merged the fix for issue 7057 and now the tests are all passing.

pavel-jm (approved):

PR #382, opened 2025-09-17
0007061: Fixed $(sourceExternalId) and $(targetExternalId) variables not getting replaced when included in a router's target catalog or schema

evan-miller-jumpmind:
I looked through the history of `ExtractDataReader` and it looks like the `substituteVariables()` method has never replaced these variables since it was added in 3.7.0. Here is the commit:

https://github.com/jumpmindinc/symmetric-ds/commit/503a394f460f0e5bbe6ddc86badef5ba46908a46

They do get replaced as expected in `ColumnsAccordingToTriggerHistory` where it calls `SymmetricUtils.replaceNodeVariables()`.


pavel-jm (approved):
Thank you!!!

PR #387, opened 2025-09-19
0007070: Recorded the number of gaps and unrouted rows to sym_node_host_stats

evan-miller-jumpmind:
I thought of 3 options to populate these new columns:

1. Call `RouterService.getUnroutedDataCount()` and `DataService.countDataGaps()` on every `StatisticManager.flush()`
2. Only update the stats when `RouterService.getUnroutedDataCount()`, `DataService.countDataGaps()`, or `DataService.findDataGaps(false)` is called
3. Start with option 2 and fall back to option 1 if there are no stats

I didn't go with option 1 because it would cause many unnecessary database queries. I didn't go with option 2 because it's very uncommon for the open-source version to call `RouterService.getUnroutedDataCount()`, so the `data_unrouted_count` would be left null. I figured option 3 was a good compromise.

I decided to remove the check for `HostStats.isNonZero()` because even when all other stats are 0, the data gap and unrouted data stats are still available.

Because `StatisticManager` now needs to access `RouterService` but `AbstractSymmetricEngine.init()` calls `createStatisticManager()` before `buildRouterService()`, I had to change the constructor to pass in the `ISymmetricEngine` rather than the individual services.

As usual, let me know if any of this needs changed.

### DIFF
--- a/symmetric-client/src/main/java/org/jumpmind/symmetric/ClientSymmetricEngine.java
+++ b/symmetric-client/src/main/java/org/jumpmind/symmetric/ClientSymmetricEngine.java
@@ -435,8 +435,7 @@ public class ClientSymmetricEngine extends AbstractSymmetricEngine {
                 throw new RuntimeException(e);
             }
         }
-        return new StatisticManager(parameterService, nodeService,
-                configurationService, statisticService, clusterService);
+        return new StatisticManager(this);
     }
 
     protected static void waitForAvailableDatabase(DataSource dataSource) {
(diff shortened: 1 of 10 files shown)


w3bbd3v-jm:
Did this introduce a failing unit test? 
DataServiceTest > testFindDataGaps2() FAILED
    java.lang.NullPointerException at DataServiceTest.java:143

evan-miller-jumpmind:
Yes. I pushed a change that should fix the `NullPointerException`.

w3bbd3v-jm (approved):

pavel-jm (approved):

PR #388, opened 2025-09-23
0007068: implement github release workflow

w3bbd3v-jm:


symmetric-assemble/build.gradle
```diff
@@ -39,16 +37,6 @@ task develop {
     dependsOn tasks.cleanEclipse, tasks.eclipse, tasks.cleanIdea, tasks.idea, tasks.compileTestJava
 }
 
-github {
```
w3bbd3v-jm, line 42: This github gradle plugin is no longer needed. Instead we are replacing it with steps in the release workflow.

symmetric-assemble/common.gradle
```diff
@@ -238,6 +238,7 @@ subprojects { subproject ->
 		}
 		
 		signing {
+    		useInMemoryPgpKeys(findProperty("signingKey"), findProperty("signingPassword"))
```
w3bbd3v-jm, line 241: Allows us to continue to use our signing key in the GitHub release workflow.

pavel-jm (approved):

PR #395, opened 2025-09-26
SYM-7031: Add support for AWS SQS (load-only endpoint)

catherinequamme:
Ticket: https://issues.symmetricds.org/view.php?id=7031


w3bbd3v-jm:
Last I heard we were targeting release/3.17 branch. Did that change?

evan-miller-jumpmind (commented):
Would it be better to put the parameters in `ConsoleConstants`  and `symmetric-console-default.properties` instead since AWS SQS is a pro feature?

pavel-jm:
> Would it be better to put the parameters in `ConsoleConstants` and `symmetric-console-default.properties` instead since AWS SQS is a pro feature?

I agree that the AWS SQS is a feature specific to the PRO edition.

philipmarzullo64:
Not using this issue nor pull requests.

New JIRA issue:
https://jumpmind.atlassian.net/browse/SYM-7459

New pull requests:
https://github.com/jumpmindinc/symmetric-pro/pull/661
https://github.com/jumpmindinc/symmetric-ds/pull/662

PR #397, opened 2025-09-26
0007068: update GitHub release for 3.15

w3bbd3v-jm:


.github/workflows/release.yml
```diff
@@ -50,7 +50,7 @@ jobs:
             
             - name: Publish
               run: |
-                ./gradlew clean jar releaseSymmetric deploy publishDocker publishDockerLatest  \
```
w3bbd3v-jm, line 53: Instead of having different release workflows for different release branches, we can handle this programmatically. That effort will be a separate task. In the meantime, removing the task in `release/3.15` branch.

pavel-jm (approved):

PR #401, opened 2025-09-30
0007084: Check at startup trigger.capture.ddl.check.trigger.hist

pavel-jm:
0007084: Check trigger.capture.ddl.check.trigger.hist parameter (to detect change in value) at startup

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/common/ParameterConstants.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/common/ParameterConstants.java
@@ -448,7 +448,8 @@ final public class ParameterConstants {
     public final static String POSTGRES_SECURITY_DEFINER = "postgres.security.definer";
     public final static String POSTGRES_CONVERT_INFINITY_DATE_TO_NULL = "postgres.convert.infinity.date.to.null";
     public final static String[] STARTUP_DB_OBJECTS_SETUP_PARAMS = new String[] { TRIGGER_CAPTURE_DDL_CHANGES,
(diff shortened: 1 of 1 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/common/ParameterConstants.java
```diff
@@ -448,7 +448,8 @@ private ParameterConstants() {
     public final static String POSTGRES_SECURITY_DEFINER = "postgres.security.definer";
     public final static String POSTGRES_CONVERT_INFINITY_DATE_TO_NULL = "postgres.convert.infinity.date.to.null";
     public final static String[] STARTUP_DB_OBJECTS_SETUP_PARAMS = new String[] { TRIGGER_CAPTURE_DDL_CHANGES,
-            POSTGRES_TRIGGER_CAPTURE_TRUNCATE, DB_USER, "db.url", "target.db.url" };
+            POSTGRES_TRIGGER_CAPTURE_TRUNCATE, TRIGGER_CAPTURE_DDL_CHECK_TRIGGER_HIST, TRIGGER_CAPTURE_DDL_DELIMITER,
+            DB_USER, "db.url", "target.db.url" };
```
pavel-jm, line 452: Adding new parameters to this list will force the to call AbstractSymmetricEngine.setupDatabase(force=true), which sets up shared database objects (including functions) and rebuilds triggers.

w3bbd3v-jm (approved):

erilong (approved):

PR #403, opened 2025-10-01
0007086: Fixed registration failure when a console role exists with a role ID that comes before its parent role ID alphabetically

evan-miller-jumpmind:


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
```diff
@@ -273,18 +273,22 @@ public void extractConfigurationStandalone(Node targetNode, Writer writer, Strin
...
-                Table table = symmetricDialect.getPlatform().getTableFromCache(triggerHistory.getSourceCatalogName(), triggerHistory.getSourceSchemaName(),
-                        triggerHistory.getSourceTableName(), false);
-                String initialLoadSql = "1=1 order by ";
-                String quote = platform.getDdlBuilder().getDatabaseInfo().getDelimiterToken();
-                Column[] pkColumns = table.getPrimaryKeyColumns();
-                for (int j = 0; j < pkColumns.length; j++) {
-                    if (j > 0) {
-                        initialLoadSql += ", ";
+                String tableName = triggerRouter.getTrigger().getSourceTableName();
+                if (!tableName.endsWith(TableConstants.SYM_NODE_IDENTITY)) {
```
pavel-jm, line 277: Thank you finding and fixing the issue!
Could I ask you for a small favor?
Move the code for building SQL into a separate (reusable + testable) utility method, please.

pavel-jm (changes requested):

pavel-jm (approved):

PR #410, opened 2025-10-17
0007092: Added a target_node_id column to sym_node_channel_ctl

evan-miller-jumpmind:


symmetric-core/src/main/java/org/jumpmind/symmetric/model/ChannelMap.java
```diff
@@ -34,58 +35,101 @@
...
+    public String getIgnoreChannelsAsString(String targetNodeId) {
+        return getChannelsAsString(CHANNELS_IGNORE, targetNodeId);
     }
 
-    public String getIgnoreChannelsAsString() {
-        return StringUtils.join(map.get(CHANNELS_IGNORE), ',');
+    private String getChannelsAsString(String key, String targetNodeId) {
+        Set<String> channelIdSet = new HashSet<String>();
+        Map<String, Set<String>> includedChannelMap = map.get(key);
+        Map<String, Set<String>> excludedChannelMap = map.get(key.equals(CHANNELS_SUSPEND) ? CHANNELS_UNSUSPEND : CHANNELS_UNIGNORE);
```
pavel-jm: In order to be more defensive about potential future problem, it is best to write any assumptions made about the "key" argument to this method in code? 
Let's guard against values not matching those assumptions, like... null, CHANNELS_UNIGNORE, CHANNELS_UNSUSPEND (!), etc.
This could be done by throwing an IllegalArgumentException at the very top of this method.

symmetric-core/src/main/java/org/jumpmind/symmetric/model/ChannelMap.java
```diff
@@ -34,58 +35,101 @@
  */
 public class ChannelMap {
     public static final String CHANNELS_SUSPEND = "Suspended-Channels";
+    public static final String CHANNELS_UNSUSPEND = "Unsuspended-Channels";
     public static final String CHANNELS_IGNORE = "Ignored-Channels";
+    public static final String CHANNELS_UNIGNORE = "Unignored-Channels";
     private String channelQueue;
-    private Map<String, Set<String>> map;
+    private final Map<String, Map<String, Set<String>>> map = new HashMap<String, Map<String, Set<String>>>();
```
pavel-jm: When you have some time, let's meet and brainstorm on creating a new wrapper class(es)?
It would cover three main goals:
1. Keep code clean and easy to read by replacing lengthy Map<String, Set<String>> (and, also Map<String, Map<String, Set<String>>>) with a single class name.
2. Move some tedious, but must-unit-test functionality into these "wrapper" classes.
3. Simplify unit testing down to a trivial level.

There are a couple ways to implement this, so let's meet and chat about just about the class+package architecture, when you have some time?

symmetric-core/src/main/java/org/jumpmind/symmetric/model/OutgoingBatches.java
```diff
@@ -119,6 +120,27 @@ public List<OutgoingBatch> filterBatchesForChannels(Set<String> channels) {
...
+                    filtered.add(batch);
+                } else if (excludedNodeIdSet.contains(NodeChannelControl.ALL)) {
+                    Set<String> includedNodeIdSet = includedMap.get(channelId);
+                    if (includedNodeIdSet == null || !includedNodeIdSet.contains(nodeId)) {
+                        filtered.add(batch);
+                    }
+                }
+            }
+        }
+        batches.removeAll(filtered);
```
pavel-jm: If all logic above this line were in a separate method, it would be must more readable and easy to unit-test! 
Both benefits are very valuable to the team long-term.

symmetric-core/src/main/java/org/jumpmind/symmetric/route/ConfigurationChangedDataRouter.java
```diff
@@ -98,6 +99,17 @@ public Set<String> routeToNodes(SimpleRouterContext routingContext, DataMetaData
                  */
                 routeNodeTables(nodeIds, columnValues, rootNetworkedNode, me, routingContext,
                         dataMetaData, possibleTargetNodes, initialLoad);
+            } else if (tableMatches(dataMetaData, TableConstants.SYM_NODE_CHANNEL_CTL)) {
+                String sourceNodeId = columnValues.get("NODE_ID");
```
pavel-jm: From the first look the code inside the IF block appears to be a great candidate for a new method?

pavel-jm (changes requested):
Let's meet (on Tuesday?) to discuss these comments and which ones we want to prioritize for adoption?

symmetric-core/src/main/java/org/jumpmind/symmetric/model/ChannelNodesMap.java
```diff
@@ -0,0 +1,79 @@
...
+package org.jumpmind.symmetric.model;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.TreeMap;
+
+public class TargetNodeMap {
+    private Map<String, Set<String>> map = new TreeMap<String, Set<String>>();
```
pavel-jm, line 33: Could you add a note (here or at a class level) about the purpose of this collection and it's elements, please?

evan-miller-jumpmind:
I made the changes that we discussed:

- Removed the "ALL" keyword
- Changed `ChannelMapWrapper` -> `ChannelMap` -> `TargetNodeMap` to `NodeChannels` -> `ChannelNodesMap` and got rid of the outermost map that just had 2 entries
- Added `isBatchSuspended()` and `isBatchIgnored()` methods to `NodeChannels`
- Added "channel" to some `ChannelNodesMap` method names
- When unregistering a node, deleted all `sym_node_channel_ctl` rows that contain its node ID
- Excluded `last_extract_time` column from sync
- Updated `last_extract_time` column description

Here are the remaining questions that I have, which we can discuss here or in a meeting:

1. The `target_node_id` column isn't needed when keeping track of the `last_extract_time`. Previously, I was using a row with `target_node_id='ALL'` to keep track of the `last_extract_time`. Now that "ALL" has been removed, I'm using a row where `target_node_id=node_id`, and also excluding such rows from routing because they have no other use. Is that okay?
2. What should we do about the new `target_node_id` column during an upgrade from a pre-3.17 version?

pavel-jm:
RE:
>     1. The `target_node_id` column isn't needed when keeping track of the `last_extract_time`. Previously, I was using a row with `target_node_id='ALL'` to keep track of the `last_extract_time`. Now that "ALL" has been removed, I'm using a row where `target_node_id=node_id`, and also excluding such rows from routing because they have no other use. Is that okay?

Without looking at the actual SQL and runtime details, I'm thinking that both nodes should be in the query ?
where node_id=<current_node_id> and target_node_id=<target_node_id>

My concern is that using only the target_node_id column to filter is not specific enough, especially in a 3-tier system.

Maybe we should have a micro-demo in the next meeting to better understand it?

pavel-jm:
RE: 
>     2. What should we do about the new `target_node_id` column during an upgrade from a pre-3.17 version?

My proposal is to delete all records from the SYM_NODE_CHANNEL_CTL and list this as a breaking change in documentation. Hardly anyone uses this table, so spending a lot of time on this seems wasteful?

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
```diff
@@ -558,10 +558,10 @@ protected List<OutgoingBatch> extract(final ProcessInfo extractInfo, final Node
             long batchesSelectedAtMs = System.currentTimeMillis();
             OutgoingBatch currentBatch = null;
             ExecutorService executor = null;
+            Node sourceNode = nodeService.findIdentity();
             try {
```
pavel-jm, line 562: I've noticed the call to nodeService.findIdentity() had moved to be outside the try{}catch block.
How does the risk landscape change with this, please?
Hoping this is arrangement is "easier" for support team to troubleshoot?

evan-miller-jumpmind: Looking at the implementation of `NodeService.findIdentity()`, I don't see a risk of it throwing an exception. It may have to run a SQL query if the identity isn't cached, but it catches a `SqlException` if it occurs.

pavel-jm (approved):

PR #411, opened 2025-10-22
0007097: Route data should handle missing Node entries from the cache

philipmarzullo64:


evan-miller-jumpmind (commented):
Looking at the implementation of `findAvailableNodes()`, I'm not sure if this change will make a difference. The first time this method is called, either it gets the set of nodes from the `context.getAvailableNodes()` map or it finds the nodes and puts them into the map. When this method is called the second time with the same arguments, won't it just get the same set of nodes from the first call?

pavel-jm (approved):
Awesome find!

w3bbd3v-jm (approved):

philipmarzullo64:
I have removed the list of nodes for the target group from the context.

evan-miller-jumpmind (approved):

PR #413, opened 2025-10-22
0007098: Symmetric does not sort by FK references

mdrouard-jumpmind:
Modified data service to force initial load to always sort by FK

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
@@ -1103,8 +1103,7 @@ public class DataService extends AbstractService implements IDataService {
                                         .getActiveTriggerHistories(new Trigger(reloadRequest.getTriggerId(), null)));
                             }
                         }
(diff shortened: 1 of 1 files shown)


pavel-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
```diff
@@ -1103,8 +1103,7 @@ public Map<Integer, ExtractRequest> insertReloadEvents(Node targetNode, boolean
                                         .getActiveTriggerHistories(new Trigger(reloadRequest.getTriggerId(), null)));
                             }
                         }
-                        boolean sortByFk = !(isFullLoad && parameterService.is(ParameterConstants.INITIAL_LOAD_DEFER_CREATE_CONSTRAINTS, false) &&
-                                reloadRequests != null && reloadRequests.size() > 0 && reloadRequests.get(0).isCreateTable());
+                        boolean sortByFk = true;
```
pavel-jm, line 1106: We plan to re-optimize this as a result of work on ticket 0007099

PR #421, opened 2025-10-24
0007106: Fixed CVE-2025-59419

evan-miller-jumpmind:


pavel-jm (approved):
Thanks Evan! Compiled and worked stand-alone just fine

PR #431, opened 2025-11-03
SYM-7015: DDL capture support for cross platform

joshahicks:
While capturing DDL changes on a source and sending to a target platform that is not the same as the target the DDL captured may not work on the target.   This change would resend the table definition so that the target can create the appropriate alters to match the table definition sent from the source.


symmetric-client/src/main/java/org/jumpmind/symmetric/db/mssql/MsSqlTriggerTemplate.java
```diff
@@ -449,8 +446,8 @@ public MsSqlTriggerTemplate(ISymmetricDialect symmetricDialect) {
 "      set @channelId = 'config'\n" +
 "    insert into " + defaultCatalog + "$(defaultSchema)$(prefixName)_data\n" +
 "    (table_name, event_type, trigger_hist_id, row_data, channel_id, source_node_id, create_time)\n" +
-"    values (@tableName, '" + DataEventType.SQL.getCode() + "', @histId,\n" +
-"    '\"delimiter " + delimiter + ";' + CHAR(13) + char(10) + replace(replace(@data.value('(/EVENT_INSTANCE/TSQLCommand/CommandText)[1]', 'nvarchar(max)'),'\\','\\\\'),'\"','\\\"') + '\",ddl',\n" +
+"    values (@tableName, '" + ddlEventType + "', @histId,\n" +
```
evan-miller-jumpmind: What happens when the captured DDL isn't for a specific table and `@tableName` is `sym_node`? It looks it it would end up sending a create event for `sym_node`.

symmetric-core/src/main/resources/symmetric-default.properties
evan-miller-jumpmind, line 1: I think we should document that this parameter is only compatible with SQL Server.

symmetric-db/src/main/java/org/jumpmind/db/alter/ModelComparator.java
evan-miller-jumpmind, line 1: What is the reason for this change?

symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/writer/DefaultDatabaseWriter.java
evan-miller-jumpmind, line 1: What is the reason for this change?

evan-miller-jumpmind (changes requested):
I like the idea of sending a create event rather than the exact DDL statement that ran on the source. Are there any plans to implement this for Postgres and Oracle, or is this issue just for SQL Server?

Edit: I see now that a separate issue has been created for Postgres and Oracle (SYM-7122).

evan-miller-jumpmind:
Also, the issue in the tracker says that the target version is 3.17.0. Should this PR be for 3.17 instead?

Edit: These changes will be included in both 3.16 and 3.17.

symmetric-db/src/main/java/org/jumpmind/db/alter/ModelComparator.java
evan-miller-jumpmind: I see now that this goes with your PR for issue 7005.

evan-miller-jumpmind (approved):

PR #437, opened 2025-11-05
SYM-7006: google sheets integration

mdrouard-jumpmind:
PR for https://issues.symmetricds.org/view.php?id=7006
Pro PR: https://github.com/jumpmindinc/symmetric-pro/pull/428


symmetric-core/src/main/java/org/jumpmind/symmetric/route/AbstractFileParsingRouter.java
```diff
@@ -59,91 +72,71 @@ public abstract class AbstractFileParsingRouter extends AbstractDataRouter {
...
+                    if (options.isIncludeTransactionId()) {
                         transactionId = String.valueOf(System.currentTimeMillis());
                     }
                     for (Map.Entry<Integer, String> tableEntry : tableNames.entrySet()) {
                         String contextId = filePath + "[" + tableEntry.getValue() + "]";
-                        Integer lineNumber = contextService.getString(contextId) == null ? 0 : Integer.valueOf(contextService.getString(contextId));
-                        List<String> dataRows = parse(file, lineNumber, tableEntry.getKey());
+                        Integer lineNumber = 0;
+                        if (options.isTailFile()) {
+                            lineNumber = contextService.getString(contextId) == null ? 0 : Integer.valueOf(contextService.getString(contextId));
```
evan-miller-jumpmind: You could call `contextService.getInt(contextId, 0)` here instead.

evan-miller-jumpmind (approved):
I just found one small optional change you could make. Otherwise it looks good to me.

symmetric-core/src/main/java/org/jumpmind/symmetric/route/AbstractFileParsingRouter.java
pavel-jm: Great suggestion! It makes code much more readable

PR #447, opened 2025-11-10
SYM-7126: Added logging when a column name comes back as null after reading metadata

evan-miller-jumpmind:


symmetric-jdbc/src/main/java/org/jumpmind/db/platform/AbstractJdbcDdlReader.java
```diff
@@ -947,7 +947,11 @@ protected Column readColumn(DatabaseMetaDataWrapper metaData, Map<String, Object
         Column column = new Column();
         PlatformColumn platformColumn = new PlatformColumn();
         platformColumn.setName(platform.getName());
-        column.setName((String) values.get(getName("COLUMN_NAME")));
+        String columnName = (String) values.get(getName("COLUMN_NAME"));
+        column.setName(columnName);
+        if (columnName == null) {
+            log.warn("Encountered null column name when reading column metadata: {}", values);
```
w3bbd3v-jm, line 953: Anything useful to log in `metaData` object that we don't gain here?

w3bbd3v-jm (approved):

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/AbstractJdbcDdlReader.java
evan-miller-jumpmind: Not that I can think of. When I tested this, the `values` map contained additional information like the table name. I think it included the catalog name and schema name as well.

PR #456, opened 2025-11-16
fix(docker): avoid delayed termination due to signal handling

USA-RedDragon:
This commit introduces `tini` (https://github.com/krallin/tini) as an init process for the container. This should fix several issues that arise from the current Docker setup.

Currently, the container runs `sym_service` via a shell command (`sh -c "/opt/symmetric-ds/bin/sym_service start && tail -F ..."`). This setup causes the shell to become PID 1, and signals such as `SIGTERM` or `SIGINT` sent to the container do not reliably reach `sym_service`. In practice, this leads to:

- Ctrl+C in local testing often times out because the service never receives the termination signal.
- `docker stop` commands take longer than necessary, as the service cannot shut down gracefully.
- Kubernetes deployments experience delayed pod termination.

Tini (https://github.com/krallin/tini) acts as a minimal init process that properly forwards termination signals to child processes. It is lightweight, widely adopted, and has been included in Docker since version 1.13.

With this change:
- `sym_service` will receive termination signals promptly.
- Local testing (Ctrl+C) and `docker stop` behave predictably.
- Kubernetes pod shutdown and rollout times are reduced and predictable.

The Dockerfile remains minimal and maintainable, with only a single dependency added.

### DIFF
--- a/symmetric-assemble/Dockerfile
+++ b/symmetric-assemble/Dockerfile
@@ -15,6 +15,9 @@ RUN mkdir -p /opt/
 RUN mv symmetric-server* /opt/symmetric-ds
 RUN chmod -R u=rwX,g=,o= /opt/symmetric-ds/security
 
+# Install tini
+RUN apk add --no-cache tini
+
 ENV SYM_HOME /opt/symmetric-ds
 
 # Create a volume containing the engines, tmp, conf, and security directories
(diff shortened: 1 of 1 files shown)


pavel-jm:
Thank you for this suggestion! Will discuss this week with the team and get back to you.

w3bbd3v-jm:
@USA-RedDragon we want to ensure tini allows for a graceful and clean shutdown before termination, especially allowing busy database connections to finish work, return to the pool, and for resource de-allocation. Do you have a way to test this with SymmetricDS and tini to ensure this happens?

pavel-jm:
Thank you for the tip, @USA-RedDragon ! We have implemented the PID 1 requirement (and native SIGTERM compatibility) in the 3.18.0 release, available here: https://symmetricds.org/news/symmetricds-3-18-0-released/

PR #459, opened 2025-11-19
SYM-7137: Updated the user guide with the changes to the Load Data Wizard, Manage Outgoing Loads screen, and Manage Incoming Loads screen

evan-miller-jumpmind:


pavel-jm (approved):
I very good-looking update! Thanks

PR #460, opened 2025-11-19
SYM-7135: Added H2 SYM_DATA_DATA_ID_SEQ sequence if missing during an upgrade

evan-miller-jumpmind:
It would have been good to limit the `information_schema.sequences` query to only run on an upgrade from 3.16.7 or earlier. Usually there's a check for a particular schema change to determine the version that is being upgraded from, but there are no schema changes in the upgrade from an earlier 3.16 version to 3.16.8.

If it isn't okay to run this query every time an H2 node is upgraded, then I could make this change in 3.17.0 instead or I could try checking `sym_node.symmetric_version` instead of relying on a schema change.

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
@@ -340,6 +340,33 @@ public class DatabaseUpgradeListener implements IDatabaseUpgradeListener, ISymme
                     + " where source_node_id = ? and completed = 1)",
                     engine.getNodeId(), engine.getNodeId());
         }
+        if (engine.getDatabasePlatform().getName().equals(DatabaseNamesConstants.H2)) {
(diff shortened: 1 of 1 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -340,6 +340,33 @@ public String afterUpgrade(ISymmetricDialect symmetricDialect, String tablePrefi
                     + " where source_node_id = ? and completed = 1)",
                     engine.getNodeId(), engine.getNodeId());
         }
+        if (engine.getDatabasePlatform().getName().equals(DatabaseNamesConstants.H2)) {
+            String dataIdSequenceName = symmetricDialect.getSequenceName(SequenceIdentifier.DATA).toUpperCase() + "_SEQ";
```
pavel-jm: This code block is large yet it applies only to H2 database. One way to improve readability (and test-ability) is to move code to a separate utility method.
Another path to consider is a new H2-specific upgrade listener class?
 (We can review this in our Thursday meeting)

pavel-jm (approved):
Thank you!

PR #464, opened 2025-11-21
SYM-7152: Limit time purge job spends on stranded data

pavel-jm:
Limit time purge job spends on stranded data using two parameters:
job.purge.recapture.stranded.data=false --> Turns recapture off altogether
job.purge.stranded.max.time.ms=900000 --> Limits time spent looping though non-existent stranded data IDs


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/PurgeService.java
```diff
@@ -668,11 +681,21 @@ private int purgeByMinMax(long[] minMax, MinMaxDeleteSql identifier, OutgoingCon
             log.debug("Deleted {} rows", count);
             statConsumer.accept(count);
             totalCount += count;
+            long currentRunTimeMs = System.currentTimeMillis() - ts;
             if (count == 0 && (identifier == MinMaxDeleteSql.STRANDED_DATA || identifier == MinMaxDeleteSql.STRANDED_DATA_EVENT)) {
-                log.info("Ending purge of {} early at {} after finding empty space", name, maxId);
-                break;
+                long runtimeLimit = parameterService.getLong(ParameterConstants.PURGE_STRANDED_DATA_TIME_LIMIT_MS);
+                if (runtimeLimit > 0 && currentRunTimeMs >= runtimeLimit) {
```
evan-miller-jumpmind, line 689: Maybe I'm misunderstanding something, but won't this cause the Purge Outgoing job to spend more time on stranded data and stranded data events?

By default, `purgeByMinMax()` won't break out of the loop when finding empty space unless at least 15 minutes have passed. After 15 minutes have passed, it won't break out of the loop until it finds empty space again. Also, it could wait 15 minutes during the stranded data purge and then another 15 minutes during the stranded data event purge.

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/PurgeService.java
```diff
@@ -457,7 +460,12 @@ private int purgeDataGapsExpired(OutgoingContext context) {
...
             int[] argTypes = new int[] { symmetricDialect.getSqlTypeForIds(), symmetricDialect.getSqlTypeForIds() };
             for (DataGap gap : dataGapsExpiredToCheck) {
-                int count = dataService.reCaptureData(gap.getStartId(), gap.getEndId());
+                int count = 0;
+                if (parameterService.is(ParameterConstants.PURGE_STRANDED_DATA_RECAPTURE_ENABLED)) {
+                    count = dataService.reCaptureData(gap.getStartId(), gap.getEndId());
+                } else if (log.isDebugEnabled()) {
+                    log.debug("Skipped recapture of stranded data for gap {} - {}", gap.getStartId(), gap.getEndId());
+                }
                 purgedDataRowCount += count;
```
erilong, line 473: Should purgedDataRowCount be incremented by the delete now, since the count can be 0 if the recapture is disabled?  Down on line 472:

purgedDataRowCount += sqlTemplate.update(getSql("deleteDataByRangeSql"), args, argTypes);

erilong (changes requested):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/PurgeService.java
pavel-jm: This change introduces two paths to consider:
1. Default is 15 minutes, and the intent is to spend at least 15 minutes purging stranded data while ignoring empty spots.
  Since the FOR loop combines processing of all items in the list, the purgeOutgoing() method should not double-dip (to add  30 minutes extra as mentioned above).
2. User now has more control than before, because of the new parameter PURGE_STRANDED_DATA_TIME_LIMIT_MS.
 This allows to limit (or remove the limit ==0) amount of time purge spends on this.

evan-miller-jumpmind: Thanks for the explanation. I have a better understanding of the intent of this change now. Originally I was thinking that its purpose was to reduce the time that the purge job spends on stranded data but I misunderstood.

Regarding your 2nd sentence for path 1, I'm unsure which FOR loop you're referring to. I was just pointing out that `purgeStranded()` calls `purgeByMinMax()` once with `identifier=MinMaxDeleteSql.STRANDED_DATA_EVENT` and again with `identifier=MinMaxDeleteSql.STRANDED_DATA`, and the time limit applies separately to both of these calls.

I'll go ahead and mark this as resolved and I'll let Eric and Phil decide whether to approve this PR since you requested their review specifically.

erilong (approved):

PR #467, opened 2025-11-24
SYM-0007149: SQL_Server_allow_config_to_not_query_for_compressed,_filtered,_and_include_columns_on_indexes

philipmarzullo64:
Parameter to stop querying for SQL Server compressed tables ,and compressed, filtered, and include columns on indexes.
This will speed up the retrieval of metadata if compressed tables, and compressed, filtered, and included columns on indexes are not used or the metadata is not sent to any other nodes.


evan-miller-jumpmind (approved):

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSqlDdlReader.java
```diff
@@ -161,44 +164,49 @@ protected Table readTable(Connection connection, DatabaseMetaDataWrapper metaDat
...
-                                } else if (compressionType == 3) {
-                                    log.debug("table: " + tableName + " has compression: " + CompressionTypes.COLUMNSTORE.name());
-                                    table.setCompressionType(CompressionTypes.COLUMNSTORE);
-                                } else if (compressionType == 4) {
-                                    log.debug("table: " + tableName + " has compression: " + CompressionTypes.COLUMNSTORE_ARCHIVE.name());
-                                    table.setCompressionType(CompressionTypes.COLUMNSTORE_ARCHIVE);
-                                } else {
-                                    table.setCompressionType(CompressionTypes.NONE);
+            if (compressAndFilterAndIncludecolumns) {
+                if (platform instanceof MsSql2008DatabasePlatform) {
```
w3bbd3v-jm: Can we address the SonarCloud Code Analysis warning here? Some of its alerts I think we can ignore. This one seems like an easy one to address and helps with code maintainability.

erilong (approved):

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSqlDdlReader.java
pavel-jm: Another way to organize code for readability (and testability) is to move this into a helper method with a catchy name.

w3bbd3v-jm (approved):

PR #474, opened 2025-12-03
SYM-7166: fix for wrong fix version on issues with multiple.

w3bbd3v-jm:


evan-miller-jumpmind (approved):
Looks good to me, but you have a couple of SonarCloud warnings to check.

PR #478, opened 2025-12-07
SYM-7173: Save hostname on startup and log how it was obtained

pavel-jm:
Save hostname on startup (so it does not get re-evaluated after engine already started) and log how it was obtained


evan-miller-jumpmind (approved):
You have some SonarCloud warnings to look at, but otherwise it looks good to me.

PR #481, opened 2025-12-09
SYM-7186: Fixed NullPointerException when resolving a unique key violation on a table that is referenced by a table in different schema

evan-miller-jumpmind:
Currently `AbstractJdbcDdlReader.lookupForeignTable()` uses the referenced table's catalog and schema instead of the referencing table's catalog and schema. As a result, if the tables are in different schemas, either the table is not found or the wrong table is found, which causes the `NullPointerException`. The change to this method makes it use the referencing table's catalog and schema instead.

I reproduced this issue using Postgres and found that the referencing table's catalog and schema were not populated in the `ForeignKey` object because the keys in the metadata were in lowercase instead of uppercase. The change to `AbstractJdbcDdlReader.readExportedKey()` makes it fall back to the lowercase keys for catalog and schema if the uppercase keys are not found.

### DIFF
--- a/symmetric-jdbc/src/main/java/org/jumpmind/db/platform/AbstractJdbcDdlReader.java
+++ b/symmetric-jdbc/src/main/java/org/jumpmind/db/platform/AbstractJdbcDdlReader.java
@@ -1200,11 +1200,11 @@ public abstract class AbstractJdbcDdlReader implements IDdlReader {
             fk = new ForeignKey(fkName);
             fk.setForeignTableName((String) values.get(getName("FKTABLE_NAME")));
             try {
-                fk.setForeignTableCatalog((String) values.get(getName("FKTABLE_CAT")));
+                fk.setForeignTableCatalog((String) values.getOrDefault(getName("FKTABLE_CAT"), values.get(getName("fktable_cat"))));
             } catch (Exception e) {
             }
             try {
-                fk.setForeignTableSchema((String) values.get(getName("FKTABLE_SCHEM")));
+                fk.setForeignTableSchema((String) values.getOrDefault(getName("FKTABLE_SCHEM"), values.get(getName("fktable_schem"))));
             } catch (Exception e) {
             }
             knownFks.put(fkName, fk);
@@ -1832,7 +1832,7 @@ public abstract class AbstractJdbcDdlReader implements IDdlReader {
 
     private Table lookupForeignTable(IDatabasePlatform platform, ForeignKey fk, TableRow tableRow, boolean clearPrimaryKeys) {
         Table foreignTable = null;
-        Table table = platform.getTableFromCache(tableRow.getTable().getCatalog(), tableRow.getTable().getSchema(), fk.getForeignTableName(), false);
+        Table table = platform.getTableFromCache(fk.getForeignTableCatalog(), fk.getForeignTableSchema(), fk.getForeignTableName(), false);
         if (table == null) {
             table = fk.getForeignTable();
             if (table == null) {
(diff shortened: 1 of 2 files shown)


symmetric-jdbc/src/main/java/org/jumpmind/db/platform/AbstractJdbcDdlReader.java
```diff
@@ -1832,7 +1832,7 @@ protected List<TableRow> getImportedForeignTableRows(List<TableRow> tableRows, S
 
     private Table lookupForeignTable(IDatabasePlatform platform, ForeignKey fk, TableRow tableRow, boolean clearPrimaryKeys) {
```
w3bbd3v-jm: Can we add unit tests for both changed methods, as they are on the smaller side?

w3bbd3v-jm:
> Currently `AbstractJdbcDdlReader.lookupForeignTable()` uses the referenced table's catalog and schema instead of the referencing table's catalog and schema. As a result, if the tables are in different schemas, either the table is not found or the wrong table is found, which causes the `NullPointerException`. The change to this method makes it use the referencing table's catalog and schema instead.
> 
> I reproduced this issue using Postgres and found that the referencing table's catalog and schema were not populated in the `ForeignKey` object because the keys in the metadata were in lowercase instead of uppercase. The change to `AbstractJdbcDdlReader.readExportedKey()` makes it fall back to the lowercase keys for catalog and schema if the uppercase keys are not found.

Should we fix this in 3.15 too?

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/AbstractJdbcDdlReader.java
evan-miller-jumpmind: We ended up discussing this PR in a meeting. Pavel mentioned that it might not be necessary to add unit tests for now since I didn't add any new methods and because we don't want to delay this fix. We decided that I should just add a test for `readExportedKey()` for now since a test for `lookupForeignTable()` seems like it would be more difficult to write. I'll write a test when I have the time.

evan-miller-jumpmind:
> > Currently `AbstractJdbcDdlReader.lookupForeignTable()` uses the referenced table's catalog and schema instead of the referencing table's catalog and schema. As a result, if the tables are in different schemas, either the table is not found or the wrong table is found, which causes the `NullPointerException`. The change to this method makes it use the referencing table's catalog and schema instead.
> > I reproduced this issue using Postgres and found that the referencing table's catalog and schema were not populated in the `ForeignKey` object because the keys in the metadata were in lowercase instead of uppercase. The change to `AbstractJdbcDdlReader.readExportedKey()` makes it fall back to the lowercase keys for catalog and schema if the uppercase keys are not found.
> 
> Should we fix this in 3.15 too?

Typically I don't backport a fix to 3.15 unless a customer running 3.15 is affected by the bug. It's not a bad idea though. I could add 3.15 to the target versions so it gets automatically cherry-picked over if we think that would be best.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/AbstractJdbcDdlReader.java
w3bbd3v-jm: Seems reasonable to me. Thanks!

w3bbd3v-jm (approved):

pavel-jm:
FYI - I have tagged SYM-7186 for the 3.15.22 release.
If it's not too much time to write a test for readExportedKey(), let's do that! But testing lookupForeignTable() would be more time consuming...

pavel-jm (approved):

PR #490, opened 2025-12-18
SYM-7177: switch to maven central

erilong:


w3bbd3v-jm:
@erilong Is the build failing because the new symmetric-modules package needs to be added?

pavel-jm (approved):

w3bbd3v-jm (approved):

pavel-jm (approved):

PR #492, opened 2025-12-19
SYM-7177: symmetric modules to maven central

erilong:


symmetric-modules/lib/READ_ME.txt
awebbdev1: Can we delete this along with /lib directory?

w3bbd3v-jm (approved):

PR #496, opened 2025-12-22
SYM-7206: Fix JDBC bulk loader boundary record count issue

pavel-jm:
Hold off merging - still testing this change!


symmetric-client/src/main/java/org/jumpmind/symmetric/io/JdbcBatchBulkDatabaseWriter.java
```diff
@@ -81,14 +81,11 @@ protected LoadStatus update(CsvData data, boolean applyChangesOnly, boolean useC
...
-        checkForConflict(true);
+        checkForConflict();
         return LoadStatus.SUCCESS;
     }
 
-    protected void checkForConflict(boolean isDml) {
-        if (isDml) {
-            expectedRowCount++;
-        }
+    protected void checkForConflict() {
```
pavel-jm, line 88: This database writer is only used for DML events, so there is no need to account for non-DML events.

w3bbd3v-jm (approved):

pavel-jm:
Another fix eliminates the need for this change= https://github.com/jumpmindinc/symmetric-ds/pull/537

PR #498, opened 2025-12-23
SYM-7209: load is running, non reload channels are still synching

philipmarzullo64:


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
```diff
@@ -430,6 +430,12 @@ public List<TableReloadStatus> getActiveOutgoingTableReloadStatus() {
                 new TableReloadStatusMapper(), engine.getNodeId());
     }
 
+    @Override
+    public List<TableReloadStatus> getActiveOutgoingTableReloadStatusByTargetNodeId(String targetNodeId) {
```
pavel-jm, line 434: Let's track time spent on the query in getActiveOutgoingTableReloadStatusByTargetNodeId 

Just to catch and document (in the log) situations where this query is expensive or blocked for an unforeseen reason...

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
```diff
@@ -388,6 +408,14 @@ private List<OutgoingBatch> filterBatchesForExtraction(OutgoingBatches batches,
         return batches.getBatches();
     }
 
+    private void logQueryDuration(String message, long duration, long maxDuration) {
```
pavel-jm, line 412: Let's declare a constant for the "magic" number 5000l sprinkled in the code.
MAX_DURATION_FOR_QUERY

pavel-jm (approved):

PR #505, opened 2025-12-31
SYM-7071: Added methods to ProcessType and StatisticManager to support the Pipelines Dashboard panel improvements

evan-miller-jumpmind:


symmetric-core/src/main/java/org/jumpmind/symmetric/model/ProcessType.java
```diff
@@ -141,4 +144,63 @@ public String toStringShort() {
                 return name();
         }
     }
+
+    public String toStringShortPastTense() {
```
w3bbd3v-jm, line 79: Instead of repeating this pattern should we use a messages class or something like that to encapsulate these strings? Another approach might be to create a new ProcessTypeMessages class or something like that.

evan-miller-jumpmind: I can't use `Messages` here because it's a pro class. I'll create a `ProcessTypesMessages` class and move the strings there.

w3bbd3v-jm: I think that's a step in the right direction. The other problem though is that we'd potentially have to remember to update the strings in all three methods which are spread out in different switch statements located in different parts of the class. If we forget to add one, it would fall silently through to name(). Can you take this a step further to co-locate the data in a single location in the enum? Something like this:
`public enum ProcessType { 
      ANY(ANY_MSG, ANY_MSG, ANY_MSG),                                                                                                                                     
      PUSH_JOB_EXTRACT(DATABASE_PUSH_EXTRACT, EXTRACTING, EXTRACTED),                                                                                                     
      PUSH_JOB_TRANSFER(DATABASE_PUSH_TRANSFER, TRANSFERRING, TRANSFERRED),  
      ROUTER_JOB(ROUTING, ROUTING, ROUTED),
      ...
}`
Then ProcessType could be:

     
`{     ...
       private final String displayName;                                                                                                                                   
       private final String shortName;                                                                                                                                     
       private final String shortPastTense;                                                                                                                                
                                                                                                                                                                            
       ProcessType(String displayName, String shortName, String shortPastTense) {                                                                                          
             this.displayName = displayName;                                                                                                                                 
             this.shortName = shortName;                                                                                                                                     
             this.shortPastTense = shortPastTense;                                                                                                                           
       }                                     
       
       @Override
       public String toString() {
              return displayName != null ? displayName : name();  
       }
       
        public String toStringShort() {
               return shortName != null ? shortName : name();                                                                                                                  
       }
      ... 
`
It might take a little refactoring but, it will prevent us from missing an update or adding a new type in the future. It also follows the DRY pattern.

w3bbd3v-jm: Let me know if you have any questions or I can help with a PR for the changes.

evan-miller-jumpmind: Good idea! I made some changes to pass the strings into the constructor instead. Let me know if any more changes should be made.

w3bbd3v-jm: Looks good! Thanks!

w3bbd3v-jm (approved):

PR #515, opened 2026-01-09
SYM-7198: push and pull job schedule configuration

w3bbd3v-jm:
Wanted to get this out for people to look at. Will open the PR for the PRO side shortly. Still doing some testing.
This PR does not include changing the custom jobs. That will be a separate PR.


symmetric-client/src/main/java/org/jumpmind/symmetric/job/AbstractJob.java
```diff
@@ -81,16 +81,27 @@ public AbstractJob(String jobName, ISymmetricEngine engine, ThreadPoolTaskSchedu
...
+                Date lastLockTime = (lock != null) ? lock.getLastLockTime() : null;
+                Instant lastCompletion = (lastLockTime != null) ? lastLockTime.toInstant() : null;
+                SimpleTriggerContext triggerContext = new SimpleTriggerContext(lastCompletion, lastCompletion, lastCompletion);
+                Instant firstRunCouldBeInPast = cronTrigger.nextExecution(triggerContext);
+                Instant now = Instant.now();
+                Instant firstRun = (firstRunCouldBeInPast != null && firstRunCouldBeInPast.isBefore(now))
+                        ? cronTrigger.nextExecution(new SimpleTriggerContext(now, now, now))
+                        : firstRunCouldBeInPast;
+                log.info("Starting job '{}' on cron schedule '{}', with the first run at {}, and previous run at {}.", jobName, cronExpression,
+                        Date.from(firstRun), lastLockTime);
```
evan-miller-jumpmind, line 104: It might be good to exclude the previous run time from the log message when `lastLockTime` is null so we don't log a message ending with "previous run at null". There's another log message for a periodic schedule below that could also be improved.

evan-miller-jumpmind (changes requested):
I made one suggestion on this PR, but also see my review of the pro PR.

symmetric-client/src/main/java/org/jumpmind/symmetric/job/AbstractJob.java
w3bbd3v-jm: I coded it that way just so we don't have the extra code and so the last lock time is explicitly indicated even when `null`. If you still want to change it, let me know. I'm not hard pressed either way.

pavel-jm (approved):

evan-miller-jumpmind (approved):

PR #516, opened 2026-01-11
SYM-7234: Clean up log messages to help find the real ERROR

pavel-jm:
The ERROR-level log message for IgnoreRowException thrown by custom BSH transform should be downgraded to DEBUG level.
The "Purging incoming error rows" message from the Purge job trips log monitors on the word "error" surrounded by spaces (false alarm!)


w3bbd3v-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/io/data/transform/BshColumnTransform.java
```diff
@@ -163,26 +164,42 @@ public NewAndOldValue transform(IDatabasePlatform platform,
             }
         } catch (TargetError evalEx) {
             Throwable ex = evalEx.getTarget();
-            log.error("TargetError detected for " + transformBeanshellInfo, ex);
             if (ex instanceof IgnoreColumnException) {
+                if (log.isDebugEnabled()) {
```
w3bbd3v-jm, line 168: Approved -- you'll need to fix the style violations before merging however.

PR #520, opened 2026-01-13
SYM-7234: Clean up log messages to help find the real ERROR (#516)

pavel-jm:
Cherry-pick 3.16 PR
https://github.com/jumpmindinc/symmetric-ds/pull/516


symmetric-modules/lib/README.md
```diff
@@ -1 +1 @@
-Use this directory to temporarily place the modules you want to upload. For more information, please see the README file in the symmetric-modules directory.
+Use this directory to temporarily place the modules you want to upload. For more information, please see the README file in the symmetric-modules directory.
```
pavel-jm, line 1: Apparently this file was recently modified on Windows and had CR+LF line endings.
I have added an instruction to the .gitattributes to works around this issue (which otherwise blocks developer from switching GIT branches!).

w3bbd3v-jm: Does your change to gitattributes fix the issue? 
I'm wondering because there is already an entry for 
*.md text diff=markdown
*.MD text diff=markdown
, and the extension of the md file of interest is lowercase. Because this is a new file and module, introduced with maven repo migration, I'm wondering if it's just not using the gitattributes file for some reason...? Just a thought. 

Also, the file extensions are case sensitive, but I believe something like this is supported.
*.[mM][dD] text diff=markdown

evan-miller-jumpmind (approved):

PR #521, opened 2026-01-13
SYM-7237: Fixed NumberFormatException when sending a load involving a SQL Server time column over a subselect or Beanshell router

evan-miller-jumpmind:
The `NumberFormatException` occurs when the `time` value is prefixed with a date, like `1970-01-01 02:30:00.000`. This seems to only happen during a load, not during CDC. The fix is to remove the date component before parsing the value.

I went ahead and created this PR after creating a patch for a customer. I can add unit tests if necessary.


w3bbd3v-jm:
> The `NumberFormatException` occurs when the `time` value is prefixed with a date, like `1970-01-01 02:30:00.000`. This seems to only happen during a load, not during CDC. The fix is to remove the date component before parsing the value.
> 
> I went ahead and created this PR after creating a patch for a customer. I can add unit tests if necessary.

Unit tests would be nice here especially since there is straight forward input/output expectations. It's also hard to visually parse and understand without context. Unit tests would bridge the gap. If you wanted to unit test it directly, you could make the method protected, IMO.

w3bbd3v-jm (approved):

PR #526, opened 2026-01-15
SYM-7231: Move MsSql classes to pro

Sai-Gumpalli:
@Sai-Gumpalli you'll need to resolve the conflicts before the review. 

A good practice is pulling down changes to your target branch, in this case release/3.17, frequently to avoid having to work through multiple or complicated merge conflicts, to keep your feature branch in sync. I can help or let me know if you have any questions.


symmetric-db/src/main/java/org/jumpmind/db/sql/DmlStatement.java
```diff
@@ -551,14 +550,14 @@ protected String escapeText(String value) {
...
+        this.sql = this.sql.replaceAll(prefix + ":",
+                prefix + ":" + value);
     }
 
-    public static String updateCteExpression(String sql, String value) {
-        return sql.replaceAll(MsSql2008DdlBuilder.CHANGE_TRACKING_SYM_PREFIX + ":",
-                MsSql2008DdlBuilder.CHANGE_TRACKING_SYM_PREFIX + ":" + value);
+    public static String updateCteExpression(String sql, String value, String prefix) {
+        return sql.replaceAll(prefix + ":",
+                prefix + ":" + value);
```
evan-miller-jumpmind: These changes to the two `DmlStatement.updateCteExpression()` methods concern me.

There are references to these methods in `DefaultDatabaseWriterConflictResolver` that pass in the result of `IDdlBuilder.getCteExpressionPrefix()` as the `prefix` argument and the source node ID as the `value` argument. However, the `AbstractDdlBuilder.getCteExpressionPrefix()` method returns an empty string, which will be the case for most database platforms since only SQL Server overrides it. In this case, all instances of `:` in the SQL will be replaced with `:[source node ID]`, which seems like it might cause issues.

Would it be better to only call `replaceAll()` if the `prefix` argument is not empty?

evan-miller-jumpmind (approved):

PR #527, opened 2026-01-20
SYM-7247: Spring and Spring Boot dependency updates.

w3bbd3v-jm:
Last attempt to update Spring and Spring Boot got reverted due to slow UI, attributed to Vaadin dependency changes. I tested this in standalone mode and noticed some slowness, but it seemed to be short-lived and found the UI functional. During this time, I did see some log entries like this: 
```
2026-01-20 16:22:04,596 WARN [gui] [ServerRpcHandler] [qtp1472044075-44] Resynchronizing UI by client's request. A network message was lost before reaching the client and the client is reloading the full UI state. This typically happens because of a bad network connection with packet loss or because of some part of the network infrastructure (load balancer, proxy) terminating a push (websocket or long-polling) connection. If you are using push with a proxy, make sure the push timeout is set to be smaller than the proxy connection timeout
```
Can someone else test this branch to see how the experience is? Notice any difference from prior releases?


evan-miller-jumpmind:
I tested your branch in standalone mode and it got stuck when opening the Deploy Wizard. It showed the "Loading Deploy Wizard..." dialog and hanged. Restarting didn't help and it also got stuck on the "Testing Connections..." dialog when re-entering the Canvas. I checked the log and didn't see any unusual messages.

I'm not sure what would cause this. I tested your branch out of Eclipse and didn't experience the issue.

w3bbd3v-jm:
> I tested your branch in standalone mode and it got stuck when opening the Deploy Wizard. It showed the "Loading Deploy Wizard..." dialog and hanged. Restarting didn't help and it also got stuck on the "Testing Connections..." dialog when re-entering the Canvas. I checked the log and didn't see any unusual messages.
> 
> I'm not sure what would cause this. I tested your branch out of Eclipse and didn't experience the issue.

Thanks for checking, Evan! I'll going to do some more digging before moving this forward.

w3bbd3v-jm:
Aligning with the Jetty version included in Spring Boot's BOM seemed to fix the problem for me. Can I get a second opinion/tester?

evan-miller-jumpmind (approved):
I tested the standalone version again and the Jetty upgrade resolved the issues.

w3bbd3v-jm:
> I tested the standalone version again and the Jetty upgrade resolved the issues.

Thanks again, @evan-miller-jumpmind !

PR #533, opened 2026-01-23
SYM-7250: Fix Java object Byte references in CSV data.

w3bbd3v-jm:


symmetric-db/src/main/java/org/jumpmind/db/sql/Row.java
```diff
@@ -150,7 +150,13 @@ public String csvValue() {
             if (index > 0) {
                 concatenatedRow.append(",");
             }
-            concatenatedRow.append(obj != null ? obj.toString() : "");
+            if (obj != null) {
+                if (obj instanceof byte[]) {
+                    concatenatedRow.append(Hex.encodeHexString((byte[]) obj));
```
pavel-jm, line 155: I wonder how does the target node "know" that the text value is Hex-encoded and not just a random string with goofy characters? In the past the encoding was recorded as a property of the SymmetricDS batch?

w3bbd3v-jm:
Closing this PR in favor of a separate fix.

PR #538, opened 2026-02-02
SYM-7258: Fixed DDL looping back to the source node when sending a load with create_table=1 to a node with DDL capture enabled

evan-miller-jumpmind:
The problem is that `DefaultDatabaseWriter.create()` uses a `SqlScript` to execute DDL instead of the `ISqlTransaction` that has the session variable set that populates the `sym_data.source_node_id` column.

I used Claude to come up with a plan to resolve this and found that a solution is to add an `IConnectionHandler` to set the session variable on the `Connection` used by the `SqlScript` before the DDL is executed, then un-set the variable after it's done. This required the addition of new `Connection`-specific `disableSyncTriggers()` and `enableSyncTriggers()` methods to the dialects that support DDL capture.

### DIFF
--- a/symmetric-client/src/main/java/org/jumpmind/symmetric/db/postgresql/PostgreSqlSymmetricDialect.java
+++ b/symmetric-client/src/main/java/org/jumpmind/symmetric/db/postgresql/PostgreSqlSymmetricDialect.java
@@ -20,9 +20,13 @@
  */
 package org.jumpmind.symmetric.db.postgresql;
 
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.sql.Statement;
 import java.sql.Types;
 import java.util.Date;
 import java.util.List;
+import java.util.Objects;
 
 import org.apache.commons.lang3.StringUtils;
 import org.jumpmind.db.model.Table;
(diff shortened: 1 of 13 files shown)


symmetric-client/src/main/java/org/jumpmind/symmetric/db/postgresql/PostgreSqlSymmetricDialect.java
```diff
@@ -275,21 +279,55 @@ public void removeDdlTrigger(StringBuilder sqlBuffer, String catalogName, String
         }
     }
 
+    private String getDisableSyncTriggersSql() {
+        return "select set_config('" + SYNC_TRIGGERS_DISABLED_VARIABLE + "', '1', false)";
```
pavel-jm, line 288: Per our 2/3/26 conversation let's add a member variable to track disable and enable events is hope this provides us with better awareness:
Example:
int triggersDisabledCount=0;
public int getTriggersDisabledCount() {return triggersDisabledCount; }

symmetric-client/src/main/java/org/jumpmind/symmetric/db/postgresql/PostgreSqlSymmetricDialect.java
```diff
@@ -275,21 +279,55 @@ public void removeDdlTrigger(StringBuilder sqlBuffer, String catalogName, String
...
+    }
+
+    private String getDisableSyncNodeSql(String nodeId) {
+        return "select set_config('" + SYNC_NODE_DISABLED_VARIABLE + "', '" + nodeId + "', false)";
```
pavel-jm, line 293: Per our 2/3/26 conversation let's add a member variable to track disable and enable events in hope this provides us with better awareness:
Example:
int nodeDisabledCount=0;
public int getNodeDisabledCount() {return nodeDisabledCount; }

pavel-jm (changes requested):

pavel-jm (approved):

pavel-jm:
The integration test appears to be failing. I wonder if this is a wider issue as two other PRs have this symptom!
https://github.com/jumpmindinc/symmetric-ds/actions/runs/21638794777/job/62372033885?pr=538

evan-miller-jumpmind:
Yeah, [this PR's](https://github.com/jumpmindinc/symmetric-ds/pull/542) integration tests are failing for the same reason.

```
SimpleIntegrationTest > test11SuspendIgnorePushRemoteBatches() FAILED
    java.lang.AssertionError: There should be one outgoing batches. expected:<1> but was:<0>
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:645)
        at org.jumpmind.symmetric.test.SimpleIntegrationTest.lambda$test11SuspendIgnorePushRemoteBatches$10(SimpleIntegrationTest.java:389)
 ```

PR #544, opened 2026-02-05
SYM-7280: Fixed TriggerRouterService.getActiveTriggerHistoriesFromCache() not filtering out inactive trigger histories

evan-miller-jumpmind:


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/TriggerRouterService.java
```diff
@@ -478,10 +478,13 @@ public TriggerHistory getNewestTriggerHistoryForTrigger(String triggerId, String
...
 
-    @SuppressWarnings("unchecked")
     @Override
     public List<TriggerHistory> getActiveTriggerHistoriesFromCache() {
-        return new ArrayList<TriggerHistory>(historyMap != null ? historyMap.values() : Collections.EMPTY_LIST);
+        List<TriggerHistory> historyList = new ArrayList<TriggerHistory>();
+        if (historyMap != null) {
+            historyList.addAll(historyMap.values().stream().filter(h -> h.getInactiveTime() == null).collect(Collectors.toList()));
+        }
+        return historyList;
```
w3bbd3v-jm, line 487: Generally it's better to return an empty list instead of null.

evan-miller-jumpmind: The `historyList` isn't null. Maybe you're confusing it with the `historyMap`, which is the cache?

w3bbd3v-jm: Yep! Nevermind me!

w3bbd3v-jm (approved):

PR #547, opened 2026-02-05
SYM-7279: Clarification of the http.timeout.ms parameter

pavel-jm:


evan-miller-jumpmind (approved):

symmetric-core/src/main/resources/symmetric-default.properties
w3bbd3v-jm, line 529: While we're at it, should we update the description for this parameter, as it can be easily confused with `http.timeout.ms`?

w3bbd3v-jm (approved):

PR #548, opened 2026-02-05
SYM-7276: Integration test fails for MySQL - wrong number of rows in each batch

Sai-Gumpalli:


pavel-jm (approved):

w3bbd3v-jm (changes requested):
This fix feels hack-y. Why does this fix the test? Are we just covering up a real issue?
If this is truely a fix, then can we document it to explain why the sleep is there?

w3bbd3v-jm (changes requested):
For the 2 TODO comments, can we either remove them or update them with what really needs to take place to fix the issue?

pavel-jm (approved):

pavel-jm:
> For the 2 TODO comments, can we either remove them or update them with what really needs to take place to fix the issue?

Done!

pavel-jm:
> This fix feels hack-y. Why does this fix the test? Are we just covering up a real issue? If this is truely a fix, then can we document it to explain why the sleep is there?

It is an odd one! In MySQL a commit is NOT enough to switch Transaction ID to the next value?!
MySQL is the worst DB of the year 2025 in my book ;-)

w3bbd3v-jm (approved):

PR #553, opened 2026-02-10
SYM-7200: Minio with S3 upgrade  

w3bbd3v-jm:
Feature from @joshahicks and @philipmarzullo64 for review


symmetric-core/src/main/java/org/jumpmind/symmetric/model/FileSnapshot.java
```diff
@@ -332,6 +342,11 @@ public boolean equals(Object obj) {
                 return false;
         } else if (!routerId.equals(other.routerId))
             return false;
+        if (externalFileData == null) {
```
evan-miller-jumpmind, line 346: I think if we add `externalFileData` to `equals()`, then we should also add it to `hashCode()`.

symmetric-core/src/main/java/org/jumpmind/symmetric/route/RelationalToFileRouter.java
```diff
@@ -0,0 +1,62 @@
...
+    @Override
+    public Set<String> routeToNodes(SimpleRouterContext context, DataMetaData dataMetaData, Set<Node> nodes,
+            boolean initialLoad, boolean initialLoadSelectUsed, TriggerRouter triggerRouter) {
+        dataMetaData.getData().getAttribute(CsvData.ROW_DATA);
+        DataContext ctx = null;
+        LastEventType lastEventType = null;
+        if (ctx.getData().getDataEventType() == DataEventType.INSERT) {
+            lastEventType = LastEventType.CREATE;
+        }
+        FileTriggerRouter fileTriggerRouter = engine.getFileSyncService().getFileTriggerRouter("S3", "router", false);
```
evan-miller-jumpmind: What is happening here? Is the user expected to create a `sym_file_trigger_router` row with `trigger_id='S3'` and `router_id='router'` before using this router? If so, we should document this.

Also, the line below should handle the case where this call returns `null`. Currently, if `fileTriggerRouter` is `null`, then the `FileSnapshot` constructor will throw a `NullPointerException`.

symmetric-core/src/main/java/org/jumpmind/symmetric/route/RelationalToFileRouter.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public boolean isConfigurable() {
+        // TODO Auto-generated method stub
+        return false;
+    }
+
+    @Override
+    public boolean isDmlOnly() {
+        // TODO Auto-generated method stub
```
evan-miller-jumpmind: There are 4 `TODO` comments in this file. Does something need implemented here or can they be removed?

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/FileSyncService.java
```diff
@@ -1082,6 +1084,7 @@ protected List<IncomingBatch> processZip(InputStream is, String sourceNodeId,
                                     }
                                 }
                                 try {
+                                    // log.info("SCRIPT : " + script);
```
evan-miller-jumpmind: It looks like this line can be removed.

evan-miller-jumpmind (changes requested):
I noticed that there's no documentation added in this PR. At a minimum, I think we should document the new router type.

symmetric-core/src/main/java/org/jumpmind/symmetric/route/RelationalToFileRouter.java
joshahicks: This class should not have been checked in it was an initial attempt to capture a table change and convert it to a file change during routing.   This was getting a bit more involved to hit deadlines so we did this through configuration using a BSH transform.   Longer term this might be nice to introduce back to include an out of the box method of capturing a table change to initiate file sync.

joshahicks:
These changes have been committed

evan-miller-jumpmind (approved):
Looks good! There is still no documentation apart from the descriptions of the new parameters, but if you don't think any more documentation is needed, then this should be good to merge.

PR #554, opened 2026-02-10
SYM-7281: Add target delimited identifier mode

Sai-Gumpalli:


symmetric-core/src/main/resources/symmetric-default.properties
```diff
@@ -400,10 +400,20 @@ db.metadata.ignore.case=true
 # identifiers (which may only contain alphanumerical characters and the
 # underscore, must start with a letter and cannot be a reserved keyword).
 #
+# DatabaseOverrideable: true
```
pavel-jm: In order to keep this parameter in the "Startup Parameters" category (same as before this change), please set the hint to:
DatabaseOverrideable: false

This should match the target.db.delimited.identifier.mode

pavel-jm (changes requested):

symmetric-core/src/main/resources/symmetric-default.properties
```diff
@@ -400,10 +400,20 @@ db.metadata.ignore.case=true
 # identifiers (which may only contain alphanumerical characters and the
 # underscore, must start with a letter and cannot be a reserved keyword).
 #
+# DatabaseOverrideable: false
 # Tags: database
 # Type: boolean
 db.delimited.identifier.mode=true
 
+# Determines whether delimited identifiers are used or normal SQL92
+# identifiers on target databse connection (which may only contain alphanumerical characters and the
```
evan-miller-jumpmind: There's a typo here

evan-miller-jumpmind (approved):
There's a typo that needs fixed but it looks good otherwise

w3bbd3v-jm (approved):

pavel-jm (approved):

symmetric-core/src/main/resources/symmetric-default.properties
pavel-jm: Thank you! Fixed

PR #574, opened 2026-02-22
SYM-7346: Fix DB2 LUW platform bugs

gp510:
## Summary

Fixes DB2 LUW platform bugs discovered during the FELE POC (one-way DB2 LUW → SQL Server replication). These issues prevent trigger creation, can cause destructive DDL migrations, and break foreign key / column metadata reading.

### Changes

- **SYM-7346**: Override `getSourceTablePrefix()` in `Db2TriggerTemplate` to strip catalog from trigger DDL, preventing `SQLCODE=-108` from 3-part qualified names
- **SYM-7347**: Return `null` from `Db2DatabasePlatform.getDefaultCatalog()` — DB2 doesn't support catalogs, matching Oracle and AS/400 which already return `null`. The previous `""` return value leaked into JDBC `meta.getTables()` where `""` means "tables without a catalog" (returning nothing) vs `null` meaning "don't filter by catalog". This fix also eliminates 4 workaround overrides in `Db2DdlReader` (`getCatalogNames`, `readTableFromConnection`, `getTableNames`, `getTableNamesFromDatabase`) that were only compensating for `""` leaking into JDBC calls
- **SYM-7350**: Map `LONG VARCHAR` to `LONGVARCHAR` (-1) instead of `CLOB` (2005) in `mapUnknownJdbcTypeForColumn()` to prevent destructive DDL migration cycle where `ModelComparator` detects a type mismatch and drops/recreates `SYM_DATA`, losing `IDENTITY` columns
- **SYM-7351**: Handle non-standard DB2 JDBC metadata keys (`NAME` → `COLUMN_NAME`, `REFTBNAME`/`TBNAME` → `PKTABLE_NAME`, `COLNAME` → `PKCOLUMN_NAME`) as fallbacks in `readColumn()` and `readForeignKey()`

### Unit tests

Added `Db2DdlReaderTest` with 11 tests:
- `getDefaultCatalog()` returns null
- `mapUnknownJdbcTypeForColumn` for LONG VARCHAR, CLOB, DBCLOB, ROWID, XML
- `readColumn` NAME fallback (with and without standard COLUMN_NAME key)
- `readForeignKey` fallbacks for REFTBNAME, TBNAME, COLNAME, and skip on missing table name

## Test plan

- [x] Unit tests pass (11 tests in `Db2DdlReaderTest`)
- [x] Build passes (`symmetric-jdbc`, `symmetric-client`)
- [x] End-to-end POC: DB2 LUW 12.1 → SQL Server with field-level transforms
  - [x] `sync-triggers` creates triggers with 2-part names (`DB2INST1.SYM_*`), no SQLCODE errors
  - [x] `SYM_DATA` not dropped/recreated on startup (LONG VARCHAR mapping)
  - [x] Initial load extracts and delivers successfully (5 rows each in PURCHASE_ORDER, ITEM_MASTER)
  - [x] Zero error batches on hub and spoke
  - [x] Transforms applied correctly (CYYMMDD → DATE, implied decimals → DECIMAL, CHAR → trimmed VARCHAR)

### DIFF
--- a/symmetric-assemble/common.gradle
+++ b/symmetric-assemble/common.gradle
@@ -131,6 +131,9 @@ subprojects { subproject ->
     }
 
 	java {
+		toolchain {
+    		languageVersion = JavaLanguageVersion.of(17)
+    	}
     	sourceCompatibility=JavaVersion.VERSION_17
     	targetCompatibility=JavaVersion.VERSION_17
     }
@@ -345,13 +348,13 @@ subprojects { subproject ->
         powerMockVersion = '2.0.9'
         mysqlVersion = '8.4.0'
         servletVersion = '4.0.1'
-        springVersion = '6.2.11'
-        springBootVersion = '3.5.5'
+        springVersion = '6.2.15'
+        springBootVersion = '3.5.9'
         jtdsVersion = '1.3.1'
         bouncyCastleVersion = '1.81'
         animalSnifferVersion = '1.23'
         jnaVersion = '5.14.0'
-        jettyVersion = '12.0.25'
+        jettyVersion = '12.0.31'
         websocketVersion = '2.2.0'
         env = System.getenv()
     }
--- a/symmetric-assemble/common.gradle
+++ b/symmetric-assemble/common.gradle
@@ -233,37 +233,36 @@ subprojects { subproject ->
 			sign publishing.publications.symmetricLibrary
 		}
 
-        // Commented out for local build - not needed for development
-        // apply plugin: 'eu.kakde.gradle.sonatype-maven-central-publisher'
-        //
-        // def licenseName = project.name.contains("-pro") ? "Commercial" : "GNU GPL version 3.0";
-        // sonatypeCentralPublishExtension {
-        //     groupId.set("${project.group}")
-        //     artifactId.set("${project.name}")
-        //     version.set("${project.version}")
-        //     componentType.set("java")
-        //     publishingType.set("AUTOMATIC")
-        //     username.set(getProperty("sonatypeUser"))
-        //     password.set(getProperty("sonatypePassword"))
-        //     pom {
-        //         name.set("${project.name}")
-        //         description.set("Data replication")
-        //         url.set("https://symmetricds.org/")
-        //         licenses {
-        //             license {
-        //                 name.set(licenseName)
-        //             }
-        //         }
-        //         developers {
-        //             developer {
-        //                 name.set("Jumpmind")
-        //             }
-        //         }
-        //         scm {
-        //             url.set("https://github.com/jumpmindinc/symmetric-ds/")
-        //         }
-        //     }
-        // }
+        apply plugin: 'eu.kakde.gradle.sonatype-maven-central-publisher'
+
+        def licenseName = project.name.contains("-pro") ? "Commercial" : "GNU GPL version 3.0";
+        sonatypeCentralPublishExtension {
+            groupId.set("${project.group}")
+            artifactId.set("${project.name}")
+            version.set("${project.version}")
+            componentType.set("java")
+            publishingType.set("AUTOMATIC")
+            username.set(getProperty("sonatypeUser"))
+            password.set(getProperty("sonatypePassword"))
+            pom {
+                name.set("${project.name}")
+                description.set("Data replication")
+                url.set("https://symmetricds.org/")
+                licenses {
+                    license {
+                        name.set(licenseName)
+                    }
+                }
+                developers {
+                    developer {
+                        name.set("Jumpmind")
+                    }
+                }
+                scm {
+                    url.set("https://github.com/jumpmindinc/symmetric-ds/")
+                }
+            }
+        }
     }
 
     task deploy {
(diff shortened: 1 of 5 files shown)


symmetric-jdbc/src/main/java/org/jumpmind/db/platform/db2/Db2DatabasePlatform.java
evan-miller-jumpmind, line 1: DB2 [doesn't support catalog](https://downloads.jumpmind.com/symmetricds/doc/3.16/html/user-guide.html#_catalog_and_schema), so the current implementation is correct. See `OracleDatabasePlatform.getDefaultCatalog()` for another example. The issue for SYM-7347 isn't that the Source Catalog dropdown is empty, it's that the Source Schema dropdown should be displayed instead.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/db2/Db2DdlReader.java
```diff
@@ -324,14 +332,86 @@ protected Integer mapUnknownJdbcTypeForColumn(Map<String, Object> values) {
...
+            return Types.LONGVARCHAR;
         } else if (typeName != null && typeName.endsWith("XML")) {
             return Types.SQLXML;
         } else {
             return super.mapUnknownJdbcTypeForColumn(values);
         }
     }
 
+    @Override
+    public List<String> getCatalogNames() {
```
evan-miller-jumpmind: We don't need to support getting a list of catalog names for DB2 when it doesn't support catalogs. The catalog dropdown in the web console should be replaced by a schema dropdown.

evan-miller-jumpmind (changes requested):
Instead of creating one PR for six issues, we usually create a separate PR for each issue. It can be confusing to view the history for a code change and see it linked to an unrelated issue.

Also, it would be a good idea to add unit tests for some of this new code.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/db2/Db2DatabasePlatform.java
gp510: Agreed — DB2 doesn't support catalogs. We've updated `getDefaultCatalog()` to return `null` (matching Oracle and AS/400), which also eliminates the need for the `getCatalogNames()`, `readTableFromConnection()`, `getTableNames()`, and `getTableNamesFromDatabase()` overrides in `Db2DdlReader` — those were only needed to work around `""` leaking into JDBC `meta.getTables()` calls where `null` means "don't filter by catalog" but `""` means "tables without a catalog" (returning nothing on DB2).

The AS/400 platform already returns `null` with the comment: _"This must return null for AS400, an empty string will return no match on readTable meta data."_ — same issue, same fix.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/db2/Db2DdlReader.java
gp510: Agreed — the `getCatalogNames()` override has been removed. With `getDefaultCatalog()` now returning `null`, none of the catalog-related overrides in `Db2DdlReader` are needed anymore (`getCatalogNames`, `readTableFromConnection`, `getTableNames`, `getTableNamesFromDatabase`) — all four have been removed.

gp510:
Thanks for the review, Evan. Here's what we've done:

**Unit tests**: Added `Db2DdlReaderTest` with 11 tests covering all the DB2-specific changes — type mappings (LONG VARCHAR, CLOB, DBCLOB, ROWID, XML), `readColumn` NAME fallback, `readForeignKey` fallbacks (REFTBNAME, TBNAME, COLNAME), and `getDefaultCatalog()` returning null.

**Catalog fix simplified**: Rather than working around `""` leaking into JDBC calls, we changed `getDefaultCatalog()` to return `null` (matching Oracle and AS/400). This eliminated 4 override methods in `Db2DdlReader` that were only compensating for the wrong return value, making the PR significantly smaller and cleaner.

**Tested end-to-end** against a DB2 LUW 12.1 → SQL Server POC with field-level transforms. Zero runtime errors — triggers created with proper 2-part names, initial load extracted and delivered successfully, all transforms applied correctly.

Regarding splitting into separate PRs: these fixes are tightly coupled (e.g., the catalog fix and the DDL reader fallbacks both affect table/column reading on DB2), so they're kept together to avoid partial fixes that could break in isolation.

symmetric-assemble/common.gradle
evan-miller-jumpmind, line 1: What is the reason for these changes to `common.gradle`?

gp510: These were accidental local build changes — I had commented out the Sonatype publisher plugin during development. Reverted in 1a4198b.

evan-miller-jumpmind: What about the `toolchain` change and the dependency version downgrades?

gp510: Good catch — the toolchain removal and dependency version downgrades (Spring, Spring Boot, Jetty) were also accidental local build changes. Reverted in c5daf1e.

evan-miller-jumpmind (approved):

PR #592, opened 2026-02-26
SYM-7341: Made multiple Java transforms on the same table work correctly

evan-miller-jumpmind:
The problem is that the `TRANSFORM_KEY` is based on `JavaColumnTransform.hashCode()`, but there isn't a separate instance of `JavaTransformColumn` per transform column. The fix appends the primary key values from `sym_transform_column` to the `TRANSFORM_KEY` so that each transform column gets a unique context key.

This bug also affects 3.16. We discussed fixing this in 3.17.0, but let me know if the fix should be included in 3.16.11 as well.


w3bbd3v-jm (approved):
I'd vote to include the fix in 3.16.

PR #594, opened 2026-02-26
SYM-7342: postgres ddl drop column support.

w3bbd3v-jm:
I think this would be good for 3.16 too. I'll plan on including it unless there are any objections?

### DIFF
--- a/symmetric-client/src/main/java/org/jumpmind/symmetric/db/postgresql/PostgreSqlTriggerTemplate.java
+++ b/symmetric-client/src/main/java/org/jumpmind/symmetric/db/postgresql/PostgreSqlTriggerTemplate.java
@@ -363,7 +363,7 @@ createTriggerCommandBeginning + "$(triggerName) after delete on $(schemaName)$(t
 "rowData = current_query();\n" +
 "for cmd in select * from pg_event_trigger_dropped_objects() loop\n" +
 "    if (upper(cmd.object_identity) not like upper('$(prefixName)%') and upper(cmd.object_identity) not like upper('%.$(prefixName)%') and" + 
-"    upper(cmd.object_identity) not like upper('f$(prefixName)%') and upper(cmd.object_identity) not like upper('%.f$(prefixName)%') and cmd.original) then\n" +
+"    upper(cmd.object_identity) not like upper('f$(prefixName)%') and upper(cmd.object_identity) not like upper('%.f$(prefixName)%') and cmd.original and upper(rowData) not like 'ALTER%') then\n" +
 "        select trigger_hist_id into histId from $(defaultSchema)$(prefixName)_trigger_hist where upper(source_table_name) = upper('$(prefixName)_node') and inactive_time is null;\n" +
 "        insert into $(defaultSchema)$(prefixName)_data\n" +
 "        (table_name, event_type, trigger_hist_id, row_data, channel_id, source_node_id, create_time)\n" +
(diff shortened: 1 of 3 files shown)


symmetric-client/src/main/java/org/jumpmind/symmetric/db/postgresql/PostgreSqlTriggerTemplate.java
```diff
@@ -363,7 +363,7 @@ public PostgreSqlTriggerTemplate(ISymmetricDialect symmetricDialect) {
 "rowData = current_query();\n" +
 "for cmd in select * from pg_event_trigger_dropped_objects() loop\n" +
 "    if (upper(cmd.object_identity) not like upper('$(prefixName)%') and upper(cmd.object_identity) not like upper('%.$(prefixName)%') and" + 
-"    upper(cmd.object_identity) not like upper('f$(prefixName)%') and upper(cmd.object_identity) not like upper('%.f$(prefixName)%') and cmd.original) then\n" +
+"    upper(cmd.object_identity) not like upper('f$(prefixName)%') and upper(cmd.object_identity) not like upper('%.f$(prefixName)%') and cmd.original and upper(rowData) not like 'ALTER%') then\n" +
```
w3bbd3v-jm, line 366: Prevent `ALTER ... DROP` SQL statements from being captured by these PostgreSQL dropped object event triggers. These are already captured by the `ddl_command_end` trigger.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlJdbcSqlTemplate.java
```diff
@@ -46,7 +46,7 @@ public PostgreSqlJdbcSqlTemplate(DataSource dataSource, SqlTemplateSettings sett
         dataTruncationStates = new String[] { "22001" };
         deadlockSqlStates = new String[] { "40P01" };
         objectAlreadyExistsStates = new String[] { "42723", "42P05", "42P06", "42P07", "42712", "42710" };
-        objectDoesNotExistStates = new String[] { "42883", "42P01", "42704" };
+        objectDoesNotExistStates = new String[] { "42883", "42P01", "42704", "42703" };
```
w3bbd3v-jm, line 49: PostgreSQL error code: 42703 	UNDEFINED COLUMN 	undefined_column

evan-miller-jumpmind (approved):

symmetric-client/src/main/java/org/jumpmind/symmetric/db/postgresql/PostgreSqlTriggerTemplate.java
pavel-jm: Would this affect scenarios: 
ALTER TABLE ... RENAME COLUMN 
ALTER TABLE ... ALTER COLUMN ... DROP DEFAULT
ALTER TABLE ... DROP CONSTRAINT...

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlJdbcSqlTemplate.java
pavel-jm: Nice find!

symmetric-client/src/main/java/org/jumpmind/symmetric/db/postgresql/PostgreSqlTriggerTemplate.java
w3bbd3v-jm: Yes, these would already be captured by the `ddl_command_end` trigger. So the fix should prevent them from being replicated as well.

pavel-jm (approved):

PR #596, opened 2026-02-27
SYM-7357: Aligned the oracle.* and tibero.* parameters with the Pro edition

evan-miller-jumpmind:


symmetric-core/src/main/java/org/jumpmind/symmetric/route/DataGapDetector.java
```diff
@@ -99,7 +99,7 @@ public void beforeRouting() {
                 Date date = symmetricDialect.getEarliestTransactionStartTime();
                 if (date != null) {
                     earliestTransactionTime = date.getTime() - parameterService.getLong(
-                            ParameterConstants.DBDIALECT_ORACLE_TRANSACTION_VIEW_CLOCK_SYNC_THRESHOLD_MS, 60000);
+                            ParameterConstants.ROUTING_GAPS_TRANSACTION_VIEW_CLOCK_SYNC_THRESHOLD_MS, 10000);
                 }
```
pavel-jm, line 103: I think the larger value of 60000 is a "safer" default value, because it will match more transactions for gap detection?

symmetric-core/src/main/java/org/jumpmind/symmetric/route/DataGapFastDetector.java
```diff
@@ -152,7 +149,7 @@ protected void reset() {
             Date date = symmetricDialect.getEarliestTransactionStartTime();
             if (date != null) {
                 earliestTransactionTime = date.getTime() - parameterService.getLong(
-                        ParameterConstants.DBDIALECT_ORACLE_TRANSACTION_VIEW_CLOCK_SYNC_THRESHOLD_MS, 60000);
+                        ParameterConstants.ROUTING_GAPS_TRANSACTION_VIEW_CLOCK_SYNC_THRESHOLD_MS, 10000);
                 log.debug("Earliest transaction time is {}", earliestTransactionTime);
```
pavel-jm, line 153: I think the larger value of 60000 is a "safer" default value, because it will match more transactions for gap detection?

symmetric-client/src/test/java/org/jumpmind/symmetric/ClientSymmetricEngineTest.java
```diff
@@ -1,37 +0,0 @@
...
-    @Test
-    void testCreateSqlTemplateSettings_setJdbcLobHandling_withParameter() {
-        TypedProperties properties = new TypedProperties();
-        properties.setProperty(ParameterConstants.DBDIALECT_ORACLE_JDBC_LOB_HANDLING, "streamlob");
-        SqlTemplateSettings settings = ClientSymmetricEngine.createSqlTemplateSettings(properties);
-        assertEquals(JdbcLobHandling.STREAMLOB.name(), settings.getJdbcLobHandling().name());
-    }
-
-    @Test
-    void testCreateSqlTemplateSettings_setJdbcLobHandling_fixTurkeyLocaleIssue() {
```
w3bbd3v-jm, line 29: Can we preserve this Turkey locale issue test somewhere?

evan-miller-jumpmind: When I had Claude try to do this, it said that the original test did not work correctly because `FormatUtils.isInfamousTurkey` is set in a `static` block before the test runs. It wasn't able to create a working test for it. I could take another shot at it though.

w3bbd3v-jm (approved):

PR #599, opened 2026-03-02
SYM-7368: Throw exception early in the initial load process (prevent looping)

pavel-jm:


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
```diff
@@ -651,7 +653,14 @@ public void updateTableReloadRequestsLoadId(ISqlTransaction transaction, long lo
                 close(transaction);
             }
         } else {
-            transaction.prepareAndExecute(sql, args, types);
+            rowsAffected = transaction.prepareAndExecute(sql, args, types);
+        }
+        if (rowsAffected < 1) {
```
pavel-jm, line 658: Main Change 1 (of 2)
Throw exception whenever record could not be found (to update it)

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
```diff
@@ -1073,12 +1083,11 @@ public Map<Integer, ExtractRequest> insertReloadEvents(Node targetNode, boolean
                         String createBy = reverse ? nodeSecurity.getRevInitialLoadCreateBy() : nodeSecurity.getInitialLoadCreateBy();
                         if (reloadRequests != null && reloadRequests.size() > 0) {
                             createBy = reloadRequests.get(0).getLastUpdateBy();
-                            createTableReloadStatus(platform.supportsMultiThreadedTransactions() ? null : transaction,
-                                    loadId, isFullLoad, reloadRequests.get(0).getSourceNodeId(), reloadRequests.get(0).getTargetNodeId());
+                            // These three actions must share transaction: new table_reload_status, update original table_reload_request and node status:
```
pavel-jm, line 1081: Main Change 2 (of 2)
Share same transaction for three actions - to enforce consistency:
* createTableReloadStatus
* updateTableReloadRequestsLoadId
* setPartialLoadStarted

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
```diff
@@ -1243,6 +1252,16 @@ public Map<Integer, ExtractRequest> insertReloadEvents(Node targetNode, boolean
         return extractRequests;
     }
 
+    private long generateNewLoadId(ISqlTransaction transaction) {
```
pavel-jm, line 1250: Small refactoring - give a readable method name (used in insertReloadEvents code above) to this IF block

philipmarzullo64 (approved):

w3bbd3v-jm (approved):

PR #600, opened 2026-03-02
SYM-7307: Redact SMTP Password Parameter

Sai-Gumpalli:


symmetric-client/src/main/java/org/jumpmind/symmetric/util/SnapshotUtil.java
```diff
@@ -115,6 +115,10 @@ public class SnapshotUtil {
     protected static final int THREAD_INDENT_SPACE = 50;
     public static final String SNAPSHOT_DIR = "snapshots";
     public static final String ERROR_BATCHES_SUBDIR = "batches";
+    public static final String REDACTED = "***REDACTED***";
```
w3bbd3v-jm: Do these need to be `public`? If not we should keep the encapsulated with `private`.

symmetric-client/src/main/java/org/jumpmind/symmetric/util/SnapshotUtil.java
```diff
@@ -1193,6 +1149,64 @@ protected static Map<String, List<TransformTable>> toMap(List<TransformTableNode
...
+                }
+            }
+            redactParameters(changedParameters);
+            writeProperties(changedParameters, tmpDir, "parameters-changed.properties");
+        } catch (Exception e) {
+            log.warn("Failed to export parameters-changed information", e);
+        }
+    }
+
+    private static void redactParameters(Properties parameters) {
```
w3bbd3v-jm: Could we make this a separate utility? Snapshot utility doesn't need to know anything about redacting.

symmetric-client/src/main/java/org/jumpmind/symmetric/util/SnapshotUtil.java
pavel-jm: Public is good for code re-use, but this conversation made me think of a new (better) place for this constant: ParameterConstants class. I can see us re-using this value during an import process (in the future) to check whether incoming value had been redacted or not?

symmetric-client/src/main/java/org/jumpmind/symmetric/util/SnapshotUtil.java
pavel-jm: Sure! Something like ParameterUtils class might be helpful?

Sai-Gumpalli: I made a RedactionUtils class with just the redactParameters method. Should I instead make a ParameterUtils class that has the writeRuntimeParameters, writeChangedParameters and redactParameters methods?

w3bbd3v-jm: I think what you have is a nice improvement. Could you add a unit test for the new class/method, please?

I'd argue that the redact constants could be kept in your new util class as private. Then we could add a public method later for `public boolean isRedacted(...)`, as needed.

w3bbd3v-jm (approved):

symmetric-client/src/main/java/org/jumpmind/symmetric/util/ParametersUtil.java
```diff
@@ -0,0 +1,24 @@
+package org.jumpmind.symmetric.util;
```
w3bbd3v-jm, line 21: I think we add the copyright header to new files in the open-source project.

symmetric-client/src/test/java/org/jumpmind/symmetric/util/ParametersUtilTest.java
```diff
@@ -0,0 +1,45 @@
+package org.jumpmind.symmetric.util;
```
w3bbd3v-jm, line 21: I think we add the copyright header to new files in the open-source project.

pavel-jm (approved):

PR #603, opened 2026-03-02
SYM-7364: Aligned the mssql.* parameters with the Pro edition

evan-miller-jumpmind:


symmetric-core/src/main/java/org/jumpmind/symmetric/load/AbstractDataLoaderFactory.java
```diff
@@ -49,7 +50,7 @@ public DatabaseWriterSettings buildParameterDatabaseWriterSettings(List<? extend
 
     public DatabaseWriterSettings buildParameterDatabaseWriterSettings(final ISymmetricDialect symmetricDialect,
             List<? extends Conflict> conflictSettings) {
-        DatabaseWriterSettings settings = new DatabaseWriterSettings();
+        DatabaseWriterSettings settings = AppUtils.newInstance(DatabaseWriterSettings.class, DatabaseWriterSettings.class);
```
w3bbd3v-jm, line 53: Remind me why we need to do this?

w3bbd3v-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/load/AbstractDataLoaderFactory.java
evan-miller-jumpmind: `DatabaseWriterSettings` had a `msSqlBulkLoadBcpCodePage` variable, so I moved it to a new `ProDatabaseWriterSettings` class to align SQL Server with Pro. `AppUtils.newInstance()` will return a `DatabaseWriterSettings` object when running open-source and a `ProDatabaseWriterSettings` object when running Pro.

PR #604, opened 2026-03-02
SYM-7368: Add log messages around queuing new loads

pavel-jm:


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/InitialLoadService.java
```diff
@@ -74,36 +74,54 @@ public InitialLoadService(ISymmetricEngine engine) {
...
+            log.debug("Not queuing new loads because cluster lock was not acquired. " + nodeInfo);
+            return;
+        }
+        ProcessInfo processInfo = null;
+        try {
+            processInfo = engine.getStatisticManager().newProcessInfo(
+                    new ProcessInfoKey(identity.getNodeId(), null, ProcessType.INSERT_LOAD_EVENTS));
+            log.debug("Queuing initial loads for " + nodeInfo);
+            processInfo.setStatus(ProcessInfo.ProcessStatus.PROCESSING);
+            queueLoadsSynchronized(identity, processInfo);
```
pavel-jm, line 103: This is the only change (the rest is refactoring existing code) .
This change minimizes amount of time callers spend waiting for the Synchronized call.
For example, if a cluster lock is already in place, then there is no need to block threads - just exit this method!

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/InitialLoadService.java
```diff
@@ -74,36 +74,54 @@ public InitialLoadService(ISymmetricEngine engine) {
...
+        }
+        if (!force && !engine.getClusterService().lock(ClusterConstants.INITIAL_LOAD_QUEUE)) {
+            log.debug("Not queuing new loads because cluster lock was not acquired. " + nodeInfo);
+            return;
+        }
+        ProcessInfo processInfo = null;
+        try {
+            processInfo = engine.getStatisticManager().newProcessInfo(
+                    new ProcessInfoKey(identity.getNodeId(), null, ProcessType.INSERT_LOAD_EVENTS));
+            log.debug("Queuing initial loads for " + nodeInfo);
```
evan-miller-jumpmind: Is it just me or could this log message cause confusion? If I send a load from node A to node B, then I would expect to see node A log a message like "Queuing initial loads for B" since B is the recipient of the load. However, the `nodeInfo` comes from the `identity` node, which would be node A in my example. I know we don't know the target node IDs at this point, but the wording seems confusing to me. The same applies to the below log message on line 107.

pavel-jm: Yep!
I've updated log messages in InitialLoadService

evan-miller-jumpmind (approved):

PR #611, opened 2026-03-04
SYM-7330: Add Documentation For the Rest Endpoint

Sai-Gumpalli:


symmetric-assemble/src/asciidoc/appendix/rest.ad
```diff
@@ -0,0 +1,56 @@
+
+=== Rest
+
+Send changes from your relational database to a Rest endpoint in a XML or JSON.  A RabbitMQ node can be setup as a <<Write-Only Node>> to receive changes from another node that is capturing changes.
```
evan-miller-jumpmind: 1. I think the 1st sentence would be more readable if it ended with "in XML or JSON format"
2. There's a reference to RabbitMQ here

symmetric-assemble/src/asciidoc/appendix/rest.ad
```diff
@@ -0,0 +1,56 @@
...
+
+ifdef::pro[]
+Setup the Rest node by using the configure canvas and selecting Rest as the type.   The URL will be the connection point to RabbitMQ.
```
evan-miller-jumpmind: 1. There's a small grammar error here. The 1st sentence should start with "Set up" instead of "Setup".
2. We renamed the Configure Canvas to the Configure Diagram in 3.17
3. There's another reference to RabbitMQ here

evan-miller-jumpmind (changes requested):

symmetric-assemble/src/asciidoc/appendix/rest.ad
mdrouard-jumpmind: RabbitMq is mentioned. These instances should be replaced with REST

mdrouard-jumpmind: Resolving, Evan has already made a comment for RabbitMq

evan-miller-jumpmind (approved):

symmetric-assemble/src/asciidoc/appendix/rest.ad
```diff
@@ -0,0 +1,56 @@
+
+=== Rest
+
+Send changes from your relational database to a Rest endpoint in a XML or JSON format.  A Rest node can be setup as a <<Write-Only Node>> to receive changes from another node that is capturing changes.
```
mdrouard-jumpmind, line 4: I think it would be fine if it was "...in XML or JSON format" removing the "a".

symmetric-assemble/src/asciidoc/appendix/rest.ad
```diff
@@ -0,0 +1,56 @@
...
+----
+
+==== Output Message Format
+
+.Set the output message format with the following property
+----
+Rest.writer.format=JSON|XML
+----
+
+[horizontal]
```
mdrouard-jumpmind, line 30: Format is not the same for JSON and XML blocks

symmetric-assemble/src/asciidoc/appendix/rest.ad
```diff
@@ -0,0 +1,56 @@
...
+Further settings can be configured in the deploy wizard.
+
+image::images/appendix/rest-advanced-settings.png[]
+endif::pro[]
+
+==== API Key
+
+.Define an API key with the following property.
+----
+Rest.writer.api.key
```
mdrouard-jumpmind, line 20: parameters should be lower case

symmetric-assemble/src/asciidoc/appendix/rest.ad
mdrouard-jumpmind: I think this is due to a lack of a [horizontal] tag above the XML block

PR #612, opened 2026-03-04
SYM-7330: Rest Documentation Fixes

Sai-Gumpalli:


symmetric-assemble/src/asciidoc/appendix/rest.ad
```diff
@@ -19,7 +19,7 @@ endif::pro[]
 ----
 Rest.writer.api.key
```
mdrouard-jumpmind: The parameters should probably start with a lower case letter. That seems to be the common format

pavel-jm (approved):

symmetric-assemble/src/asciidoc/appendix/rest.ad
pavel-jm: Agreed, all SymmetricDS parameters are in lowercase

w3bbd3v-jm (approved):

mdrouard-jumpmind (approved):

PR #617, opened 2026-03-06
SYM-7317: Fixed log message when registering using internal communication

evan-miller-jumpmind:


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataLoaderService.java
```diff
@@ -283,14 +283,14 @@ public void loadDataFromPull(Node remote, RemoteNodeStatus status) throws IOExce
                         requestProps.putAll(props);
                     }
                 }
-                transport = transportManager.getRegisterTransport(local,
-                        parameterService.getRegistrationUrl(), requestProps);
-                log.info("Using registration URL of {}", transport.getUrl());
+                transport = transportManager.getRegisterTransport(local, registrationUrl, requestProps);
+                String transportUrl = StringUtils.defaultIfBlank(transport.getUrl(), registrationUrl);
+                log.info("Using registration URL of {}", transportUrl);
                 for (INodeRegistrationListener l : registrationListeners) {
```
w3bbd3v-jm: I know this isn't part of your code, but since we're already in here...Can we change `l` to something more descriptive? i.e. `regListener` or even `listener`?

w3bbd3v-jm (approved):

PR #623, opened 2026-03-09
SYM-7147: Prevented config import from replacing a valid license key with an invalid license key

evan-miller-jumpmind:


symmetric-client/src/main/java/org/jumpmind/symmetric/SymmetricAdmin.java
```diff
@@ -507,6 +513,11 @@ private void importConfig(CommandLine line, List<String> args) {
                 }
                 String sql = SymmetricUtils.removeInvalidTablesFromImportedSql(getSymmetricEngine().getTablePrefix(),
                         new SqlScriptReader(new InputStreamReader(url.openStream(), StandardCharsets.UTF_8.name())));
+                IConfigImportInterceptor interceptor = getSymmetricEngine().getExtensionService()
```
w3bbd3v-jm: Since the code changes are  basically the same in this class, we could pull the code into a separate method, then call it twice. This would reduce the cognitive complexity score Sonar is complaining about.

evan-miller-jumpmind: Interesting, I haven't seen an inline Sonar comment like that on any PR in the past few months. I moved the code into a separate method.

w3bbd3v-jm (approved):

PR #627, opened 2026-03-11
SYM-7415: Add :SOURCE_SCHEMA and :SOURCE_CATALOG tokens to column router

gp510:
## Summary
- Add `SOURCE_SCHEMA` and `SOURCE_CATALOG` as virtual columns (left-side) and tokens (right-side) in the column match data router
- Enables routing by source schema/catalog from `TriggerHistory` using expressions like `SOURCE_SCHEMA=:EXTERNAL_ID`
- Replaces the fragile MSSQL-specific `log.miner.mssql.populate.external.data.with.schema` workaround (removed in companion symmetric-pro PR)

## Test plan
- [x] 3 new unit tests added: schema equals, schema not-equals, catalog equals
- [x] All 12 `ColumnMatchDataRouterTest` tests pass
- [ ] Manual verification with wildcard schema triggers on MSSQL

### DIFF
--- a/symmetric-assemble/src/asciidoc/configuration/routers/column.ad
+++ b/symmetric-assemble/src/asciidoc/configuration/routers/column.ad
@@ -8,8 +8,10 @@ Column routers are configured by setting the router_type column on the ROUTER ta
 the router_expression column to an equality expression that represents the expected value of the column.
 endif::pro[]
 
-The first part of the expression is always the column name. The column name should always be defined in upper case. 
+The first part of the expression is typically a column name, defined in upper case.
 The upper case column name prefixed by OLD_ can be used for a comparison being done with the old column data value.
+The virtual column tokens :SOURCE_SCHEMA and :SOURCE_CATALOG can also be used in place of a column name to match
+against the source schema or catalog from the trigger history.
 
 The second part of the expression can be a constant value, a token that represents another column, or a token that 
 represents some other SymmetricDS concept. Token values always begin with a colon (:).
--- a/symmetric-assemble/src/asciidoc/configuration/routers/column.ad
+++ b/symmetric-assemble/src/asciidoc/configuration/routers/column.ad
@@ -72,6 +72,14 @@ endif::pro[]
 * :REDIRECT_NODE
 ====
 
+[TIP]
+.Virtual columns that resolve from trigger history
+====
+
+* :SOURCE_SCHEMA
+* :SOURCE_CATALOG
+====
+
 . Consider a table that needs to be routed to only nodes in the target group whose STORE_ID column matches the external id of a node. 
 
 ifdef::pro[]
@@ -163,6 +171,29 @@ insert into SYM_ROUTER (router_id,
 	('corp-2-store-multiple-matches','corp', 'store', 'column',
 	'STORE_ID=NULL or STORE_ID=:EXTERNAL_ID', current_timestamp,
 	current_timestamp); 
+----
+endif::pro[]
+
+. Consider a table captured with a wildcard schema trigger (source_schema_name=*) that needs to route to nodes whose external id matches the source schema name.
+
+ifdef::pro[]
+.Router Expression
+
+----
+:SOURCE_SCHEMA=:EXTERNAL_ID
+----
+endif::pro[]
+
+ifndef::pro[]
+.The following SQL statement will insert a column router to accomplish that.
+[source, SQL]
+----
+insert into SYM_ROUTER (router_id,
+	source_node_group_id, target_node_group_id, router_type,
+	router_expression, create_time, last_update_time) values
+	('corp-2-store-by-schema','corp', 'store', 'column',
+	':SOURCE_SCHEMA=:EXTERNAL_ID', current_timestamp,
+	current_timestamp);
 ----
 endif::pro[]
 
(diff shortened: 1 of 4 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/route/ColumnMatchDataRouter.java
```diff
@@ -122,6 +131,20 @@ public Set<String> routeToNodes(SimpleRouterContext routingContext,
...
+                            nodeIds = runExpression(e, columnValue, sourceSchema, nodes,
+                                    nodeIds, node);
+                        }
+                    } else if (value.equalsIgnoreCase(TokenConstants.SOURCE_CATALOG)) {
+                        String sourceCatalog = dataMetaData.getTriggerHistory() != null
+                                ? dataMetaData.getTriggerHistory().getSourceCatalogName() : null;
+                        for (Node node : nodes) {
+                            nodeIds = runExpression(e, columnValue, sourceCatalog, nodes,
+                                    nodeIds, node);
+                        }
```
evan-miller-jumpmind: Are the checks for `value.equalsIgnoreCase(TokenConstants.SOURCE_SCHEMA)` and `value.equalsIgnoreCase(TokenConstants.SOURCE_CATALOG)` necessary? These would only be necessary if we wanted to support an expression of something like `STATUS=:SOURCE_SCHEMA`, which I'm not sure would make sense.

symmetric-core/src/main/java/org/jumpmind/symmetric/route/ColumnMatchDataRouter.java
```diff
@@ -89,6 +91,13 @@ public Set<String> routeToNodes(SimpleRouterContext routingContext,
                     String column = e.getTokens()[0].trim();
                     String value = e.getTokens()[1];
                     String columnValue = columnValues.get(column);
+                    if (columnValue == null && "SOURCE_SCHEMA".equalsIgnoreCase(column)) {
+                        columnValue = dataMetaData.getTriggerHistory() != null
+                                ? dataMetaData.getTriggerHistory().getSourceSchemaName() : null;
+                    } else if (columnValue == null && "SOURCE_CATALOG".equalsIgnoreCase(column)) {
+                        columnValue = dataMetaData.getTriggerHistory() != null
+                                ? dataMetaData.getTriggerHistory().getSourceCatalogName() : null;
+                    }
```
evan-miller-jumpmind, line 101: This would prevent users from routing based on the value of a column called `SOURCE_SCHEMA` or `SOURCE_CATALOG`. I know this is an edge case, but maybe it would be better to use the `TokenConstants` here so they're prefixed with a colon.

symmetric-core/src/test/java/org/jumpmind/symmetric/route/ColumnMatchDataRouterTest.java
evan-miller-jumpmind, line 1: Can you resolve the Sonar warnings in the test class? I'm not as concerned about the warning in `ColumnMatchDataRouter` since the `routeToNodes()` method probably would have already been detected as a "Brain Method" before your changes.

evan-miller-jumpmind (dismissed):
In addition to my comments, can you add documentation for the new tokens? The documentation is in the `column.ad` file.

symmetric-core/src/main/java/org/jumpmind/symmetric/route/ColumnMatchDataRouter.java
gp510: Agreed — removed these branches. The practical use case is always `:SOURCE_SCHEMA` on the left side (e.g. `:SOURCE_SCHEMA=:EXTERNAL_ID`).

symmetric-core/src/main/java/org/jumpmind/symmetric/route/ColumnMatchDataRouter.java
gp510: Good catch. Changed to use `TokenConstants.SOURCE_SCHEMA` / `TokenConstants.SOURCE_CATALOG` for the comparison, so the expression syntax is now `:SOURCE_SCHEMA=:EXTERNAL_ID` — no collision with actual column names.

symmetric-core/src/test/java/org/jumpmind/symmetric/route/ColumnMatchDataRouterTest.java
gp510: Done — removed `public` modifier from the 3 new test methods to fix the Sonar S5786 warnings.

gp510:
Added documentation for the new tokens in `column.ad` — includes a TIP block listing `:SOURCE_SCHEMA` and `:SOURCE_CATALOG` as virtual columns, plus a full example showing wildcard schema routing with `:SOURCE_SCHEMA=:EXTERNAL_ID`.

symmetric-assemble/src/asciidoc/configuration/routers/column.ad
evan-miller-jumpmind: Here's one more place in the documentation that needs changed. It's no longer true that the first part of the expression is always the column name.

gp510: Good catch. Updated the intro to say "typically a column name" and mention that `:SOURCE_SCHEMA` and `:SOURCE_CATALOG` virtual column tokens can also be used in place of a column name.

pavel-jm (approved):
Thank you!

evan-miller-jumpmind (approved):

PR #630, opened 2026-03-12
SYM-7147: Created helper class for importing configuration

evan-miller-jumpmind:
This PR also fixes SYM-7259.

### DIFF
--- a/symmetric-client/src/test/java/org/jumpmind/symmetric/util/ConfigImportHelperTest.java
+++ b/symmetric-client/src/test/java/org/jumpmind/symmetric/util/ConfigImportHelperTest.java
@@ -1,0 +1,89 @@
+/**
+ * Licensed to JumpMind Inc under one or more contributor
+ * license agreements.  See the NOTICE file distributed
+ * with this work for additional information regarding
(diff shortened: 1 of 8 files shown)


symmetric-client/src/test/java/org/jumpmind/symmetric/util/ConfigImportHelperTest.java
```diff
@@ -0,0 +1,89 @@
...
+            String exported = helper.exportConfigAsSql();
+            assertTrue(exported.contains("Modified"));
+        }
+    }
+
+    @Test
+    void closeReleasesResources() throws IOException {
+        ConfigImportHelper helper = new ConfigImportHelper("sym");
+        helper.close();
+        // no exception means success
```
w3bbd3v-jm: You can use something like `assertDoesNotThrow(()` for these.

pavel-jm (approved):
Please see also test-related suggestion from  Andy

evan-miller-jumpmind:
The Sonar comment in `AbstractSymmetricEngine` is incorrect. The `fileUrl` argument is used in `ClientSymmetricEngine`'s implementation of the method.

PR #637, opened 2026-03-13
SYM-7099: Put indexes and foreign keys in separate batches when an initial load defers constraints so that tables don't need sorted in foreign key order

evan-miller-jumpmind:
Claude told me it would be impractical to create unit tests for `DataExtractorService.checkSendDeferredForeignKeys()` and `DataExtractorService.checkSendDeferredIndexes()` because "the mocking burden is disproportionate to the value". I decided to not have it create any unit tests for these methods, but let me know if I should.


evan-miller-jumpmind:
I'm working on addressing the Sonar comments. Some of them will require some refactoring.

pavel-jm (approved):

PR #643, opened 2026-03-18
SYM-7412: Handled UniqueKeyException during routing after correcting overlapping gaps

evan-miller-jumpmind:
Currently the fix will be included in 3.16 and 3.17. Let me know if it should also be included in 3.15.


evan-miller-jumpmind:
I didn't address the Sonar comment on `DataGapFastDetector.afterRouting()` because I didn't change this method at all. If I need to address this comment, let me know.

pavel-jm (approved):

PR #652, opened 2026-03-20
SYM-7433: 3.18 version marker

pavel-jm:


evan-miller-jumpmind (approved):

symmetric-assemble/gradle.properties
w3bbd3v-jm: I think we should update this too, to 3.17.0.

pavel-jm: Thanks! Updated

PR #658, opened 2026-03-24
SYM-7450: Use OBJECT_ID() for SQL Server estimated row count

gp510:
## Summary

- Replace slow 3-way JOIN on `sys.tables`/`sys.partitions`/`sys.schemas` with a single `OBJECT_ID()` lookup in `MsSql2008DatabasePlatform.getEstimatedRowCount()`
- Use `Table.getQualifiedTableName()` for consistent identifier quoting
- Return `-1` via `ISNULL()` when table not found, so the caller falls back to `count(*)`

## Problem

During initial load queue processing, `getEstimatedRowCount()` is called once per table. On SQL Server databases with many schemas, the 3-way JOIN takes **110–184 seconds per table**. With 72 tables this means 2+ hours of sequential row count queries, freezing the UI and blocking all other SymmetricDS operations (routing, batch monitoring, node security).

Customer (AutoCrib/Snap-on) confirmed the `OBJECT_ID()` alternative runs instantly on their database.

## Test plan

- [ ] Deploy patched build to AutoCrib dev environment (DEV41SYNC)
- [ ] Trigger an initial load for node 11498 (72 tables)
- [ ] Verify row count queries complete in <1 second per table (check logs for absence of "Long Running" messages)
- [ ] Verify UI remains responsive during load creation
- [ ] Verify estimated row counts match actual table sizes

Jira: [SYM-7450](https://jumpmind.atlassian.net/browse/SYM-7450)

[SYM-7450]: https://jumpmind.atlassian.net/browse/SYM-7450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSql2008DatabasePlatform.java
+++ b/symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSql2008DatabasePlatform.java
@@ -24,6 +24,7 @@ import javax.sql.DataSource;
 
 import org.apache.commons.lang3.StringUtils;
 import org.jumpmind.db.model.Table;
+import org.jumpmind.db.platform.DatabaseInfo;
 import org.jumpmind.db.platform.DatabaseNamesConstants;
 import org.jumpmind.db.platform.IDdlBuilder;
 import org.jumpmind.db.platform.PermissionResult;
@@ -55,18 +56,17 @@ public class MsSql2008DatabasePlatform extends MsSql2005DatabasePlatform {
 
     @Override
     public long getEstimatedRowCount(Table table) {
-        String catalog = StringUtils.isNotBlank(table.getCatalog()) ? table.getCatalog() : "";
-        if (catalog.length() > 0) {
-            if (getDdlBuilder().isDelimitedIdentifierModeOn()) {
-                catalog = getDdlBuilder().getDatabaseInfo().getDelimiterToken() + catalog + getDdlBuilder().getDatabaseInfo().getDelimiterToken();
-            }
-            catalog = catalog + ".";
+        DatabaseInfo dbInfo = getDatabaseInfo();
+        String quote = getDdlBuilder().isDelimitedIdentifierModeOn() ? dbInfo.getDelimiterToken() : "";
+        String qualifiedName = table.getQualifiedTableName(quote, dbInfo.getCatalogSeparator(), dbInfo.getSchemaSeparator());
+        String catalogPrefix = "";
+        if (StringUtils.isNotBlank(table.getCatalog())) {
+            catalogPrefix = (quote.length() > 0 ? quote + table.getCatalog() + quote : table.getCatalog()) + ".";
         }
-        return getSqlTemplateDirty().queryForLong("select sum(p.rows) from " + catalog + "sys.tables t inner join " +
-                catalog + "sys.partitions p on t.object_id = p.object_id and p.index_id IN (0, 1) inner join " + catalog
-                + "sys.schemas s on t.schema_id = s.schema_id " +
-                "where t.name = ? and s.name = ?",
-                table.getName(), table.getSchema());
+        return getSqlTemplateDirty().queryForLong(
+                "SELECT ISNULL(SUM(rows), -1) FROM " + catalogPrefix + "sys.partitions " +
+                "WHERE object_id = OBJECT_ID(?) AND index_id IN (0, 1)",
+                qualifiedName);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSql2008DatabasePlatform.java
```diff
@@ -55,18 +56,17 @@ public String getName() {
...
     public long getEstimatedRowCount(Table table) {
-        String catalog = StringUtils.isNotBlank(table.getCatalog()) ? table.getCatalog() : "";
-        if (catalog.length() > 0) {
-            if (getDdlBuilder().isDelimitedIdentifierModeOn()) {
-                catalog = getDdlBuilder().getDatabaseInfo().getDelimiterToken() + catalog + getDdlBuilder().getDatabaseInfo().getDelimiterToken();
-            }
-            catalog = catalog + ".";
+        DatabaseInfo dbInfo = getDatabaseInfo();
+        String quote = getDdlBuilder().isDelimitedIdentifierModeOn() ? dbInfo.getDelimiterToken() : "";
+        String qualifiedName = table.getQualifiedTableName(quote, dbInfo.getCatalogSeparator(), dbInfo.getSchemaSeparator());
```
pavel-jm, line 61: Could you check if using the table.getQualifiedTableName() would produce the same result without a need to hard-code catalogPrefix = ""

pavel-jm (changes requested):
Thank for this optimization.
Also, could you fill out the Jira card detail to communicate: Fix versions, Components, Status, Assignee, please.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSql2008DatabasePlatform.java
```diff
@@ -55,18 +56,17 @@ public String getName() {
 
     @Override
     public long getEstimatedRowCount(Table table) {
```
w3bbd3v-jm, line 58: This seems to be another good case for:
1. Establishing baseline (passing) unit test for existing public method. 
2. Changing public method's implementation. 
3. Re-running baseline test against new implementation to ensure no breaking change. 

This is a simple rinse/repeat pattern that will serve us well. Claude can follow this pattern as well.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSql2008DatabasePlatform.java
gp510: Checked — the `catalogPrefix` is still needed because it serves a different purpose than the `qualifiedName`:

- **`qualifiedName`** (from `table.getQualifiedTableName()`) → bind parameter for `OBJECT_ID(?)`, e.g. `"AutoCribNet"."11498"."Pack"`
- **`catalogPrefix`** → prefixed onto `sys.partitions` in the FROM clause, e.g. `"AutoCribNet".sys.partitions`

`sys.partitions` is scoped to the current database. Without the catalog prefix, a cross-database table would resolve via `OBJECT_ID()` to an object_id in the target database, but `sys.partitions` would search the *current* database's partitions — no match.

So both are needed. The `qualifiedName` tells `OBJECT_ID()` which table to find; the `catalogPrefix` tells SQL Server which database's `sys.partitions` to query.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSql2008DatabasePlatform.java
gp510: Added in 547b39b — `MsSql2008DatabasePlatformTest` with 5 tests covering:

1. Catalog + schema  
2. Schema only, no catalog
3. Table name only
4. Delimited mode off (unquoted identifiers)
5. Table-not-found returns -1

Uses a mocked `ISqlTemplate` with `ArgumentCaptor` to verify the exact SQL and bind parameter for each scenario. Follows the pattern from `MsSql2000DatabasePlatformTest` and `MsSql2008DdlReaderTest`.

gp510:
@pavel-jm Jira card updated:
- **Status**: Ready for QE
- **Assignee**: Glen Patzlaff
- **Components**: dialect: sql-server
- **Fix Versions**: 3.16.9, 3.17.1, 3.18.0

Also added unit tests (5 scenarios) in the latest push and replied to both review comments. Let me know if anything else is needed.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSql2008DatabasePlatform.java
w3bbd3v-jm: Nice! Thank you! My only followups are:

- We don't use `org.junit.jupiter.api.DisplayName;` while it doesn't hurt it's not something we have considered using and is outside convention. If we want to start using it, I would make it a separate discussion.
- Simplify comment `/**Unit tests for {@link MsSql2008DatabasePlatform}*/`
- Remove any kind of customer reference

pavel-jm (approved):

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/mssql/MsSql2008DatabasePlatform.java
pavel-jm: Updated tests and removed customer references

PR #660, opened 2026-03-25
SYM-7446: Observability package and OpenTelemetry

pavel-jm:
This is a new area for us and needs review. Purpose of new classes is to report a new metric (number of reservations) to both internal (MetricsManager) consumer and external OpenTelemetry SDK


symmetric-core/src/main/java/org/jumpmind/symmetric/observability/metrics/HostMetricsService.java
```diff
@@ -0,0 +1,90 @@
...
+    private final Map<String, LongCounter> counters = new ConcurrentHashMap<>();
+    private final Map<String, DoubleHistogram> histograms = new ConcurrentHashMap<>();
+
+    // private HostMetricsService() {
+    // log.debug("Starting Host metrics service");
+    // this.meter = MetricsManager.getGlobalInstance().getOpenTelemetry().getMeter(INSTRUMENTATION_SCOPE);
+    // MetricsManager.getGlobalInstance().register(this);
+    // log.debug("Started Host metrics service");
+    // }
+
```
philipmarzullo64: Do we need the commented lines for future use?

philipmarzullo64 (changes requested):

symmetric-assemble/common.gradle
```diff
@@ -343,6 +343,7 @@ subprojects { subproject ->
         jnaVersion = '5.14.0'
         jettyVersion = '12.0.33'
         websocketVersion = '2.2.0'
+        openTelemetryVersion = '1.40.0'
```
w3bbd3v-jm: We should begin with latest release version unless there is a good reason not to. 
[Is this it? https://github.com/open-telemetry/opentelemetry-java/releases/tag/v1.60.1](https://mvnrepository.com/artifact/io.opentelemetry/opentelemetry-bom/1.60.1)

pavel-jm: Thanks! Updated to 1.60.1

pavel-jm:
Thank you for all feedback! I have started a new branch for this project and set up an overview meeting with the team.

PR #666, opened 2026-03-27
SYM-7424: Import Symmetric Properties From Environment Variables

Sai-Gumpalli:


symmetric-assemble/src/asciidoc/setup/engine-properties.ad
```diff
@@ -15,7 +15,14 @@ other properties found in
 can be overridden for a specific engine in an engine's properties file.
 If the properties are changed in
 `conf/symmetric.properties`
-they will take effect across all engines deployed to the server. 
+they will take effect across all engines deployed to the server.
+
+The following properties can also be configured via the environmental variables of a system.
+Environmental variables override properties found in `conf/symmetric.properties` and an engine's properties file across all engines.
+Using environmental variables for properties is not recommended for multi-homed nodes since environmental variables could override engine specific properties.
```
evan-miller-jumpmind: Should we provide an example of how to specify the environment variables here? The parameter names are below, but there's nothing that tells the user how to translate them over to environment variables.

evan-miller-jumpmind (commented):
I just have one comment related to documentation. Otherwise, this looks good so I'll approve it once the SonarCloud comments are addressed.

symmetric-assemble/src/asciidoc/setup/engine-properties.ad
pavel-jm: Agreed, an example would be helpful!

pavel-jm:
> I just have one comment related to documentation. Otherwise, this looks good so I'll approve it once the SonarCloud comments are addressed.

Most of Sonar comments apply to old code, which was moved to a new class.
The only valid comments I've noticed are about Test classes: Remove this 'public' modifier.

evan-miller-jumpmind (approved):

pavel-jm (approved):

PR #670, opened 2026-03-30
SYM-7467: Fixed SqlException during routing or purge when a trigger is configured with a wildcarded source catalog name

evan-miller-jumpmind:
This fix also handles a wildcarded source schema name

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
@@ -106,9 +106,9 @@ import org.jumpmind.symmetric.model.TableReloadRequestKey;
 import org.jumpmind.symmetric.model.TableReloadStatus;
 import org.jumpmind.symmetric.model.Trigger;
 import org.jumpmind.symmetric.model.TriggerHistory;
-import org.jumpmind.symmetric.model.TriggerReBuildReason;
 import org.jumpmind.symmetric.model.TriggerRouter;
 import org.jumpmind.symmetric.service.ClusterConstants;
+import org.jumpmind.symmetric.util.TriggerUtils;
 import org.jumpmind.symmetric.service.IDataService;
 import org.jumpmind.symmetric.service.IExtensionService;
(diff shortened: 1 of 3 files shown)


evan-miller-jumpmind:
I'll address the SonarCloud comment on the `findOrCreateTriggerHistory()` method. The other comments are all on methods that I did not change.

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
```diff
@@ -3664,62 +3664,96 @@ public Data mapRow(Row row) {
...
                     break;
                 }
             }
             if (table != null && trigger != null) {
+                return reconcileTriggerHistory(trigger, table, triggerHistId, data, tableName, isExistingTriggerHist);
+            }
+            return buildStubTriggerHistory(triggerHistId, tableName, data.getDataId());
+        }
+
+        private Table resolveTableForTrigger(Trigger trigger, String tableName) {
```
pavel-jm: This resolveTableForTrigger looks like a brand new method? 
   (Same question for reconcileTriggerHistory )

Would it be possible to move it to some other class, which aligns better with this functionality?
  I'm asking because the DataService class is growing out of control (covers everything and a proverbial kitchen sink). It is likely overdue for a reduction of scope...
Would it make sense to create a new TriggerManager, WildcardTriggerUtils or simply TriggerUtils class?

evan-miller-jumpmind: Sure, I'll move the new methods to a separate class.

evan-miller-jumpmind: Three of the four new methods have been moved to a new `TriggerUtils` class. I don't think it will be possible to move `reconcileTriggerHistory()` to `TriggerUtils` without exceeding the SonarCloud limit of 7 parameters per method. Let me know if it's okay to keep `reconcileTriggerHistory()` in `DataService`.

pavel-jm (approved):

PR #685, opened 2026-04-02
SYM-7167: Increase Queue Name Size

Sai-Gumpalli:


symmetric-core/src/main/resources/symmetric-schema.xml
```diff
@@ -437,7 +437,7 @@
 
     <table name="node_communication" description="This table is used to coordinate communication with other nodes.">
         <column name="node_id" type="VARCHAR" size="50" required="true" primaryKey="true" description="Unique identifier for a node." />
-        <column name="queue" type="VARCHAR" size="25" required="true" default="default" primaryKey="true" description="The queue name to use in relation to the channel." />
+        <column name="queue" type="VARCHAR" size="128" required="true" default="default" primaryKey="true" description="The queue name to use in relation to the channel." />
```
w3bbd3v-jm, line 440: Do we need to update the queue column on other tables as well, i.e. channel?

symmetric-core/src/main/resources/symmetric-schema.xml
pavel-jm: Let's increase size of this queue column entry to 128?

pavel-jm (changes requested):
Please update table definition for the SYM_CHANNEL . Also look for any other tables with the same column...

w3bbd3v-jm (approved):

pavel-jm (approved):

PR #694, opened 2026-04-06
SYM-7495: MySQL create table with auto increment and composite key

w3bbd3v-jm:


symmetric-db/src/main/java/org/jumpmind/db/platform/mysql/MySqlDdlBuilder.java
```diff
@@ -285,6 +285,25 @@ protected boolean shouldGeneratePrimaryKeys(Column[] primaryKeyColumns) {
         return true;
     }
 
+    @Override
+    protected void writeEmbeddedPrimaryKeysStmt(Table table, StringBuilder ddl) {
```
pavel-jm, line 289: If you have a moment, please add a unit test.

pavel-jm (approved):

PR #696, opened 2026-04-07
SYM-7497: Add capture of exported foreign keys to table objects in cache

philipmarzullo64:


evan-miller-jumpmind (commented):
Looks good to me, but there are some SonarCloud warnings to look at. I'm not sure if all of them need fixed, but at least a few of them would be easy to fix.

w3bbd3v-jm (approved):

evan-miller-jumpmind (approved):

PR #704, opened 2026-04-09
SYM-7426: Container Command Line Option

Sai-Gumpalli:


symmetric-client/src/main/java/org/jumpmind/symmetric/AbstractCommandLauncher.java
```diff
@@ -346,5 +349,13 @@ protected void addCommonOption(Options options, String opt, boolean hasArg) {
         options.addOption(opt, null, hasArg, Message.get(COMMON_MESSAGE_KEY_PREFIX + opt));
     }
 
+    protected void setIsContainerized(boolean containerized) {
```
evan-miller-jumpmind: This method name would be easier to read if it was changed to `setContainerized()`. This also lines up with the "Other Information" section of the Jira ticket description.

symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
```diff
@@ -1028,6 +1028,7 @@ public synchronized void stop() {
 
     @Override
     public synchronized void destroy() {
+        log.info("recieved shutdown request");
```
evan-miller-jumpmind: There's a typo here, it should say "Received"

evan-miller-jumpmind (changes requested):
It looks like the only effect of the `--container` option is that an extra message is logged during startup. It seems like more functionality would need implemented to meet the acceptance criteria from the Jira ticket description. Am I missing something?

Sai-Gumpalli:
> It looks like the only effect of the `--container` option is that an extra message is logged during startup. It seems like more functionality would need implemented to meet the acceptance criteria from the Jira ticket description. Am I missing something?

The main change is in the docker file. We have changed the entry point from sym_service to sym --container. This change makes the Java process for symmetric PID 1, as apposed to the tail command being PID 1.

evan-miller-jumpmind (approved):

PR #706, opened 2026-04-09
SYM-7499: MySQL and MariaDB Bulk Writer should fallback when FK cascade on delete exists

philipmarzullo64:


evan-miller-jumpmind (commented):
Since this setting is specific to Pro, would it be better to add it to `ProDatabaseWriterSettings` instead?

philipmarzullo64:
> Since this setting is specific to Pro, would it be better to add it to `ProDatabaseWriterSettings` instead?

It has been moved to ProDatabaseWriterSettings.
Deleting pull request.

philipmarzullo64:
No open source changes for this change, closing pull request.

PR #710, opened 2026-04-10
SYM-7426: Dockerfile - use Java 21 and ensure SymmetricDS is PID 1

pavel-jm:


pavel-jm (commented):
Change entrypoint to ensure SymmetricDS is PID 1 and can receive SIGTERM properly

evan-miller-jumpmind (approved):

PR #717, opened 2026-04-14
SYM-7483: Add hikari property and dependency.

w3bbd3v-jm:


symmetric-core/src/main/resources/symmetric-default.properties
```diff
@@ -86,6 +86,12 @@ db.user=please set me
 # Type: encrypted
 db.password=
 
+# Set to true to use HikariCP as the connection pool instead of the default pool.
+# HikariCP is included in the SymmetricDS distribution and requires no additional dependencies.
+# Tags: database
+# Type: boolean
+db.pool.hikari.enable=false
```
pavel-jm: Could we future-proof this parameter by changing the name/meaning to the TYPE/name of the library?
db.pool.type
(empty) = current Apache DBCP2 lib
(apache_dbcp2) = also Apache DBCP2 lib
hikari = new Hikari library (could become the default value for this parameter in 3.18.0 ?)
(TBD value) = some future kick-ass library for quantum databases on the Moon ;-)

w3bbd3v-jm: Yep, I think that's a good idea. See latest changes to review.

pavel-jm (approved):
Nice!!!

PR #727, opened 2026-04-20
SYM-7446: Observability metrics service, tables and repository

pavel-jm:


symmetric-client/src/main/java/org/jumpmind/symmetric/job/BuiltInJobs.java
```diff
@@ -48,6 +48,8 @@ public List<IJob> getBuiltInJobs(ISymmetricEngine engine, ThreadPoolTaskSchedule
         builtInJobs.add(new OutgoingPurgeJob(engine, taskScheduler));
         builtInJobs.add(new IncomingPurgeJob(engine, taskScheduler));
         builtInJobs.add(new StatisticFlushJob(engine, taskScheduler));
+        builtInJobs.add(new PurgeMetricStatsJob(engine, taskScheduler));
+        builtInJobs.add(new RefreshBacklogReportJob(engine, taskScheduler));
```
evan-miller-jumpmind: Unless I'm missing something, you need to call `createJob(RefreshBacklogReportJob.class, engine, taskScheduler)` here in order for the `ProRefreshBacklogReportJob` to be used when running Pro.

symmetric-client/src/main/java/org/jumpmind/symmetric/job/RefreshBacklogReportJob.java
```diff
@@ -0,0 +1,109 @@
...
+        String currentNodeId = null;
+        String currentStatus = null;
+        long totalBatches = 0;
+        long totalDataRows = 0;
+        long processedEntriesCount = 0;
+        long currentEntriesCount = 0;
+        for (OutgoingBatchSummaryByNodeBriefStats row : stats) {
+            String entryNodeId = row.nodeId();
+            String entryStatus = row.status();
+            if (StringUtils.isBlank(entryNodeId) || StringUtils.isBlank(entryStatus) || entryNodeId.equals("-1")) {
```
evan-miller-jumpmind: Could you use `Constants.UNROUTED_NODE_ID` here instead of `"-1"`?

symmetric-client/src/main/java/org/jumpmind/symmetric/job/RefreshBacklogReportJob.java
```diff
@@ -0,0 +1,109 @@
...
+                continue;
+            }
+            if (!entryNodeId.equals(currentNodeId) || !entryStatus.equals(currentStatus)) {
+                if (currentNodeId != null) {
+                    outgoingBatchMetrics.setBatchAndRowCounts(currentNodeId, currentStatus, totalBatches, totalDataRows);
+                    if (log.isDebugEnabled()) {
+                        log.debug("Recorded backlog dobservation. Node={}, Batch.status={}, batches={}, rows={}, entries={}",
```
evan-miller-jumpmind: There's a typo here.

symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
```diff
@@ -389,6 +393,13 @@ protected void registerSymDSDriver(TypedProperties engineProperties) {
         }
     }
 
+    protected IEngineMetricsService createMetricsService() {
+        return null;
+    }
```
evan-miller-jumpmind, line 433: If `createMetricsService()` is not overridden, then `init()` will throw an NPE. `AndroidSymmetricEngine` does not implement `createMetricsService()`, so it will throw an NPE on startup. In addition to fixing `AndroidSymmetricEngine`, maybe it would be better to make `createMetricsService()` abstract?

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/metrics/MetricDefinitionFactory.java
```diff
@@ -0,0 +1,358 @@
...
+            new ContextDefinition(MetricContext.UNDEFINED, List.of(new MetricAttribute("UNDEFINED", "UNDEFINED"))), // sentinel for attribute-less metrics
+            new ContextDefinition(10000102009L, List.of(new MetricAttribute(CHANNEL, "default"))), // SymmetricDS default channels
+            new ContextDefinition(10000112009L, List.of(new MetricAttribute(CHANNEL, "config"))),
+            new ContextDefinition(10000122009L, List.of(new MetricAttribute(CHANNEL, "system"))),
+            new ContextDefinition(10000132009L, List.of(new MetricAttribute(CHANNEL, "reload"))),
+            new ContextDefinition(10000142009L, List.of(new MetricAttribute(CHANNEL, "heartbeat"))),
+            new ContextDefinition(10000152009L, List.of(new MetricAttribute(CHANNEL, "monitor"))),
+            new ContextDefinition(10000162009L, List.of(new MetricAttribute(CHANNEL, "dynamic"))),
+            new ContextDefinition(10000172009L, List.of(new MetricAttribute(CHANNEL, "filesync"))),
+            new ContextDefinition(10000182009L, List.of(new MetricAttribute(CHANNEL, "filesync_reload"))),
```
evan-miller-jumpmind: Can you use the constants from `Constants.java` for the built-in channel names?

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/metrics/MetricDefinitionFactory.java
```diff
@@ -0,0 +1,358 @@
...
+            new ContextDefinition(10400302009L, List.of(new MetricAttribute(JOB, "report_status"))),
+            new ContextDefinition(10400312009L, List.of(new MetricAttribute(JOB, "routing"))),
+            new ContextDefinition(10400322009L, List.of(new MetricAttribute(JOB, "stage_management"))),
+            new ContextDefinition(10400332009L, List.of(new MetricAttribute(JOB, "stat_flush"))),
+            new ContextDefinition(10400342009L, List.of(new MetricAttribute(JOB, "sync_config"))),
+            new ContextDefinition(10400352009L, List.of(new MetricAttribute(JOB, "sync_triggers"))),
+            new ContextDefinition(10400362009L, List.of(new MetricAttribute(JOB, "watchdog"))),
+            new ContextDefinition(10400372009L, List.of(new MetricAttribute(JOB, "custom_bsh"))),
+            new ContextDefinition(10400382009L, List.of(new MetricAttribute(JOB, "custom_java"))),
+            new ContextDefinition(10400392009L, List.of(new MetricAttribute(JOB, "custom_sql"))),
```
evan-miller-jumpmind: When someone adds a new job, will it be necessary for them to edit this class and add a `ContextDefinition` containing the name of the job? This seems like something that will be easy to forget.

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/metrics/MetricDefinitionFactory.java
```diff
@@ -0,0 +1,358 @@
...
+                            case COUNTER -> service.registerIncreasingCounter(def, channelAttrs);
+                            case DOUBLE_GAUGE -> service.registerDoubleGauge(def, channelAttrs);
+                            case LONG_GAUGE -> service.registerLongGauge(def, channelAttrs);
+                            default -> {
+                                continue;
+                            }
+                        }
+                        count++;
+                    } catch (Exception e) {
+                        log.warn("Failed to pre-register channel-scoped metric {} for channel {}", metricId, channelId, e);
```
evan-miller-jumpmind: This logging won't work correctly because the exception will be treated as a variable to be substituted into the log message. You'll have to use `String.format()` or concatenation for the `metricId` and `channelId`.

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/metrics/MetricsManager.java
```diff
@@ -0,0 +1,278 @@
...
+import io.opentelemetry.api.metrics.Meter;
+import io.opentelemetry.api.metrics.ObservableDoubleGauge;
+import io.opentelemetry.api.metrics.ObservableLongCounter;
+import io.opentelemetry.api.metrics.ObservableLongGauge;
+import io.opentelemetry.api.metrics.ObservableLongUpDownCounter;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk;
+
+/**
+ * The singular and central point of managin both embedded and OpenTelemetry metrics.
```
evan-miller-jumpmind: There's a typo here.

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/models/ObservationDouble.java
evan-miller-jumpmind, line 1: This class is missing the licensing header.

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/models/ObservationLong.java
evan-miller-jumpmind, line 1: This class is missing the licensing header.

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/repository/MetricsRepository.java
```diff
@@ -0,0 +1,872 @@
...
+            transaction = sqlTemplate.startSqlTransaction();
+            for (MetricIntervalStatsRecord statsRecord : records) {
+                saveMetricIntervalStatsInternal(transaction, statsRecord.key(), statsRecord.contextId(), statsRecord.stats());
+            }
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            log.error("Failed to save {} metric interval stats records", records.size(), e);
```
evan-miller-jumpmind: This logging won't work correctly because the exception will be treated as a variable to be substituted into the log message. You'll have to use `String.format()` or concatenation for `records.size()`.

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/stats/MetricIntervalStatsQueue.java
evan-miller-jumpmind, line 1: This class is missing the licensing header.

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/metrics/MetricsManagerTest.java
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    @Test
+    void isOtelPublishingEnabled_propertySetToTrue_returnsTrue() {
+        System.setProperty("metrics.opentelemetry.enabled", "true");
+        try {
+            MetricsManager m = new MetricsManager(OpenTelemetry.noop());
+            assertTrue(m.getHostMetricsService().isOtelPublishingEnabled());
+        } finally {
+            System.clearProperty("metrics.opentelemetry.enabled");
```
evan-miller-jumpmind: Can you use `ParameterConstants.OTEL_METRICS_ENABLED` here (and elsewhere in this file)?

evan-miller-jumpmind (changes requested):
I left some comments regarding some small issues that I found, but otherwise it looks good. It would also be good if you could address the SonarCloud alerts. Not all of them are relevant, like the one in `OutgoingBatchService` which is for code that you didn't change.

.claude/settings.local.json
```diff
@@ -0,0 +1,8 @@
+{
```
w3bbd3v-jm: I don't think we want to commit the  claude `settings.local.json` files. These are used for personal local overrides.

symmetric-assemble/.claude/settings.local.json
```diff
@@ -0,0 +1,7 @@
+{
```
w3bbd3v-jm: I don't think we want to commit the claude settings.local.json files. These are used for personal local overrides.

symmetric-core/src/main/resources/symmetric-schema.xml
```diff
@@ -1017,4 +1025,75 @@
...
+        <column name="attr2_value"     type="VARCHAR"   size="255"  required="false" />
+        <column name="attr3_name"      type="VARCHAR"   size="255"  required="false" />
+        <column name="attr3_value"     type="VARCHAR"   size="255"  required="false" />
+        <column name="create_time"     type="TIMESTAMP" required="true" />
+        <index name="idx_metric_context_hash">
+            <index-column name="attributes_hash" />
+        </index>
+    </table>
+
+    <table name="metric_stats_float64" description="Fact table for floating-point observations aggregated across a time window (per metric). Statistics are time-weighted using a step-function model.">
```
w3bbd3v-jm, line 1065: Can you explain the name `metric_stats_float64`? Wouldn't `metric_stats_double` make more sense?

symmetric-core/src/main/resources/symmetric-schema.xml
```diff
@@ -1017,4 +1025,75 @@
...
+        <column name="avg"                     type="DOUBLE"    required="true"                    description="Time-weighted average value over the interval." />
+        <column name="mean"                    type="DOUBLE"    required="true"                    description="Simple arithmetic mean of observed values during the interval." />
+        <column name="std_dev"                 type="DOUBLE"    required="true"                    description="Time-weighted standard deviation of the value during the interval." />
+        <column name="outlier"                 type="SMALLINT"                                     description="1 if this interval was flagged as a statistical outlier, 0 otherwise." />
+        <index name="idx_metric_stats_float64_key">
+            <index-column name="metric_key" />
+        </index>
+    </table>
+
+    <table name="metric_stats_int64" description="Fact table for integer observations aggregated across a time window (per metric). Statistics are time-weighted using a step-function model. Integer variant: min, max, avg, mean are BIGINT.">
```
w3bbd3v-jm, line 1083: And wouldn't `metric_stats_bigint` or `metric_stats_long`  make more sense?

symmetric-core/src/test/java/org/jumpmind/symmetric/observability/interfaces/InvalidMetricDataExceptionTest.java
```diff
@@ -0,0 +1,70 @@
...
+package org.jumpmind.symmetric.observability.interfaces;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertSame;
+
+import org.junit.jupiter.api.Test;
+
+class InvalidMetricDataExceptionTest {
+    // ── message-only constructor ──────────────────────────────────────────────
```
w3bbd3v-jm: Can we remove these comments...I think the constructor speaks for itself.

symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
pavel-jm: Updated init() method to not trip NPE, and createMetricsService() to log a warning for null MetricsService. 
In the future we'd want to either decommission AndroidSymmetricEngine or add Metrics to it?..

symmetric-core/src/main/resources/symmetric-schema.xml
pavel-jm: Yep!
Double is a Java term, and is equivalent to the 64-bit Floating Point number.
Since customers (DBAs and ops team) can read table names, I was looking for a name which is less programmer-centric for two stats (a.k.a. facts) tables:
sym_metric_stats_float64
sym_metric_stats_int64

The main concern was to avoid "scaring" users into thinking that there's data duplication "double'ing" going on...

One more thought: the term "float64" also specifies precision and storage size. Presumably, in the next decade the "quantum" numbers will be used and they are likely to require more bits to store them.

symmetric-core/src/main/resources/symmetric-schema.xml
pavel-jm: I have considered that also, but was more concerned about creating a pattern covering both tables at the same time (see comment about sym_metric_stats_float64)...

symmetric-core/src/test/java/org/jumpmind/symmetric/observability/interfaces/InvalidMetricDataExceptionTest.java
pavel-jm: Thanks! Done.

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/metrics/MetricDefinitionFactory.java
pavel-jm: Excellent suggestion! Done.

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/metrics/MetricDefinitionFactory.java
pavel-jm: No, it will not be necessary. New contexts get lazy-registered by the MetricsRepository class and saved into database. 
Also, new context is only helpful if the "new job" has some important metrics associated with it - to provide additional/custom analytics.

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/metrics/MetricDefinitionFactory.java
pavel-jm: Great catch! 
Updated

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/metrics/MetricsManager.java
pavel-jm: Thanks! Fixed

.claude/settings.local.json
pavel-jm: OK! Deleted

symmetric-client/src/main/java/org/jumpmind/symmetric/job/BuiltInJobs.java
pavel-jm: Thank you! Updated.

symmetric-client/src/main/java/org/jumpmind/symmetric/job/RefreshBacklogReportJob.java
pavel-jm: Thanks! Eliminated another "magic number"

symmetric-client/src/main/java/org/jumpmind/symmetric/job/RefreshBacklogReportJob.java
pavel-jm: Thanks. No more dobservations

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/models/ObservationLong.java
pavel-jm: Thanks. Added licensing header.

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/models/ObservationDouble.java
pavel-jm: Thanks. Added license header

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/stats/MetricIntervalStatsQueue.java
pavel-jm: Thanks. Addd header

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/repository/MetricsRepository.java
pavel-jm: Thanks. Updated exception logging

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/metrics/MetricsManagerTest.java
pavel-jm: Thanks. Updated

symmetric-assemble/.claude/settings.local.json
pavel-jm: Thanks. Removed

pavel-jm:
> I left some comments regarding some small issues that I found, but otherwise it looks good. It would also be good if you could address the SonarCloud alerts. Not all of them are relevant, like the one in `OutgoingBatchService` which is for code that you didn't change.

I've added a few more unit tests to improve coverage of important areas.

symmetric-stats/build.gradle
```diff
@@ -0,0 +1,19 @@
+apply from: symAssembleDir + '/common.gradle'
+
+    description = 'SymmetricDS statistics and observability implementation.'
+
+    dependencies {
+        api project(':symmetric-core')
+
+        // OpenTelemetry
```
w3bbd3v-jm: Can we remove this comment?
It's clearly related to open telemetry.
If someone adds a different dependency the comment will no longer be truthful, or will prompt the developer to add another comment perpetuating the pattern. In my view it's just not worth it.

symmetric-stats/build.gradle
```diff
@@ -0,0 +1,19 @@
...
+        api platform("io.opentelemetry:opentelemetry-bom:$openTelemetryVersion")
+        api 'io.opentelemetry:opentelemetry-api'
+        implementation 'io.opentelemetry:opentelemetry-sdk'
+        implementation 'io.opentelemetry:opentelemetry-sdk-extension-autoconfigure'
+
+        // Logging
```
w3bbd3v-jm: Can we remove this comment?
Is it really related to logging? How? I could use a util class from it and the comment would no longer be truthful. + same rationale I mentioned on the comment above applies as well.

symmetric-stats/src/test/java/org/jumpmind/symmetric/transport/ConcurrentConnectionManagerTest.java
```diff
@@ -0,0 +1,304 @@
...
+
+    @Test
+    public void Reservation_equals_sameNodeId_returnsTrue() {
+        Reservation a = new Reservation("node1", System.currentTimeMillis() + 10000, ReservationType.HARD);
+        Reservation b = new Reservation("node1", System.currentTimeMillis() + 20000, ReservationType.SOFT);
+        assertEquals(a, b);
+    }
+
+    @Test
+    public void Reservation_equals_differentNodeId_returnsFalse() {
```
w3bbd3v-jm: Are you planning on fixing these sonar warnings and errors? Since it's net new code, I think it would be a good idea where it's practical especially.

symmetric-stats/src/test/java/org/jumpmind/symmetric/transport/ConcurrentConnectionManagerTest.java
```diff
@@ -0,0 +1,304 @@
...
+        assertEquals(0, mgr.getReservationCount("push"));
+    }
+
+    @Test
+    public void getReservationCount_afterReservation_returnsOne() {
+        ConcurrentConnectionManager mgr = newMgr(5);
+        mgr.reserveConnection("node1", "0", "push", ReservationType.HARD, false);
+        assertEquals(1, mgr.getReservationCount("push"));
+    }
+    // ── getPullReservationsByNodeId ────────────────────────────────────────────
```
w3bbd3v-jm: I think Claude adds these comments in the unit tests. Can you remove them please? The test name itself is sufficient for that.

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/stats/Float64StatsAccumulatorTest.java
```diff
@@ -0,0 +1,269 @@
...
+        ISymIntervalStats stats = acc.toStats();
+        assertEquals(T, stats.getStartEpoch());
+        assertEquals(T + D, stats.getEndEpoch());
+    }
+    // ── closeAtObservation ────────────────────────────────────────────────────
+
+    @Test
+    void closeAtObservation_atIntervalEnd_doesNotThrow() {
+        Float64StatsAccumulator acc = new Float64StatsAccumulator(T);
+        acc.closeAtObservation(T + D); // exactly at end: OK
```
w3bbd3v-jm: Add `assertDoesNotThrow` ?

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/stats/PrimaryMetricAggregator.java
```diff
@@ -0,0 +1,147 @@
...
+     * 20-second iteration estimate. An external {@link InterruptedException} causes an immediate exit without final processing.
+     */
+    private void run() {
+        long iterationTargetStart = 0;
+        while (running && !Thread.currentThread().isInterrupted()) {
+            long remainingMs = iterationTargetStart - System.currentTimeMillis();
+            iterationTargetStart = System.currentTimeMillis() + AGGREGATOR_PROCESSING_INTERVAL_MS;
+            try {
+                if (remainingMs > 0) {
+                    log.trace("Sleeping {} thread for {} milliseconds. ThreadId={} ....", remainingMs, AGGREGATOR_PROCESSING_THREAD);
```
w3bbd3v-jm: There's 3 args for the log message but only 2 passed in here.

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/stats/PrimaryMetricAggregator.java
```diff
@@ -0,0 +1,147 @@
...
+                Thread.currentThread().interrupt();
+                log.warn("Thread interrupted " + AGGREGATOR_PROCESSING_THREAD, e);
+                closeAll(); // Interruptions require immediate exit, without delay from processing observations!
+                return;
+            } catch (Exception e) {
+                log.error("Error during metrics processing", e);
+            }
+        }
+        processAll(); // Final processing before exit
+        closeAll(); // Close all open accumulators and metrics before exit
```
w3bbd3v-jm: Rename method to `closeAllAccumulatorsAndMetrics`  instead of adding a comment?

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/stats/MetricSeriesSlidingWorkset.java
```diff
@@ -0,0 +1,206 @@
...
+import org.jumpmind.symmetric.observability.models.MetricSeriesInterquartileRange;
+
+/**
+ * Maintains a sliding workset of {@link ISymIntervalStats} records, computes the interquartile range (IQR) of their {@code mean} values, and detects outliers
+ * using Tukey's fences. Addition of new outliers into the main workset is delayed until the internal outliers buffer is full.
+ * <p>
+ * See also: https://en.wikipedia.org/wiki/Interquartile_range
+ */
+public class MetricSeriesSlidingWorkset {
+    public static final int IQR_OUTLIERS_MAX = 10; // Maximum number of outlier intervals deferred for later adoption into workset (should be less than
```
w3bbd3v-jm: Do these constants need to be part of the public API for this class? After looking at the first 4, they are all used only within this class.

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/stats/MetricSeriesSlidingWorkset.java
```diff
@@ -0,0 +1,206 @@
...
+    private void addToWorkset(ISymIntervalStats interval) {
+        if (intervals.size() >= IQR_INTERVALS_MAX) {
+            intervals.removeLast();
+        }
+        intervals.add(interval);
+        checkExpirationOfOldestOutlier(interval.getStartEpoch() - EXPIRE_OUTLIERS_RATIO * interval.getDurationMillis());
+    }
+
+    /** Adds to the outliers collection. If the outliers buffer is full, then moves all to the current workset (adopts all outliers as new trend). */
+    private void addToOutliers(ISymIntervalStats interval) {
```
w3bbd3v-jm, line 81: This method really does more than one thing well. 
It clones the interval if it's not an outlier. 
It adds the interval to the outliers collection. 
If the outliers buffer is full then it moves then to a workset. 

You could use your `markAsOutlier` method here. 
The other two things it does could be split into separate private methods. 
That way you would not need to explain through the javadoc why the method does more than the method name implies.

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/stats/MetricIntervalStatsQueue.java
```diff
@@ -0,0 +1,254 @@
...
+    public MetricIntervalStatsQueue() {
+        super();
+    }
+
+    private ConcurrentLinkedQueue<ISymIntervalStats> q() {
+        return queue.get();
+    }
+
+    @Override
+    public boolean addAll(Collection<? extends ISymIntervalStats> c) {
```
w3bbd3v-jm: In my experience, single variable names are hard to read and often should be limited to for loops when used. Can we be more descriptive with these? Applies throughout. 
For example it's hard to look at this and know what `c` is without fishing for it.
```
for (ISymIntervalStats item : c) 
```

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/stats/Int64StatsAccumulator.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /** Returns the last observed value at full {@code long} precision. */
+    public long getLastValueAsLong() {
+        return lastValue;
+    }
+
+    @Override
+    public double getLastValueAsDouble() {
+        return lastValue;
```
w3bbd3v-jm, line 56: How does this differ from `getLastValueAsLong`? Semantically these mean very different things in the Java world. Maybe this would be a good spot to map the semantics?

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/repository/SurrogateLongKeyBuffer.java
```diff
@@ -0,0 +1,110 @@
...
+
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Buffer of long surrogate keys allocated via some permanent storage (database) and used for quick allocation of several keys at once before writing records
+ * into permanent storage.
+ */
+public class SurrogateLongKeyBuffer {
+    public static final long SURROGATE_KEY_UNASSIGNED = -1l;
+    public static final long SURROGATE_KEY_BUFFER_SIZE = 10l;
```
w3bbd3v-jm: Do these need to be part of the public API? We have a `capacity` method. Is that good enough for clients to use without knowing these "internals"?

symmetric-stats/build.gradle
pavel-jm: yup!

symmetric-stats/build.gradle
pavel-jm: yup!

symmetric-core/src/test/java/org/jumpmind/symmetric/statistic/StatisticManagerTest.java
```diff
@@ -0,0 +1,428 @@
...
+    void getMostRecentUserProcessInfos_noProcesses_returnsEmpty() {
+        assertTrue(manager.getMostRecentUserProcessInfos(ProcessType.PUSH_JOB_EXTRACT).isEmpty());
+    }
+
+    @Test
+    void getMostRecentUserProcessInfos_activeUserProcess_includesIt() {
+        ProcessInfoKey k = key("src", "tgt", ProcessType.PUSH_JOB_EXTRACT);
+        ProcessInfo p = manager.newProcessInfo(k);
+        p.setCurrentTableName("user_orders");
+        p.setCurrentChannelId("default");
```
evan-miller-jumpmind: You should use `Constants.CHANNEL_DEFAULT` instead of `"default"` here and elsewhere in this file.

symmetric-stats/build/spotless-clean/spotlessJava/src/test/java/org/jumpmind/symmetric/observability/metrics/AbstractQueuedMetricTest.java
evan-miller-jumpmind, line 1: The `symmetric-stats/build` directory should be excluded from version control.

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/metrics/EngineMetricsServiceTest.java
```diff
@@ -0,0 +1,244 @@
...
+        assertTrue(manager.getEngineMetricsServices().contains(service));
+        service.shutdown();
+        assertFalse(manager.getEngineMetricsServices().contains(service));
+    }
+
+    @Test
+    void shutdown_clearsRegisteredMetrics() {
+        EngineMetricsService service = new EngineMetricsService(engine, manager, false);
+        service.registerUpDownCounter(new SymMetricDefinition("s.ud", "d", "r",
+                org.jumpmind.symmetric.observability.interfaces.SymMetricConstants.InstrumentType.UPDOWN_COUNTER));
```
evan-miller-jumpmind: The fully-qualified class name isn't necessary because `InstrumentType` is already being imported.

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/metrics/EngineMetricsServiceTest.java
```diff
@@ -0,0 +1,244 @@
...
+        ISymMetricContext ctx = mock(ISymMetricContext.class);
+        when(metric.getContext()).thenReturn(ctx);
+        when(ctx.getContextId()).thenReturn(42L);
+        assertEquals(42L, service.getOrAssignContextId(metric, mock(MetricsRepository.class)));
+    }
+
+    @Test
+    void getOrAssignContextId_noContextNoAttributes_returnsUndefined() {
+        EngineMetricsService service = new EngineMetricsService(engine, manager, false);
+        UpDownCounter metric = new UpDownCounter("m", io.opentelemetry.api.common.Attributes.empty(), List.of());
```
evan-miller-jumpmind: Could you import `Attributes` instead of using the fully-qualified class name inline? It's used 3 times in this file.

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/metrics/EngineMetricsServiceTest.java
```diff
@@ -0,0 +1,244 @@
...
+        assertEquals(MetricContext.UNDEFINED, service.getOrAssignContextId(metric, null));
+    }
+
+    @Test
+    void getOrAssignContextId_withAttributesAndRepo_registersContextAndReturns() {
+        EngineMetricsService service = new EngineMetricsService(engine, manager, false);
+        MetricAttribute attr = new MetricAttribute("channel", "default");
```
evan-miller-jumpmind: You should use `Constants.CHANNEL_DEFAULT` instead of `"default"` here and elsewhere in this file.

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/metrics/AbstractQueuedMetricTest.java
```diff
@@ -0,0 +1,402 @@
...
+    void createAccumulator_returnsInt64AccumulatorWithCorrectStart() {
+        UpDownCounter m = newCounter();
+        IStatsAccumulator acc = m.createAccumulator(T);
+        assertInstanceOf(Int64StatsAccumulator.class, acc);
+        assertEquals(T, acc.getIntervalStart());
+    }
+
+    @Test
+    void getAttributes_nonEmptyList_returnsProvidedAttributes() {
+        MetricAttribute attr = new MetricAttribute("channel", "default");
```
evan-miller-jumpmind: You should use `Constants.CHANNEL_DEFAULT` instead of `"default"` here.

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/repository/MetricsRepositoryTest.java
```diff
@@ -0,0 +1,610 @@
...
+        assertNotNull(result);
+        assertEquals(88L, result.key());
+        // Verify it was cached
+        Map<Integer, MetricKey> cache = getField(repo, "metricKeysCache");
+        assertTrue(cache.containsValue(result));
+    }
+
+    @Test
+    void getOrRegisterContext_def_inSeedCache_returnsCached() {
+        List<MetricAttribute> attrs = List.of(new MetricAttribute("channel", "default"));
```
evan-miller-jumpmind: You should use `Constants.CHANNEL_DEFAULT` instead of `"default"` here.

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/repository/MetricsRepositoryTest.java
```diff
@@ -0,0 +1,610 @@
...
+        Map<String, MetricContext> seedCache = getField(repo, "seedContextCache");
+        seedCache.put(cacheKey, ctx);
+        ContextDefinition def = new ContextDefinition(1L, attrs);
+        MetricContext result = repo.getOrRegisterContext(def);
+        assertSame(ctx, result);
+    }
+
+    @Test
+    void getOrRegisterContext_def_notInCacheFoundInDb_cachesAndReturns() throws Exception {
+        List<MetricAttribute> attrs = List.of(new MetricAttribute("channel", "reload"));
```
evan-miller-jumpmind: You should use `Constants.CHANNEL_RELOAD` instead of `"reload"` here.

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/repository/MetricsRepositoryValidateAttributesTest.java
evan-miller-jumpmind, line 1: Could these tests be included in `MetricsRepositoryTest` instead of having them in a separate file?

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/repository/MetricsRepositoryValidateAttributesTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+    @Test
+    void emptyList_throwsWithMinMessage() {
+        MetricsRepositoryException ex = assertThrows(MetricsRepositoryException.class, () -> invoke(List.of()));
+        assertTrue(ex.getMessage().contains(String.valueOf(MetricsRepository.ATTR_MIN_VALUES)));
+    }
+
+    @Test
+    void singleValidAttr_doesNotThrow() {
+        assertDoesNotThrow(() -> invoke(List.of(new MetricAttribute("channel", "default"))));
```
evan-miller-jumpmind: You should use `Constants.CHANNEL_DEFAULT` instead of `"default"` here.

symmetric-stats/src/test/java/org/jumpmind/symmetric/transport/ConcurrentConnectionManagerTest.java
pavel-jm: yup!

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/stats/PrimaryMetricAggregator.java
pavel-jm: Thanks! updated

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/stats/MetricSeriesSlidingWorkset.java
pavel-jm: OK. Updated

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/stats/PrimaryMetricAggregator.java
pavel-jm: OK. Refactored

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/stats/MetricSeriesSlidingWorkset.java
pavel-jm: yup. Done

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/stats/MetricIntervalStatsQueue.java
pavel-jm: yup. Done

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/stats/Int64StatsAccumulator.java
pavel-jm: Not sure what you mean? Let's meet when your schedule permits?

symmetric-stats/src/main/java/org/jumpmind/symmetric/observability/repository/SurrogateLongKeyBuffer.java
pavel-jm: yup. Moved to a separate constants class

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/metrics/EngineMetricsServiceTest.java
pavel-jm: Yup. Fixed

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/metrics/EngineMetricsServiceTest.java
pavel-jm: Yup. Fixed

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/metrics/AbstractQueuedMetricTest.java
pavel-jm: Yup. Fixed

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/repository/MetricsRepositoryTest.java
pavel-jm: Yup. Fixed

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/repository/MetricsRepositoryValidateAttributesTest.java
pavel-jm: Yup. Fixed

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/repository/MetricsRepositoryTest.java
pavel-jm: Yup. Fixed

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/repository/MetricsRepositoryValidateAttributesTest.java
pavel-jm: Yup. Done

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/stats/Float64StatsAccumulatorTest.java
pavel-jm: Yup. Done

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/metrics/EngineMetricsServiceTest.java
pavel-jm: Thanks. Done.

symmetric-stats/src/test/java/org/jumpmind/symmetric/transport/ConcurrentConnectionManagerTest.java
pavel-jm: Yup, just fixed a bunch!

symmetric-stats/src/test/java/org/jumpmind/symmetric/observability/metrics/EngineMetricsServiceTest.java
pavel-jm: Thanks! Fixed

symmetric-core/src/test/java/org/jumpmind/symmetric/statistic/StatisticManagerTest.java
```diff
@@ -425,4 +433,407 @@ void getLastDataLoadedRowsMap_initiallyEmpty() {
...
+        when(parameterService.is(anyString(), anyBoolean())).thenReturn(true);
+        when(parameterService.getLong(anyString(), anyLong())).thenReturn(-1L);
+        when(nodeService.getCachedIdentity()).thenReturn(node("node1"));
+        Field field = StatisticManager.class.getDeclaredField("channelStats");
+        field.setAccessible(true);
+        Map<String, ChannelStats> csMap = (Map<String, ChannelStats>) field.get(manager);
+        ChannelStats unknownStats = new ChannelStats("Unknown", "server-1", new Date(), null, "chan1");
+        unknownStats.incrementDataRouted(1);
+        csMap.put("chan1", unknownStats);
+        manager.flush();
```
evan-miller-jumpmind, line 662: This test is missing an assert

symmetric-core/src/test/java/org/jumpmind/symmetric/statistic/StatisticManagerTest.java
```diff
@@ -425,4 +433,407 @@ void getLastDataLoadedRowsMap_initiallyEmpty() {
...
+        assertDoesNotThrow(() -> manager.flush());
+    }
+
+    @Test
+    void flush_hostStatsWithUnknownNodeId_updatesToResolvedNodeId() {
+        when(parameterService.is(anyString(), anyBoolean())).thenReturn(true);
+        when(parameterService.getLong(anyString(), anyLong())).thenReturn(-1L);
+        when(nodeService.getCachedIdentity()).thenReturn(null, node("node1"));
+        manager.incrementRestart();
+        manager.flush();
```
evan-miller-jumpmind, line 685: This test is missing an assert

symmetric-stats/bin/main/org/jumpmind/symmetric/observability/CLAUDE.md
evan-miller-jumpmind, line 1: I think the `bin` directory should be excluded from version control.

evan-miller-jumpmind (changes requested):
I found just 3 more issues. Also, the unit tests are failing.

symmetric-stats/bin/main/org/jumpmind/symmetric/observability/CLAUDE.md
pavel-jm: Agreed! I'll delete this file

symmetric-core/src/test/java/org/jumpmind/symmetric/statistic/StatisticManagerTest.java
pavel-jm: OK. The verify call will fail if condition is not met, but I've added another item to check as well.

symmetric-core/src/test/java/org/jumpmind/symmetric/statistic/StatisticManagerTest.java
pavel-jm: OK. The verify call will fail if condition is not met, but I've added another item to check as well.

evan-miller-jumpmind:
There are 50 SonarCloud alerts and a lot of them seem very easy to fix. Could you take a look?

symmetric-core/src/test/java/org/jumpmind/symmetric/transport/ConcurrentConnectionManagerTest.java
```diff
@@ -0,0 +1,138 @@
...
+        ConcurrentConnectionManager mgr = newMgr(5);
+        mgr.reserveConnection("node1", "0", "push", ReservationType.HARD, false);
+        ReservationStatus status = mgr.reserveConnection("node1", "0", "push", ReservationType.HARD, false);
+        assertEquals(ReservationStatus.DUPLICATE, status);
+    }
+
+    @Test
+    void Reservation_equals_nonReservationObject_returnsFalse() {
+        Reservation r = new Reservation("node1", System.currentTimeMillis() + 10_000, ReservationType.HARD);
+        assertNotEquals(r, "string");
```
w3bbd3v-jm: Sonar is complaining that this is in the wrong order. There is 128 other code maintainability  items to address in SonarQube. I don't think you would need to fix all of them but a lot of them as some are probably linked to existing code. The net new ones for new code we should probably try to fix though and seem small in effort. https://sonarcloud.io/project/issues?id=jumpmindinc_symmetric-ds&pullRequest=727&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

pavel-jm: Thanks. Updated

symmetric-util/src/test/java/org/jumpmind/util/FormatUtilsTest.java
```diff
@@ -3,11 +3,11 @@
  * license agreements.  See the NOTICE file distributed
  * with this work for additional information regarding
  * copyright ownership.  JumpMind Inc licenses this file
- * to you under the GNU General Public License, version 3.0 (GPLv3)
+ * to you under the GNU General License, version 3.0 (GPLv3)
  * (the "License"); you may not use this file except in compliance
  * with the License.
  *
- * You should have received a copy of the GNU General Public License,
+ * You should have received a copy of the GNU General License,
```
evan-miller-jumpmind: It looks like you accidentally removed the word "public" from the license header when removing it elsewhere in the file

evan-miller-jumpmind (approved):

symmetric-util/src/test/java/org/jumpmind/util/FormatUtilsTest.java
pavel-jm: Yup! Thanks

symmetric-util/.gitignore
```diff
@@ -1,6 +1,7 @@
 *.classpath
 *.project
 *.settings
+*.factorypath
```
w3bbd3v-jm: We should only need to add this to the root .gitignore file as I understand it. The rule will be applied recursively. Is that not working? 

Also could try `**/.factorypath/` in root. 

This way we don't have to maintain, edit, add to, every gitignore file in every module.

symmetric-assemble/common.gradle
```diff
@@ -37,6 +37,7 @@ allprojects {
         maven { url = "https://packages.confluent.io/maven/" }
         maven { url = "https://repository.mulesoft.org/nexus/content/repositories/public/" }
         maven { url = "https://repo1.maven.org/maven2/" }
+        maven { url = "https://maven.jumpmind.com/repo" }
```
w3bbd3v-jm: Was this added by accident? This is the link to the old Jumpmind maven repo.

pavel-jm: Great catch! Removed this antique.

symmetric-client/src/test/java/org/jumpmind/symmetric/job/PurgeMetricStatsJobTest.java
```diff
@@ -0,0 +1,76 @@
...
+        taskScheduler = mock(ThreadPoolTaskScheduler.class);
+        when(engine.getParameterService()).thenReturn(parameterService);
+        when(parameterService.getExternalId()).thenReturn("test-node");
+        when(parameterService.getInt(anyString())).thenReturn(10000);
+    }
+
+    private PurgeMetricStatsJob newJob() {
+        return new PurgeMetricStatsJob(engine, taskScheduler);
+    }
+    // ── getDefaults ───────────────────────────────────────────────────────────
```
w3bbd3v-jm: Remove Claude comments

symmetric-core/src/main/java/org/jumpmind/symmetric/common/ServerConstants.java
```diff
@@ -49,4 +49,5 @@ public class ServerConstants {
     public final static String SERVER_CONNECTION_IDLE_TIMEOUT = "server.connection.idle.timeout";
     public final static String SERVER_SERVLET_CONTEXT_PATH = "symmetric.server.web.home";
     public final static String SERVER_SINGLE_PROPERTIES_FILE = "server.single.properties.file";
+    public static final String SYM_ENV_PREFIX = "SYM_";
```
w3bbd3v-jm, line 52: I feel like this must be declared somewhere else already. Was this addition part of a refactor? Just trying to avoid having duplicate constants that could lead to bugs down the road.

symmetric-client/src/test/java/org/jumpmind/symmetric/job/PurgeMetricStatsJobTest.java
pavel-jm: Yup. Removed

symmetric-core/src/main/java/org/jumpmind/symmetric/common/ServerConstants.java
pavel-jm: Nope!  We have this value in lower case for DB prefixes, but this one is in UPPERCASE for environment variables specifically.

symmetric-util/.gitignore
pavel-jm: Yes. Consolidated in the root .gitignore file

PR #731, opened 2026-04-22
SYM-7534: Added liveliness endpoint API for Kubernetes/Docker integration

evan-miller-jumpmind:
Note that the `IApplicationHealthTracker` is unused for now. Later on it will be used for a `/health/readiness` endpoint in the Pro version. It might also be used to determine whether the `/health/liveliness` endpoint should return some kind of error rather than a 200. This functionality would be part of a separate ticket.

I wrote some very simple documentation for now so let me know if there are any more details that need added.


symmetric-server/src/main/java/org/jumpmind/symmetric/web/LivelinessUriHandler.java
```diff
@@ -0,0 +1,52 @@
...
+import jakarta.servlet.ServletException;
+import jakarta.servlet.http.HttpServletRequest;
+import jakarta.servlet.http.HttpServletResponse;
+
+public class LivelinessUriHandler implements IUriHandler {
+    @Override
+    public void handle(HttpServletRequest req, HttpServletResponse res) throws IOException, ServletException {
+        res.setContentType("application/json");
+        res.getWriter().write("{\"status\": \"UP\"}");
+    }
```
pavel-jm, line 46: Let's use the new IApplicationHealthTracker here - to determine kind of the response to send.
isAlive ==> status=UP
isAlive ==> status=DOWN (and result HTTP code is set to 500 )!

pavel-jm (changes requested):

symmetric-server/src/main/java/org/jumpmind/symmetric/web/LivelinessUriHandler.java
evan-miller-jumpmind: Sure. I'll need to figure out a good place for the `IApplicationHealthTracker` to live first.

pavel-jm (approved):
Thank you!

PR #734, opened 2026-04-23
SYM-7539: auto-install Postgres CITEXT extension on failure.

w3bbd3v-jm:


symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDatabasePlatform.java
```diff
@@ -506,4 +507,29 @@ protected String massageInfinityDate(String date) {
...
+    @Override
+    public void alterTables(boolean continueOnError, boolean createTableIncludeApplicationTriggers, String triggerPrefix,
+            IAlterDatabaseInterceptor[] interceptors, Table... desiredTables) {
+        try {
+            super.alterTables(continueOnError, createTableIncludeApplicationTriggers, triggerPrefix, interceptors, desiredTables);
+        } catch (Exception ex) {
+            if (isMissingCitextExtensionError(ex)) {
+                log.info("Installing citext extension and retrying because a table contains a citext column");
+                getSqlTemplate().update("CREATE EXTENSION IF NOT EXISTS citext");
+                super.alterTables(continueOnError, createTableIncludeApplicationTriggers, triggerPrefix, interceptors, desiredTables);
```
w3bbd3v-jm, line 520: This call can still fail when the db user does not have permission to install the extension. The error message is self-explantory. 
`[target-dataloader-9] Failed to load batch source-16 StackTraceKey.init [SqlException:1940821458] org.jumpmind.db.sql.SqlException: ERROR: permission denied to create extension "citext"
  Hint: Must have CREATE privilege on current database to create this extension.`

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDatabasePlatform.java
```diff
@@ -506,4 +507,29 @@ protected String massageInfinityDate(String date) {
...
+
+    @Override
+    public void alterTables(boolean continueOnError, boolean createTableIncludeApplicationTriggers, String triggerPrefix,
+            IAlterDatabaseInterceptor[] interceptors, Table... desiredTables) {
+        try {
+            super.alterTables(continueOnError, createTableIncludeApplicationTriggers, triggerPrefix, interceptors, desiredTables);
+        } catch (Exception ex) {
+            if (isMissingCitextExtensionError(ex)) {
+                log.info("Installing citext extension and retrying because a table contains a citext column");
+                getSqlTemplate().update("CREATE EXTENSION IF NOT EXISTS citext");
```
w3bbd3v-jm, line 519: This also works for tables/columns outside the public schema, I found in testing.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDatabasePlatform.java
pavel-jm: Based on docs this action requires "superuser" privilege which is less and less likely to be assigned to the SymmetricDS account. In fact if either of the following parameters is set for false, we should not even attempt it!
auto.config.database=false
auto.sync.triggers=false
As we update documentation to mention conditional support for CITEXT column type, let's mention the "superuser" privilege requirement.
https://www.postgresql.org/docs/current/sql-createextension.html

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDatabasePlatform.java
pavel-jm: Cool!!! Looks like the CITEXT code is dormant and just needs activated on PostgreSQL versions 8.4 +
However, we should mention lack of support on versions prior to 8.3 in our docs.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDatabasePlatform.java
pavel-jm: I assume that this code does not get called when user specifically prohibited SymmetricDS account from changing database objects using either of the two parameters?

auto.config.database=false
auto.sync.triggers=false

See also: https://downloads.jumpmind.com/symmetricds/doc/3.17/html/user-guide.html#_preventing_symmetricds_from_running_ddl_statements

pavel-jm (changes requested):
Cool stuff!
Two changes are likely still needed ( I assume a PRO PR is coming soon?):
* PostgreSqlDdlReader class (mapUnknownJdbcTypeForColumn method ) needs to map native type CITEXT to JDBC type VARCHAR: 
     typeName.equalsIgnoreCase("citext") ....
* Documentation update: 
Conditional support for the CITEXT column type ( PostgreSQL versions 8.4 and higher).
Not supported by binary bulk loader? postgres.bulk.copy.binary=true
SymmetricDS will try to install CITEXT extension on the target database that is missing it and table creation requires it.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDatabasePlatform.java
w3bbd3v-jm: Correct -- I could only force it by calling `send-schema`.  create/alter tables is likely already gated by those flags, as it should be.

w3bbd3v-jm:
> Cool stuff! Two changes are likely still needed ( I assume a PRO PR is coming soon?):
> 
> * PostgreSqlDdlReader class (mapUnknownJdbcTypeForColumn method ) needs to map native type CITEXT to JDBC type VARCHAR:
>   typeName.equalsIgnoreCase("citext") ....
> * Documentation update:
>   Conditional support for the CITEXT column type ( PostgreSQL versions 8.4 and higher).
>   Not supported by binary bulk loader? postgres.bulk.copy.binary=true
>   SymmetricDS will try to install CITEXT extension on the target database that is missing it and table creation requires it.

I question whether we need to update the DDL Reader to map the new data type. It's automatically mapped to `LONGVARCHAR`, the same as `text` is handled today. FWIW, `TEXT` is not mapped either.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDatabasePlatform.java
w3bbd3v-jm: > Based on docs this action requires "superuser" privilege which is less and less likely to be assigned to the SymmetricDS account. In fact if either of the following parameters is set for false, we should not even attempt it! auto.config.database=false auto.sync.triggers=false As we update documentation to mention conditional support for CITEXT column type, let's mention the "superuser" privilege requirement. https://www.postgresql.org/docs/current/sql-createextension.html

FWIW, the CITEXT documentation suggests "superuser" isn't necessary since it's a "trusted" extension. 
"This module is considered “trusted”, that is, it can be installed by non-superusers who have CREATE privilege on the current database." 
https://www.postgresql.org/docs/current/citext.html

pavel-jm:
> * CITEXT

It's true that automatic mapping is correct.
Since CITEXT does not support a length modifier, it should be mapped to LONGVARCHAR (and not to a smaller VARCHAR).
Adding an explicit mapping will communicate to future readers of this code, that this mapping was intentional.

Speaking of, 
could you check if this column type is mapped to a CLOB. THis is because the overall size limit is 1 GB!

w3bbd3v-jm:
> > * CITEXT
> 
> It's true that automatic mapping is correct. Since CITEXT does not support a length modifier, it should be mapped to LONGVARCHAR (and not to a smaller VARCHAR). Adding an explicit mapping will communicate to future readers of this code, that this mapping was intentional.
> 
> Speaking of, could you check if this column type is mapped to a CLOB. THis is because the overall size limit is 1 GB!

I added an explicit mapping for it.

pavel-jm (approved):
Thank you!

PR #739, opened 2026-04-24
SYM-7271: Enabled INFO-level com.jumpmind logging

evan-miller-jumpmind:


symmetric-server/src/main/deploy/conf/logback-debug.xml
```diff
@@ -17,6 +17,7 @@
     </appender>
 
     <logger name="org.jumpmind" level="INFO" />
+    <logger name="com.jumpmind" level="INFO" />
```
w3bbd3v-jm, line 20: I assume these changes are for logging in Pro? Can we keep these local to Pro?

evan-miller-jumpmind: That is correct, since the root-level logger has changed from INFO to ERROR, we need to enable INFO logging for the Pro `com.jumpmind` classes.

I can try setting this logging level in Pro Java code instead. The problem is that we don't want to override the user's setting if they have changed the `com.jumpmind` logging level in the XML. I will look into this.

pavel-jm: I was under impression we only have this config file in one place (in this repo)? If so , is it worth time setting up another config - for one-liner?

pavel-jm (approved):

symmetric-server/src/main/deploy/conf/logback-debug.xml
evan-miller-jumpmind: Correct, we only have the XML config files in the open-source project. To move this configuration to Pro, I think we'd have to have Pro code call a (new?) method in `LogSummaryAppenderUtils` to set the logging level for the `com.jumpmind` classes to INFO only if they are not explicitly set in the XML file.

evan-miller-jumpmind: Let me know if this comment can be resolved or if I should work on moving these settings to Pro. It's a little unclear to me because the comment was phrased as a question and Pavel has now approved this PR.

w3bbd3v-jm: Sure, we can resolve it. Sounds like it's the existing architecture. If we wanted to address it, it would be a separate effort.

w3bbd3v-jm (approved):

PR #749, opened 2026-04-27
SYM-7547: Switched UUID to varchar in Postgres

mdrouard-jumpmind:
Modified PostgreSqlDdlReader to mark UUID as varchar to prevent failure when UUID is a PK and Stream lobs is enabled. 
Also caught a bug in MySqlDdlReader.

### DIFF
--- a/symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDdlReader.java
+++ b/symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDdlReader.java
@@ -212,6 +212,8 @@ public class PostgreSqlDdlReader extends AbstractJdbcDdlReader {
             return ColumnTypes.TIMETZ;
         } else if (PostgreSqlDatabasePlatform.isBlobStoredByReference(typeName)) {
             return Types.BLOB;
+        } else if (typeName != null && typeName.equalsIgnoreCase("UUID") && type != null && type == Types.OTHER) {
+            return Types.VARCHAR;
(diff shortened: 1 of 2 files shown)


evan-miller-jumpmind (approved):

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDdlReader.java
```diff
@@ -212,6 +212,8 @@ protected Integer mapUnknownJdbcTypeForColumn(Map<String, Object> values) {
             return ColumnTypes.TIMETZ;
         } else if (PostgreSqlDatabasePlatform.isBlobStoredByReference(typeName)) {
             return Types.BLOB;
+        } else if (typeName != null && typeName.equalsIgnoreCase("UUID") && type != null && type == Types.OTHER) {
```
pavel-jm: Let's reduce redundancy here:
the check
`type != null`
is already included in check:
`type == Types.OTHER`

pavel-jm (changes requested):

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDdlReader.java
mdrouard-jumpmind: Removed `type != null`

w3bbd3v-jm (approved):

pavel-jm (approved):
Thank you!

PR #750, opened 2026-04-27
SYM-7542: Watchdog should not unregister nodes in certain cases

philipmarzullo64:
Most of the changes are because of Sonar.
The important changes are in method fireOffline() and processOfflineNode() and processUnregisterRequest().

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/NodeService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/NodeService.java
@@ -1095,11 +1095,7 @@ public class NodeService extends AbstractService implements INodeService {
         if (myNode != null && !myNode.equals(node)) {
             String myNodeId = myNode.getNodeId();
             if (myNodeId.equals(node.getCreatedAtNodeId())) {
-                if (okToUnregisterNode(node)) {
-                    listener.clientNodeOffline(node);
-                } else {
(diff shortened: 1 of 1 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/NodeService.java
```diff
@@ -1100,27 +1079,67 @@ public List<String> findOfflineNodeIds(long minutesOffline) {
 
     protected void fireOffline(List<Node> offlineClientNodeList) {
         Node myNode = findIdentity();
+        Map<String, Node> offlineNodesNotUnregistered = new HashMap<>();
```
w3bbd3v-jm, line 1082: Maybe dumb question, but is `offlineNodesNotUnregistered` the same as `offlineNodesRegistered`? The latter is a little easier on my brain.

w3bbd3v-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/NodeService.java
philipmarzullo64: This is a data structure that holds the name of the offline node that did not get unregistered because of the inability to recover this node. It is used for logging (for the number of nodes not unregistered for now).

PR #773, opened 2026-05-05
SYM-7325: Fixed push registration not completing for client nodes added via the web console or REST API

evan-miller-jumpmind:


symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
```diff
@@ -290,6 +290,7 @@ public ISymmetricEngine install(Properties passedInProperties, IDatabaseInstallS
                         IRegistrationService registrationService = currentEngine.getRegistrationService();
                         if (!registrationService.isAutoRegistration() && !registrationService.isRegistrationOpen(clientNodeGroupId, externalId)) {
                             Node node = new Node(properties);
+                            node.setSyncUrl(null);
```
pavel-jm: Could you comment why removing (presumably a pre-configured) sync.url value is beneficial, please?

evan-miller-jumpmind: Have you checked the issue description? It has an explanation of the bug: https://jumpmind.atlassian.net/browse/SYM-7325

Note that this is only removing the sync URL from the `Node` object passed to `openRegistration()`. The `sync.url` parameter still gets set in the engine file.

w3bbd3v-jm: Agree -- the bug here feels like it's further up and would require a lot of refactoring. This feels like a "quick fix", which is acceptable IMO. For instance, I would not expect a method named "validateRequiredProperties" to set a default value for the sync url.

evan-miller-jumpmind:
The fix is in `RegistrationService`. There are no functionality changes in `PushService`, just creating a couple of variables and moving a multi-line `if` statement into a separate method, as Pavel suggested. I also changed `CHANNEL_DEFAULT` to `QUEUE_DEFAULT`, which makes no functional difference but clarifies that it's checking the queue name.

evan-miller-jumpmind:
Both Sonar warnings are for methods that I did not change

philipmarzullo64 (approved):

PR #779, opened 2026-05-09
SYM-7593: only send monitor events to registration server.

w3bbd3v-jm:


symmetric-core/src/main/java/org/jumpmind/symmetric/route/RegistrationServerRouter.java
```diff
@@ -0,0 +1,93 @@
...
+                return Collections.singleton(registrationNodeId);
+            }
+        }
+        return NO_NODES_TO_ROUTE;
+    }
+
+    protected boolean hasSomewhereToRoute(Node identity) {
+        return !isRegistrationServer()
+                && identity != null
+                && identity.getCreatedAtNodeId() != null;
```
philipmarzullo64: Need one more check:
&& !identity.getCreatedAtNodeId().equals(identity.getNodeId())

Sometimes the registration server has the createdAtNodeId value set to itself. This is another way to recognize the registration server.

philipmarzullo64 (changes requested):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/TriggerRouterService.java
```diff
@@ -673,6 +674,9 @@ protected TriggerRouter buildTriggerRoutersForSymmetricTables(String version, Tr
         if (TableConstants.getTableName(tablePrefix, TableConstants.SYM_FILE_SNAPSHOT).equals(
                 trigger.getSourceTableName())) {
             router.setRouterType(FileSyncDataRouter.ROUTER_TYPE);
+        } else if (TableConstants.getTableName(tablePrefix, TableConstants.SYM_MONITOR_EVENT).equals(
```
pavel-jm, line 677: While I like this addition, it is a good practice to give customer an option to turn on backwards compatibility (using a parameter)?

Note that we already have a parameter to prevent capture of monitor events (for replication): monitor.events.capture.enabled

So we could either use it or add a new one, like:
monitor.events.filter.enabled or monitor.events.route.upstream.only ?

pavel-jm (changes requested):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/TriggerRouterService.java
w3bbd3v-jm: We discussed adding a parameter as a potential approach but ultimately decided not to offer it. 
One reason is we're approaching this as a performance bug (not a feature). Hence, why we are proposing it in 3.17. We also considered the burden of supporting both approaches going forward. We think ultimately this is a solid approach to the issue and would benefit all customers. Let me know if you want to discuss...

philipmarzullo64 (approved):

pavel-jm (approved):

PR #785, opened 2026-05-12
SYM-7595: More logging and refactoring around Push Registration

pavel-jm:


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RegistrationService.java
```diff
@@ -375,57 +385,73 @@ public RegistrationRequest getLatestRegistrationRequest(String nodeGroupId, Stri
...
+            return updateIncompleteRegistrationRequestInDatabase(request);
         }
+        return recordsChanged;
     }
 
+    @Override
+    public void saveRegistrationRequest(RegistrationRequest request) {
+        String externalId = StringUtils.defaultIfEmpty(request.getExternalId(), "");
+        String nodeGroupId = StringUtils.defaultIfEmpty(request.getNodeGroupId(), "");
+        RegistrationRequest priorRequest = getLatestRegistrationRequest(externalId, nodeGroupId);
```
evan-miller-jumpmind: I checked `getLatestRegistrationRequest()` and I agree with the SonarCloud warning here. The parameters should be ordered as `nodeGroupId, externalId`.

evan-miller-jumpmind (changes requested):
If you want to include the `RegistrationService.saveRegistrationRequest()` refactor in this PR, then I think you should update the Jira ticket to reflect this. Currently, the ticket only mentions adding logging, so this refactor appears to be out of scope.

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RegistrationService.java
pavel-jm: Thanks! fixed

evan-miller-jumpmind (approved):
The style check is failing but otherwise it looks good. It would be good to update the ticket description to include the `RegistrationService.saveRegistrationRequest()` refactor.

pavel-jm:
> If you want to include the `RegistrationService.saveRegistrationRequest()` refactor in this PR, then I think you should update the Jira ticket to reflect this. Currently, the ticket only mentions adding logging, so this refactor appears to be out of scope.

Refactoring was added due to Sonar "brain method" complaint. Which was valid, because it was difficult to follow a huge chunk of code. Updated Jira ticket to add: refactor RegistrationService class to simplify testing and troubleshooting.

evan-miller-jumpmind:
> > If you want to include the `RegistrationService.saveRegistrationRequest()` refactor in this PR, then I think you should update the Jira ticket to reflect this. Currently, the ticket only mentions adding logging, so this refactor appears to be out of scope.
> 
> Refactoring was added due to Sonar "brain method" complaint. Which was valid, because it was difficult to follow a huge chunk of code. Updated Jira ticket to add: refactor RegistrationService class to simplify testing and troubleshooting.

Got it, that makes sense.

symmetric-core/src/main/java/org/jumpmind/symmetric/model/RegistrationRequest.java
```diff
@@ -32,6 +33,9 @@ public static enum RegistrationStatus {
         OK, RQ, RJ, RR, ER
     };
 
+    // This set should be coordinated with the updateRegistrationRequestSql in RegistrationServiceSqlMap:
```
w3bbd3v-jm: Does this mean they should be equal? So if one is changed, we should change the other to match? If so, can we make it so we don't have to do that?

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RegistrationService.java
```diff
@@ -375,57 +385,73 @@ public RegistrationRequest getLatestRegistrationRequest(String nodeGroupId, Stri
...
+                getSql("insertRegistrationRequestSql"),
+                new Object[] { request.getLastUpdateBy(), request.getLastUpdateTime(),
+                        request.getRegisteredNodeId(), request.getStatus().name(), nodeGroupId,
+                        externalId, request.getIpAddress(), request.getHostName(),
+                        request.getErrorMessage(), request.getCreateTime() }, new int[] { Types.VARCHAR, Types.TIMESTAMP,
+                                Types.VARCHAR, Types.VARCHAR, Types.VARCHAR, Types.VARCHAR,
+                                Types.VARCHAR, Types.VARCHAR, Types.VARCHAR, Types.TIMESTAMP });
+    }
+
+    /** Checks for matching host_name and status in ('RQ','ER','RJ') */
```
w3bbd3v-jm: Can we drop this comment or make it so we don't have to update it if we add a new status or something?

symmetric-core/src/main/java/org/jumpmind/symmetric/model/RegistrationRequest.java
pavel-jm: Wise words! Fixed and added test

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RegistrationService.java
pavel-jm: Yup. done

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/AcknowledgeService.java
```diff
@@ -70,6 +70,8 @@ public BatchAckResult ack(final BatchAck batch) {
             if (batch.isOk()) {
                 registrationService.markNodeAsRegistered(batch.getNodeId());
             } else if (batch.getSqlCode() != 0 || batch.getSqlMessage() != null) {
+                log.warn("Registration batch had failed for node {} ! Error line={}, SQL.code={}, SQL message={}",
+                        batch.getNodeId(), batch.getErrorLine(), batch.getSqlCode(), batch.getSqlMessage());
```
w3bbd3v-jm: @pavel-jm there are 6 or so sonar warnings on new code. Are you planning on addressing those? They look relatively small. The only one that seems somewhat concerning is the "logging of user-controlled data" which is flagged here. Is there potential to log customer data here?

pavel-jm: Erm, there is not "materially new" code, because this is a refactor. I've cleaned up a couple, but have my reservations about accuracy of the rest...

pavel-jm: I've addressed the security warning and would invite input on that. The rest are insignificant, while I've already spent a significant amount on this this seemingly simplistic refactor!

symmetric-core/src/test/java/org/jumpmind/symmetric/service/impl/AcknowledgeServiceTest.java
```diff
@@ -0,0 +1,418 @@
...
+        batch.setErrorLine(1);
+        batch.setSqlCode(ErrorConstants.FK_VIOLATION_CODE);
+        OutgoingBatch ob = outgoingBatch(Status.SE);
+        ob.setSqlCode(ErrorConstants.FK_VIOLATION_CODE);
+        ob.setFailedLineNumber(0);
+        List<Number> ids = Arrays.asList(55L);
+        when(outgoingBatchService.findOutgoingBatch(NORMAL_BATCH_ID, NODE_ID)).thenReturn(ob);
+        when(sqlTemplateDirty.query(anyString(), any(NumberMapper.class), anyLong())).thenReturn(ids);
+        when(parameterService.is(ParameterConstants.AUTO_RESOLVE_FOREIGN_KEY_VIOLATION)).thenReturn(true);
+        org.jumpmind.symmetric.service.IDataService dataService = mock(org.jumpmind.symmetric.service.IDataService.class);
```
evan-miller-jumpmind: Can you import `IDataService` instead of including the fully-qualified class name inline?

symmetric-core/src/test/java/org/jumpmind/symmetric/service/impl/AcknowledgeServiceTest.java
evan-miller-jumpmind, line 1: Can you remove the line numbers from the comments in this test class? They will quickly become inaccurate when someone makes changes to `AcknowledgeService`.

symmetric-core/src/test/java/org/jumpmind/symmetric/service/impl/PushServiceTest.java
evan-miller-jumpmind, line 1: This class is missing the license header.

evan-miller-jumpmind, line 1: There are a lot of `"default"` and `"reload"` strings in this class that could be replaced with `Constants.QUEUE_DEFAULT` and `Constants.QUEUE_RELOAD`.

symmetric-core/src/test/java/org/jumpmind/symmetric/service/impl/PushServiceTest.java
```diff
@@ -0,0 +1,374 @@
...
+
+    @Test
+    void testExecuteProcessesAcksWhenReceived() throws Exception {
+        ProcessInfo processInfo = mock(ProcessInfo.class);
+        setupForExecute(processInfo);
+        IOutgoingWithResponseTransport transport = mock(IOutgoingWithResponseTransport.class);
+        when(transport.readResponse()).thenReturn(new BufferedReader(new StringReader("")));
+        when(transportManager.getPushTransport(any(), any(), any(), any(), any())).thenReturn(transport);
+        BatchAck ack = new BatchAck(1L);
+        when(transportManager.readAcknowledgement(any(), any())).thenReturn(new java.util.ArrayList<>(List.of(ack)));
```
evan-miller-jumpmind: Can you import `ArrayList` instead of including the fully-qualified class name inline?

symmetric-core/src/test/java/org/jumpmind/symmetric/util/SecurityUtilsTest.java
evan-miller-jumpmind, line 1: This class is missing the license header.

evan-miller-jumpmind (changes requested):
I found a few issues. Also, please take a look at the 27 SonarCloud warnings on the test classes. All of them seem easy to fix.

symmetric-core/src/test/java/org/jumpmind/symmetric/service/impl/PushServiceTest.java
pavel-jm: Yup! Thanks

pavel-jm: Yup! Thanks

symmetric-core/src/test/java/org/jumpmind/symmetric/service/impl/PushServiceTest.java
pavel-jm: Yup! Thanks

symmetric-core/src/test/java/org/jumpmind/symmetric/util/SecurityUtilsTest.java
pavel-jm: Yup! Thanks

symmetric-core/src/test/java/org/jumpmind/symmetric/service/impl/AcknowledgeServiceTest.java
pavel-jm: Yup! Thanks

evan-miller-jumpmind (approved):

PR #800, opened 2026-05-20
SYM-7228: Finishing cleaning up the file sync canvas

mdrouard-jumpmind:
Finishing clean up of file sync canvas


symmetric-db/src/main/java/org/jumpmind/db/platform/DatabaseNamesConstants.java
```diff
@@ -82,4 +82,10 @@ private DatabaseNamesConstants() {
     public final static String DATABRICKS = "databricks";
     public static final String AWSSQS = "awssqs";
     public static final String FILESYNC = "filesync";
+    public final static String FILESYNC_LOCAL = "local-filesystem";
```
w3bbd3v-jm: Please reorder the modifiers according to Sonar's warnings for your new constants. 
So this line should be: `public static final String FILESYNC_LOCAL = "local-filesystem";`
If we get around to addressing these warnings in the existing code, it will be less we have to change then.

w3bbd3v-jm (approved):

PR #801, opened 2026-05-21
SYM-7621: Support for external data in bulk insert to sym_data.

joshahicks:


symmetric-core/src/main/java/org/jumpmind/symmetric/common/ParameterConstants.java
evan-miller-jumpmind, line 1: Can you move this to `ConsoleConstants` since it's a Pro parameter?

w3bbd3v-jm (approved):

evan-miller-jumpmind (approved):

PR #806, opened 2026-05-26
SYM-7636: Save 5 minute statistics for sym_node_host_channel_stats table even if 0 counts

philipmarzullo64:


symmetric-core/src/main/java/org/jumpmind/symmetric/statistic/StatisticManager.java
```diff
@@ -548,9 +548,7 @@ public void flush() {
                         }
                         stats.setEndTime(endTime);
                         saveAdditionalStats(endTime, stats);
-                        if (stats.isNonZero()) {
```
w3bbd3v-jm, line 551: Can we `@Deprecated(forRemoval = true)` this method since, at least as far as I can tell, this was the only place it was called?

w3bbd3v-jm: @evan-miller-jumpmind based on the git history, you may have some helpful context here. Does dropping this method sense?

evan-miller-jumpmind: It looks like Eric added the `isNonZero()` method and check as a part of [SYM-5321](https://jumpmind.atlassian.net/browse/SYM-5231). The purpose of this check was to reduce the size of the `sym_node_host_channel_stats` export in the support snapshot. Dropping the `isNonZero()` method makes sense to me.

w3bbd3v-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/statistic/StatisticManager.java
philipmarzullo64: I have removed the method from ChannelStats, since there is no caller to this method anymore.

PR #807, opened 2026-05-26
SYM-7599: Replicate data from views

evan-miller-jumpmind:


evan-miller-jumpmind:
I'll take a look at the 50 SonarCloud alerts.

evan-miller-jumpmind:
Apparently GitHub Actions [limits the annotation count to 50](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md#limitations), so there are more SonarCloud alerts than I realized. I'll continue to work on fixing them.

symmetric-core/src/main/java/org/jumpmind/symmetric/io/DbCompare.java
```diff
@@ -652,20 +655,24 @@ protected Table loadTargetTable(DbCompareTables tables, String targetTableName)
...
         } else {
             Map<String, String> tableNameParts = targetEngine.getTargetDialect().getTargetPlatform().parseQualifiedTableName(targetTableName);
             if (tableNameParts.size() == 1) {
-                targetTable = targetEngine.getTargetDialect().getTargetPlatform().getTableFromCache(targetTableName, true);
+                targetTable = (Table) targetEngine.getTargetDialect().getTargetPlatform().getRelationFromCache(targetTableName, true);
             } else {
-                targetTable = targetEngine.getTargetDialect().getTargetPlatform().getTableFromCache(tableNameParts.get("catalog"), tableNameParts.get("schema"),
+                targetTable = (Table) targetEngine.getTargetDialect().getTargetPlatform().getRelationFromCache(tableNameParts.get(TABLE_NAME_PART_CATALOG),
+                        tableNameParts
+                                .get("schema"),
```
pavel-jm: Let's use TABLE_NAME_PART_SCHEMA here. Thank you for refactoring these!!!

symmetric-core/src/main/java/org/jumpmind/symmetric/io/DbCompare.java
```diff
@@ -652,20 +655,24 @@ protected Table loadTargetTable(DbCompareTables tables, String targetTableName)
...
                         tableNameParts
                                 .get("table"), true);
                 if (targetTable == null) {
-                    targetTable = targetEngine.getTargetDialect().getTargetPlatform().getTableFromCache(tableNameParts.get("schema"), tableNameParts.get(
-                            "catalog"),
-                            tableNameParts.get("table"), true);
+                    targetTable = (Table) targetEngine.getTargetDialect().getTargetPlatform().getRelationFromCache(tableNameParts.get(TABLE_NAME_PART_SCHEMA),
+                            tableNameParts
+                                    .get(
+                                            "catalog"),
```
pavel-jm: Let's use TABLE_NAME_PART_CATALOG here.

symmetric-core/src/main/java/org/jumpmind/symmetric/model/DataMetaData.java
```diff
@@ -20,17 +20,17 @@
...
 
 public class DataMetaData {
     private Data data;
-    private Table table;
+    private Relation relation;
     private Router router;
     private NodeChannel nodeChannel;
 
-    public DataMetaData(Data data, Table table, Router router, NodeChannel nodeChannel) {
+    public DataMetaData(Data data, Relation relation, Router router, NodeChannel nodeChannel) {
```
pavel-jm, line 31: I can't help but wonder if there is a more descriptive name for this class?
TabularData
RelationalData
Just a thought...

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
```diff
@@ -3100,23 +3126,20 @@ protected boolean reloadMissingForeignKeyRows(Data data, long batchId, String no
     }
 
     protected void sortTableRowsByForeignKeys(List<TableRow> tableRows) {
-        List<Table> tables = new ArrayList<Table>();
+        List<Table> tables = new ArrayList<>();
```
pavel-jm: We use List<Relation> a lot! An awesome touch would be to use a new class 
   RelationsList
Which not only extends List<Relation>
, but also adds helper methods.
Like this one:

public static RelationsList of(List<TableRow> tableRows){
   RelationsList list = new RelationsList(tableRows.size());
   for (TableRow tableRow : tableRows) {
            tables.add(tableRow.getTable());
   }
   return list;
}

and 

Map<Relation, Integer> getPositionMap(){
        Map<Relation, Integer> indexMap = new HashMap<>();
        int index = 0;
        for (Relation table : this) {
            indexMap.put(table, index++);
        }
     return indexMap;
}

Just a thought...

pavel-jm (approved):
Great work! Both in breadth and depth. Excited for future-proofing replication with more abstract SchemaObjects and TabluarRelations!

Just a couple comments below, which do not affect function.

symmetric-core/src/main/java/org/jumpmind/symmetric/model/DataMetaData.java
evan-miller-jumpmind: I think `RelationalData` would make the most sense. Let me know if you want me to go ahead and make this change or if you want to discuss this more.

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
evan-miller-jumpmind: Sure, I could replace all `List<Relation>` (and probably some `List<Table>`) with `RelationsList`. Let me know if you want me to go ahead and make this change or if you want to discuss this more.

symmetric-core/src/main/java/org/jumpmind/symmetric/model/DataMetaData.java
pavel-jm: Cool! Let's use RelationalData

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataService.java
pavel-jm: I've used this pattern before and it had worked well.
   Class behaves just like a "normal" List would.
One example is: 
symmetric-core/src/main/java/org/jumpmind/symmetric/observability/interfaces/MetricAttributeList.java

symmetric-core/src/main/java/org/jumpmind/symmetric/model/DataMetaData.java
evan-miller-jumpmind: I looked into this and renaming `DataMetaData` to `RelationalData` will break users' custom `IBatchAlgorithm` and `IDataRouter` extensions. Should I follow the same pattern that I followed for interfaces that were affected by the `Table` -> `Relation` change, or is this more trouble than it's worth?

pavel-jm: Agreed, this is not necessary. We would rather maintain backwards compatibility.

evan-miller-jumpmind:
I'll wait to merge this until all of the comments on the Pro PR are resolved.

PR #813, opened 2026-05-28
SYM-7639: Fixed IllegalArgumentException when syncing a Data Refresh job to a pre-3.17.4 node

evan-miller-jumpmind:


evan-miller-jumpmind:
The Sonar warning on `JobManager.getCustomJobDefinitions()` is invalid. The returned list is modified by one of the callers, so fixing this warning would cause an exception to be thrown.

symmetric-client/src/main/java/org/jumpmind/symmetric/job/JobManagerSqlMap.java
```diff
@@ -30,11 +30,11 @@ public JobManagerSqlMap(IDatabasePlatform platform, Map<String, String> replacem
         super(platform, replacementTokens);
         putSql("loadCustomJobs",
                 "select * from $(job) order by job_type, job_name");
-        putSql("insertJobSql", "insert into $(job) (description, job_type, job_expression, "
+        putSql("insertJobSql", "insert into $(job) (description, job_type, job_expression, implementation, "
```
w3bbd3v-jm, line 33: Are we expecting to handle this the same way in 3.18?

evan-miller-jumpmind: Yes, do you have any concerns?

pavel-jm: Yes, the new column implementation will contain either a class name or an interface implementing the Java-type job. This will help us create multiple Data Refresh jobs, while staying compatible with older clients.

pavel-jm (approved):

w3bbd3v-jm (approved):

symmetric-client/src/main/java/org/jumpmind/symmetric/job/JobManagerSqlMap.java
w3bbd3v-jm: > Yes, do you have any concerns?
No, was just curious with the other table refresh work if this implementation would change.

PR #816, opened 2026-05-29
SYM-7633: Don't poll wrapper.log when running in a container mode

Sai-Gumpalli:


symmetric-client/src/main/java/org/jumpmind/symmetric/AbstractCommandLauncher.java
```diff
@@ -353,6 +353,9 @@ protected void addCommonOption(Options options, String opt, boolean hasArg) {
 
     protected void setContainerized(boolean containerized) {
         isContainerEnabled = containerized;
+        if (containerized) {
+            System.setProperty(ServerConstants.CONTAINER_MODE_ENABLED, "true");
+        }
```
evan-miller-jumpmind: Should we set the system property to `false` (or clear it to allow it to fall back to the default value of `false`) if `containerized` is `false`?

symmetric-core/src/main/resources/symmetric-default.properties
```diff
@@ -3285,3 +3285,9 @@ default.values.to.translate=
 # Tags: load
 # Type: boolean
 sync.use.ready.queues=true
+
+# Set to true automatically when SymmetricDS is launched with --container.
+# 
+# Tags: other
+# Type: boolean
+container.mode.enabled=false
```
evan-miller-jumpmind: The Jira ticket description says that the parameter should be called `container.mode.enable`.

evan-miller-jumpmind (approved):

pavel-jm (approved):

PR #822, opened 2026-06-03
SYM-7425: Import JVM and server config from environment variables

Sai-Gumpalli:


symmetric-core/src/main/java/org/jumpmind/symmetric/common/ServerConstants.java
```diff
@@ -51,4 +54,15 @@ public class ServerConstants {
...
     public static final String CONTAINER_MODE_ENABLED = "container.mode.enable";
     public static final String SYM_ENV_PREFIX = "SYM_";
+    public static final Map<String, String> JVM_OVERRIDE_ENV_VARS = Map.ofEntries(entry("SYM_FILE_ENCODING", "file.encoding"),
+            entry("SYM_STAGING_DIR", "java.io.tmpdir"),
+            entry("SYM_WEB_MAX_FORM_SIZE", "org.eclipse.jetty.server.Request.maxFormContentSize"),
+            entry("SYM_WEB_MAX_FORM_KEYS", "org.eclipse.jetty.server.Request.maxFormKeys"),
+            entry("SYM_KEYSTORE_VAULT", "sym.keystore.file"),
+            entry("SYM_KEYSTORE_PASSWORD", "javax.net.ssl.keyStorePassword"),
+            entry("SYM_CERT_TRUST_VAULT", "javax.net.ssl.trustStore"),
+            entry("SYM_CRYPTO_IGNORE_CIPHERS", "symmetric.ssl.ignore.ciphers"),
```
evan-miller-jumpmind: A few of the values in this map already have constants in `SecurityConstants`. Can we use the existing constants for them instead of `String` literals?

symmetric-core/src/main/java/org/jumpmind/symmetric/util/TypedPropertiesFactory.java
evan-miller-jumpmind, line 1: Take a look at the Sonar warnings in this file. The `replace()` method returns the result rather than doing an in-place replacement. I think the second warning is referring to the check for `!addMissingProperties`. I'm not sure if the third warning is valid.

symmetric-core/src/main/java/org/jumpmind/symmetric/common/ServerConstants.java
```diff
@@ -51,4 +57,15 @@ public class ServerConstants {
     public final static String SERVER_SINGLE_PROPERTIES_FILE = "server.single.properties.file";
     public static final String CONTAINER_MODE_ENABLED = "container.mode.enable";
     public static final String SYM_ENV_PREFIX = "SYM_";
+    public static final Map<String, String> JVM_OVERRIDE_ENV_VARS = Map.ofEntries(entry("SYM_FILE_ENCODING", "file.encoding"),
+            entry("SYM_STAGING_DIR", SecurityConstants.SYSPROP_KEYSTORE),
```
evan-miller-jumpmind: This constant is incorrect. I don't think there's a constant for `"java.io.tmpdir"`.

symmetric-core/src/main/java/org/jumpmind/symmetric/common/ServerConstants.java
```diff
@@ -51,4 +57,15 @@ public class ServerConstants {
...
+            entry("SYM_WEB_MAX_FORM_SIZE", "org.eclipse.jetty.server.Request.maxFormContentSize"),
+            entry("SYM_WEB_MAX_FORM_KEYS", "org.eclipse.jetty.server.Request.maxFormKeys"),
+            entry("SYM_KEYSTORE_VAULT", SecurityConstants.SYSPROP_KEYSTORE_TYPE),
```
evan-miller-jumpmind: This constant is also incorrect. It should be `SecurityConstants.SYSPROP_KEYSTORE`.

evan-miller-jumpmind (approved):

w3bbd3v-jm (approved):

PR #831, opened 2026-06-05
SYM-7659: fix stale heartbeat batches query to work across all dbs including mysql and oracle.

w3bbd3v-jm:
Oracle does not like "as" keyword for table aliases.


ShivamJumpmind (approved):
Looks like a small syntax fix. Looks great to me!

PR #844, opened 2026-06-10
SYM-7638: Add tables for built-in metrics to the support snapshot

ShivamJumpmind:


symmetric-client/src/main/java/org/jumpmind/symmetric/util/SnapshotUtil.java
```diff
@@ -136,7 +136,7 @@ public static File createSnapshot(ISymmetricEngine engine, IProgressListener lis
         File tmpDir = new File(parameterService.getTempDirectory(), dirName);
         tmpDir.mkdirs();
         log.info("Creating snapshot file in " + tmpDir.getAbsolutePath());
-        int stepNumber = 0, totalSteps = 36;
+        int stepNumber = 0, totalSteps = 37;
```
w3bbd3v-jm: @ShivamJumpmind you didn't add this, but Sonar is complaining that the variables should be declared on a separate line. In general, I agree with the practice FWIW. Can you declare ` int totalSteps = 37` on a separate line, please?

symmetric-server/src/shortcuts/create-war.launch
```diff
@@ -1,20 +1,26 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <launchConfiguration type="org.eclipse.jdt.launching.localJavaApplication">
-<listAttribute key="org.eclipse.debug.core.MAPPED_RESOURCE_PATHS">
```
w3bbd3v-jm, line 3: Was this change and the rest of the changes in this directory accidentally added?

symmetric-client/src/main/java/org/jumpmind/symmetric/util/SnapshotUtil.java
ShivamJumpmind: Sounds good. I thought the same thing, didn't want to initially change the existing convention, but I totally agree, and will do that.

symmetric-server/src/shortcuts/create-war.launch
ShivamJumpmind: for some reason, a bunch of files in the shortcuts folder were being changed. Not sure exactly why, maybe because of a gradle build or something else. I did actually run tests after the code changes, and it seemed to be fine.

w3bbd3v-jm: Can you revert these changes since they're not needed and out of scope for this change, please?

pavel-jm: If reverting is tricky, consider this workaround:
 Create a new Dev branch and copy 2 Java files to it, then delete this branch

PR #845, opened 2026-06-11
SYM-7191: Prohibit clustered Initial load job when storage is not clustered

Sai-Gumpalli:


evan-miller-jumpmind (approved):
There's a Sonar warning to address and you need to run spotless for the build to succeed. Otherwise, it looks good to me.

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/ParameterService.java
```diff
@@ -51,6 +51,7 @@ public class ParameterService extends AbstractParameterService implements IParam
     private ISqlTemplate sqlTemplate;
     private Date lastUpdateTime;
     private List<DatabaseParameter> offlineParameters;
+    protected volatile boolean initialLoadUseExtractJobOverridden = false;
```
w3bbd3v-jm: Can we use a standard boolean prefix here on the variable name, like "has" or "is"? 
`boolean isInitialLoadUseExtractJobOverridden` reads better. If sonar complains about the name matching an existing method, you could use `has` instead.

w3bbd3v-jm (changes requested):
The `ParameterService` does not seem like the correct place to put this. If you look at the public methods on the service they should all deal with parameters and should not know anything about initial loads or jobs. This makes the class less cohesive. We should reevaluate the placement of this.

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/ParameterService.java
pavel-jm: Yep, this reads well and helps with our journey to Clean Code:
isInitialLoadUseExtractJobOverridden

symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
```diff
@@ -818,79 +818,113 @@ protected void checkForProOnlyDatabase() {
...
-            } catch (Exception e) {
-                if (ExceptionUtils.is(e, UnrecoverableKeyException.class)) {
-                    throw new SymmetricException("Failed to open keystore because keystore password is wrong.  "
-                            + "Check javax.net.ssl.keyStorePassword in conf/sym_service.conf and bin/setenv.", e);
-                }
-                throw e;
+    }
+
+    protected void checkClusteredExtractJobAllowed(Node node) {
+        if (parameterService.isInitialLoadUseExtractJobOverridden()) {
```
w3bbd3v-jm: Could you move the implementation for `isInitialLoadUseExtractJobOverridden` here into this class. It's the only client of the API. It would look better in a protected method though. Something like: 
```
protected void checkClusteredExtractJobAllowed(Node node) {
         if (parameterService.is(ParameterConstants.CLUSTER_LOCKING_ENABLED, false)
                && parameterService.is(ParameterConstants.CLUSTER_STAGING_ENABLED, true)
                && parameterService.is(ParameterConstants.INITIAL_LOAD_USE_EXTRACT_JOB, true)) {
               throw new SymmetricException(...);
         }
}
```

w3bbd3v-jm: Is the parameter override really needed since we throw the exception and fail to start the engine anyway?

Sai-Gumpalli: The problem with moving isInitialLoadUseExtractJobOverridden into checkClusteredExtractJobAllowed is that the parameter override already happens before this method is called. So checkClusteredExtractJobAllowed only has visibility to the parameters that have been "fixed" so it wouldn't be able to tell if they were ever configured incorrectly. 

I am also going to change the exception to an error log, which is what the ticket intended.

pavel-jm: Per our discussion, let's implement a concept of Auditor class(es), which are called to evaluate parameters after a refresh.

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/ClusteredExtractJobParameterAuditor.java
```diff
@@ -0,0 +1,51 @@
...
+        if (parameters.is(ParameterConstants.CLUSTER_LOCKING_ENABLED, false)
+                && parameters.is(ParameterConstants.CLUSTER_STAGING_ENABLED, true)
+                && parameters.is(ParameterConstants.INITIAL_LOAD_USE_EXTRACT_JOB, true)) {
+            parameters.setProperty(ParameterConstants.INITIAL_LOAD_USE_EXTRACT_JOB, "false");
+            this.parametersOverriden = true;
+        }
+    }
+
+    public Optional<String> validate(TypedProperties parameters, ParameterService parameterService) {
+        if (this.parametersOverriden) {
```
pavel-jm: It reads strange:
the method name is "validate", but it does not actually validate anything.
What I mean by this is the values passed in via TypedProperties parameters are not the values which triggered the this.parametersOverriden = true;

pavel-jm (changes requested):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/ClusteredExtractJobParameterAuditor.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+package org.jumpmind.symmetric.service.impl;
+
+import java.util.Optional;
+
+import org.jumpmind.properties.TypedProperties;
+import org.jumpmind.symmetric.common.ParameterConstants;
+
+public class ClusteredExtractJobParameterAuditor implements IParameterAuditor {
+    private boolean parametersOverriden = false;
```
w3bbd3v-jm: Booleans are typically named with an is/has prefix. Let's use that here, making `parametersOverridden`, `isParametersOverridden`.

pavel-jm (approved):

w3bbd3v-jm (approved):

PR #854, opened 2026-06-15
SYM-7634: Prioritize start of registration node when multi-homed with client nodes

ShivamJumpmind:
Summary

When several nodes (engines) are hosted together by the same server, it is beneficial to start the registration node first. 

Context

When client nodes start before registration node, they might throw warnings “unable to register” because the main node is not-yet available.

Acceptance criteria

Prioritizing start of an engine where registration URL is blank.


symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
```diff
@@ -155,7 +156,35 @@ public void start() {
...
+                SymmetricEngineStarter registrationStarter = null;
                 for (SymmetricEngineStarter starter : enginesStarting) {
+                    Properties props = new Properties();
+                    try (InputStream is = new FileInputStream(starter.getPropertiesFile())) {
+                        props.load(is);
+                    } catch (IOException e) {
+                        log.warn("Unable to read properties file to determine registration node: {}", starter.getPropertiesFile());
+                    }
+                    String registrationUrl = props.getProperty(ParameterConstants.REGISTRATION_URL, "");
+                    if (StringUtils.isBlank(registrationUrl)) {
```
evan-miller-jumpmind: There are 2 ways to specify the registration server's `registration.url`:

1. Leave it blank
2. Set it to the same value as the `sync.url`

I'm thinking we should also check for the second case here.

symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
```diff
@@ -155,7 +156,23 @@ public void start() {
...
+                    regExecutor.execute(registrationStarter);
+                    regExecutor.shutdown();
+                    try {
+                        regExecutor.awaitTermination(Long.MAX_VALUE, TimeUnit.DAYS);
+                    } catch (InterruptedException e) {
+                        log.warn("Interrupted while waiting for registration engine to start", e);
+                        Thread.currentThread().interrupt();
+                    }
+                }
+                // Start rest of engines only after the registration engine
```
evan-miller-jumpmind: I think the code is readable enough without this comment, especially because of the log messages that you added.

symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
```diff
@@ -155,7 +156,23 @@ public void start() {
...
                 for (SymmetricEngineStarter starter : enginesStarting) {
+                    log.info("Now starting remaining engines");
```
evan-miller-jumpmind: This log message should be moved outside of the `for` loop so it only gets logged once. It may even be better to move it inside of the `if (registrationStarter != null)` block (at the end) because it doesn't make as much sense if there is no registration server.

symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
```diff
@@ -165,6 +182,23 @@ public void start() {
         }
     }
 
+    public SymmetricEngineStarter findRegistrationStarter(Set<SymmetricEngineStarter> starters) {
+        for (SymmetricEngineStarter starter : starters) {
+            Properties props = new Properties();
+            try (InputStream is = new FileInputStream(starter.getPropertiesFile())) {
+                props.load(is);
+            } catch (IOException e) {
+                log.warn("Unable to read properties file to determine registration node: {}", starter.getPropertiesFile());
```
evan-miller-jumpmind: It looks like when this exception is caught, the engine will end up being detected as the registration server. I think it would be good to add a `continue` statement inside the `catch` block to prevent this.

symmetric-server/src/test/java/org/jumpmind/symmetric/web/SymmetricEngineHolderTest.java
```diff
@@ -0,0 +1,265 @@
...
+    @TempDir
+    File tempDir;
+
+    @Nested
+    class FindRegistrationStarter {
+        private File createPropertiesFile(String name, String registrationUrl, String syncUrl) throws IOException {
+            File file = new File(tempDir, name + ".properties");
+            Properties props = new Properties();
+            if (registrationUrl != null) {
+                props.setProperty("registration.url", registrationUrl);
```
evan-miller-jumpmind: Instead of `String` literals, can you use the constants from `ParameterConstants` for all of the parameter keys in this file?

evan-miller-jumpmind (changes requested):
I left a few comments and you have some Sonar alerts to address. Otherwise, it looks good to me. It looks like you will also need to run `gradlew spotlessApply` to format the code before it can be merged.

symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
evan-miller-jumpmind: Now you have two `if (registrationStarter != null)` statements. Can you remove the second `if` statement and put the log message within the first `if` block, just after the try/catch?

evan-miller-jumpmind (approved):
Take a look at the Sonar comments and see if you can fix them. The `public` modifier warnings are easy to fix. Otherwise, it looks good to me.

symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
```diff
@@ -155,6 +156,23 @@ public void start() {
...
+                            regExecutor.awaitTermination(Long.MAX_VALUE, TimeUnit.DAYS);
+                        } catch (InterruptedException e) {
+                            log.warn("Interrupted while waiting for registration engine to start", e);
+                            Thread.currentThread().interrupt();
+                        }
+                    }
+                }
+                if (registrationStarter != null)
+                    log.info("Now starting remaining engines");
                 for (SymmetricEngineStarter starter : enginesStarting) {
```
pavel-jm: This code is a good clue of what to re-use. 

We now have two Sets of nodes to start:
1. Registration node(s)
2. Non-registration node(s)

Either set could be empty . Either set could have nodes.
The action remains the same - start your engines! :-)

pavel-jm (changes requested):

symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
```diff
@@ -93,78 +94,150 @@ public class SymmetricEngineHolder {
...
-                engines = staticEngines;
-                enginesStarting = staticEnginesStarting;
-                enginesStartingNames = staticEnginesStartingNames;
-                enginesFailed = staticEnginesFailed;
+                switchToStaticEnginesMode();
             }
             if (autoCreate) {
-                log.info("Current directory is {}", System.getProperty("user.dir"));
+                log.info("Reading property files -> load SymmetricEngineStarters into enginesStarting. Current directory is {}", System.getProperty(
+                        "user.dir"));
```
ShivamJumpmind: SonarQube is complaining here because of an "expensive" System.getProperty() call even if logging won't happen. At the moment, I think it is not worth the clutter to address.

symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
```diff
@@ -93,78 +94,150 @@ public class SymmetricEngineHolder {
...
+
+    private ExecutorService regEngineExecutor() {
+        ExecutorService regExecutor = Executors.newSingleThreadExecutor(new CustomizableThreadFactory("symmetric-engine-startup"));
+        return regExecutor;
+    }
+
+    private ExecutorService nonRegEngineExecutor() {
+        int poolSize = Integer.parseInt(System.getProperty(SystemConstants.SYSPROP_CONCURRENT_ENGINES_STARTING_COUNT,
+                DEFAULT_CONCURRENT_ENGINES_STARTING_COUNT));
+        ExecutorService nonRegExecutor = Executors.newFixedThreadPool(poolSize, new CustomizableThreadFactory("symmetric-engine-startup"));
```
ShivamJumpmind: I believe that the 2 lines here made the code more readable and understandable than 1.

symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
```diff
@@ -93,78 +94,150 @@ public class SymmetricEngineHolder {
...
+        }
+        if (StringUtils.isNotBlank(engineFileName)) {
+            enginesStarting.add(new SymmetricEngineStarter(engineFileName, this));
+        } else {
+            log.info("No engine symmetric.properties file found");
+        }
+    }
+
+    private ExecutorService regEngineExecutor() {
+        ExecutorService regExecutor = Executors.newSingleThreadExecutor(new CustomizableThreadFactory("symmetric-engine-startup"));
```
ShivamJumpmind: I believe that the 2 lines here made the code more readable and understandable than 1.

evan-miller-jumpmind (approved):

symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
```diff
@@ -93,78 +94,150 @@ public class SymmetricEngineHolder {
...
-                            }
-                        }
-                        if (!found) {
-                            log.info("No engine *.properties files found");
-                        }
-                    } else {
-                        log.error("Unable to retrieve engine properties files from default location or from current working directory.  No engines to start.");
-                    }
+                    String enginesDirName = PropertiesUtil.getEnginesDir();
+                    loadMultiServerEngines(enginesDirName);
```
pavel-jm: If loadMultiServerEngines were to return a Set , then it would be far more easy to unit-test it (avoiding side-effect functionality):

Example:
enginesStarting.addAll(loadMultiServerEngines())

private Set<SymmetricEngineStarter> loadMultiServerEngines()

symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
```diff
@@ -93,78 +94,150 @@ public class SymmetricEngineHolder {
...
-                        log.info("No engine symmetric.properties file found");
-                    }
+                    String engineFileName = singleServerPropertiesFile;
+                    loadSingleServerEngine(engineFileName);
                 }
-                int poolSize = Integer.parseInt(System.getProperty(SystemConstants.SYSPROP_CONCURRENT_ENGINES_STARTING_COUNT, "5"));
-                ExecutorService executor = Executors.newFixedThreadPool(poolSize, new CustomizableThreadFactory("symmetric-engine-startup"));
-                for (SymmetricEngineStarter starter : enginesStarting) {
-                    executor.execute(starter);
+                log.info("Finding registration engine to start first");
```
pavel-jm: Let's more this entire code block to a new helper method...
Example:
startEnginesInParallelUsingStarters();

symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
```diff
@@ -88,83 +88,153 @@ public class SymmetricEngineHolder {
...
+            executor.execute(starter);
+        }
+        executor.shutdown();
+    }
+
+    protected Set<SymmetricEngineStarter> findRegistrationStarters(Set<SymmetricEngineStarter> starters) {
+        Set<SymmetricEngineStarter> registrationStarters = new HashSet<>();
+        for (SymmetricEngineStarter starter : starters) {
+            Properties props = new Properties();
+            try (InputStream is = new FileInputStream(starter.getPropertiesFile())) {
```
pavel-jm: File manipulation operations are not needed for this method to accomplish it's mission -  all relevant information should already be in the SymmetricEngineStarter object?

pavel-jm (changes requested):

symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
```diff
@@ -88,83 +88,153 @@ public class SymmetricEngineHolder {
...
+            String registrationUrl = props.getProperty(ParameterConstants.REGISTRATION_URL, "");
+            String syncUrl = props.getProperty(ParameterConstants.SYNC_URL, "");
+            if (StringUtils.isBlank(registrationUrl) || registrationUrl.equals(syncUrl)) {
+                registrationStarters.add(starter);
+            }
+        }
+        return registrationStarters;
+    }
+
+    private void startRegistrationEngines(Set<SymmetricEngineStarter> registrationStarters) {
```
pavel-jm: startRegistrationEngines and startNonRegistrationEngines are redundant - they perform the same task, just of different set of data. 
Please look into ways to abstract engine starting process using a Set, which comes in as an input parameter.

symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineStarter.java
```diff
@@ -59,4 +68,17 @@ public String getPropertiesFile() {
     public ISymmetricEngine getEngine() {
         return engine;
     }
+
+    public boolean isRegistrationEngineStarter() {
+        Properties props = new Properties();
+        try (InputStream is = new FileInputStream(getPropertiesFile())) {
+            props.load(is);
+        } catch (IOException e) {
+            log.warn("Unable to read properties file to determine if registration engine starter: {}", getPropertiesFile());
```
ShivamJumpmind, line 77: Engine isn't created until run(), which is after this partition step — so engine is null here and the path String is all we have. Reading the file is the only way to classify it pre-startup; the catch fails safe to non-registration. So I couldn't see a way of implementing the functionality without the file manipulations.

Pavel or Evan, I would love to hear your thoughts and receive your guidance on this matter.

evan-miller-jumpmind: That makes sense, I agree that reading the properties file is necessary here.

symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
```diff
@@ -88,83 +89,139 @@ public class SymmetricEngineHolder {
...
+                .filter(starter -> starter.isRegistrationEngineStarter() == isRegistrationEngineStarter)
+                .collect(Collectors.toSet());
+    }
+
+    private void startEnginesAndWait(Set<SymmetricEngineStarter> starters) {
+        // try-with-resources: close() blocks until these engines finish before returning
+        try (ExecutorService executor = getEngineStarterExecutor()) {
+            executeEngineStarters(executor, starters);
+        }
+    }
```
ShivamJumpmind, line 207: Not duplication — the difference is blocking. startEnginesAndWait uses try-with-resources so close() waits for registration engines to finish; startEngines submits and returns so non-reg engines start in the background. Two named methods instead of a boolean wait flag so the caller reads as intent. Shared plumbing is in executeEngineStarters.

Pavel or Evan, please let me know if you disagree or have any thoughts on this, any feedback is appreciated :)

evan-miller-jumpmind: I agree. After taking a look at how `Executors.newFixedThreadPool()` works, I can see that `startEnginesAndWait()` blocks, while `startEngines()` does not. I like how you changed the method names to make it clear how each one behaves.

symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
```diff
@@ -88,83 +89,139 @@ public class SymmetricEngineHolder {
...
+        return source.stream()
+                .filter(starter -> starter.isRegistrationEngineStarter() == isRegistrationEngineStarter)
+                .collect(Collectors.toSet());
+    }
+
+    private void startEnginesAndWait(Set<SymmetricEngineStarter> starters) {
+        // try-with-resources: close() blocks until these engines finish before returning
+        try (ExecutorService executor = getEngineStarterExecutor()) {
+            executeEngineStarters(executor, starters);
+        }
```
pavel-jm, line 206: Let's add an catch block with log.error  here to record the exception ,if any

symmetric-server/src/main/java/org/jumpmind/symmetric/web/SymmetricEngineHolder.java
```diff
@@ -88,83 +89,139 @@ public class SymmetricEngineHolder {
...
+        Set<SymmetricEngineStarter> registrationEngineStarters = filterEngineStartersByRegistrationType(true, enginesStarting);
+        Set<SymmetricEngineStarter> nonRegistrationEngineStarters = filterEngineStartersByRegistrationType(false, enginesStarting);
+        if (registrationEngineStarters.isEmpty()) {
+            log.debug("No registration engines found. Starting all other engines.");
+        } else {
+            log.debug("Starting registration engines first");
+            startEnginesAndWait(registrationEngineStarters);
+            log.info("All registration engines have been started (before non-registration engines).");
+        }
+        log.info("All engines now starting up.");
```
pavel-jm: Let's convert this to a log.debug level message?

pavel-jm (changes requested):

pavel-jm (approved):

PR #858, opened 2026-06-15
SYM-7685: Fix ReDoS in ServletUtils.java - Replace User-Influenced Regex with Hardcoded Pattern

philipmarzullo64:


symmetric-server/src/main/java/org/jumpmind/symmetric/web/ServletUtils.java
```diff
@@ -44,25 +44,31 @@ public class ServletUtils {
      * @return true if the error could be sent to the response
      * @throws IOException
      */
+    private static String unknown = "unknown";
```
w3bbd3v-jm: The javadoc above it I think is for `sendError` method. Let's move this constant above it.

w3bbd3v-jm: Make constant? 
`private static final String UNKNOWN = "unknown";`

symmetric-server/src/main/java/org/jumpmind/symmetric/web/ServletUtils.java
```diff
@@ -44,25 +44,31 @@ public class ServletUtils {
...
+    }
+
     public static boolean sendError(final HttpServletResponse resp, final int statusCode) throws IOException {
         return sendError(resp, statusCode, null);
     }
 
     public static String whereAreYou(HttpServletRequest request) {
         String ip = request.getHeader("X-Forwarded-For");
-        if (ip == null || ip.length() == 0 || "unknown".equalsIgnoreCase(ip)) {
+        if (ip == null || ip.isEmpty() || unknown.equalsIgnoreCase(ip)) {
```
w3bbd3v-jm: Since we repeat this 5 times, can we put it into a separate method?

symmetric-server/src/main/java/org/jumpmind/symmetric/web/ServletUtils.java
```diff
@@ -197,14 +203,21 @@ public static SymmetricEngineHolder getSymmetricEngineHolder(ServletContext ctx)
                 WebConstants.ATTR_ENGINE_HOLDER);
     }
 
-    public static String getEndpointNameFromUrl(HttpServletRequest request) {
+    public static String getEndpointNameFromUrl(HttpServletRequest request, ServletContext ctx) {
```
w3bbd3v-jm, line 196: Is there danger here in changing this public API contract? It will break existing clients.

symmetric-server/src/main/java/org/jumpmind/symmetric/web/ServletUtils.java
philipmarzullo64: Done

symmetric-server/src/main/java/org/jumpmind/symmetric/web/ServletUtils.java
philipmarzullo64: If we don't change it, the security constraint remains.

w3bbd3v-jm (approved):

PR #862, opened 2026-06-16
SYM-7690: fix path traversal issue.

w3bbd3v-jm:


symmetric-client/src/main/java/org/jumpmind/symmetric/SymmetricAdmin.java
```diff
@@ -686,37 +689,56 @@ private void backup(CommandLine line, List<String> args) throws IOException {
...
-                if (new File(entry.getName()).isAbsolute()) {
-                    fileToOpen = AppUtils.resolveZipEntry(entry);
-                    fileToOpen.getParentFile().mkdirs();
-                } else {
-                    fileToOpen = AppUtils.resolveZipEntry(symHome, entry);
-                }
-                try (FileOutputStream foutput = new FileOutputStream(fileToOpen)) {
+                Path entryPath = resolveFilePath(entry);
+                Files.createDirectories(entryPath.getParent());
+                try (OutputStream fileOutputStream = Files.newOutputStream(entryPath)) {
```
w3bbd3v-jm, line 706: I reported this as a false positive. When I take the same implementation and inline it in the `restore` method, the issue is resolved. However, the clean code approach of removing duplicated code, and using other protected methods to focus and do one thing, confuses sonar. This is a common compliant with sonar's scan results.

philipmarzullo64 (approved):

PR #880, opened 2026-06-23
SYM-7651: Updated S3 dependencies

evan-miller-jumpmind:
I used Claude to analyze the module dependencies and found that many of them are no longer needed after upgrading to version 2 of the AWS SDK. I was unable to test this change because I don't have access to Redshift or Snowflake.


pavel-jm (approved):
Thank you!

PR #893, opened 2026-06-29
SYM-7736: Added support for retained computed columns

evan-miller-jumpmind:


evan-miller-jumpmind:
There's 1 Sonar warning remaining, but it's on a method that I didn't modify.

symmetric-db/src/main/java/org/jumpmind/db/alter/ModelComparator.java
```diff
@@ -244,6 +244,9 @@ public void detectIndexChanges(Database sourceModel, Table sourceTable,
             IIndex targetIndex = targetTable.getIndex(indexIdx);
             IIndex sourceIndex = findCorrespondingIndex(sourceTable, targetIndex);
             if (sourceIndex == null) {
+                if (!platformInfo.isPersistedGeneratedColumnsSupported() && targetTable.doesIndexContainPersistedGeneratedColumn(targetIndex)) {
+                    continue;
```
pavel-jm, line 252: Let's add a log.debug here, please. Just in case user wants to understand why index was dropped.

symmetric-db/src/main/java/org/jumpmind/db/io/DatabaseXmlUtil.java
```diff
@@ -922,6 +924,9 @@ private static void writeColumn(Column column, Writer output) throws IOException
         if (column.isGenerated()) {
             output.write(" generated=\"" + column.isGenerated() + "\"");
         }
+        if (column.isPersisted()) {
+            output.write(" persisted=\"true\"");
+        }
```
pavel-jm, line 936: Just in case user manipulates XML manually, let's add a warning when column in NOT of a generated kind. Persisted and generated should go together.

symmetric-db/src/test/java/org/jumpmind/db/model/TableTest.java
pavel-jm: Could you swap these parameters, while you're in this file, please?

pavel-jm (changes requested):

pavel-jm (approved):

PR #896, opened 2026-06-29
SYM-7744: Add Unit Tests

alexandraboffo8:


symmetric-db/src/test/java/org/jumpmind/db/model/TypeMapTest.java
```diff
@@ -0,0 +1,114 @@
...
+		assertFalse(TypeMap.isNumericType(Types.VARCHAR));
+	}
+
+	@Test
+	void isTextTypeTrueForVarChar() {
+		assertTrue(TypeMap.isTextType(Types.VARCHAR));
+	}
+
+	@Test
+	void isTextTypeFalseForVarChar() {
```
evan-miller-jumpmind: This method should be renamed to `isTextTypeFalseForInteger()`

evan-miller-jumpmind (changes requested):
The tests mostly look good to me. I found one method that has an incorrect name. Here are some other items to address:

1. There's a license header that is required for all open-source classes. You'll need to copy it from any open-source class and paste it at the top of each test file.
2. It looks like there will be some Sonar comments to address under the "Files Changed" section. They're also listed here: https://sonarcloud.io/project/issues?id=jumpmindinc_symmetric-ds&pullRequest=896&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true
3. You'll need to run `gradlew spotlessApply` for the `build-test` check to pass.

symmetric-core/src/test/java/org/jumpmind/symmetric/model/BatchIdTest.java
```diff
@@ -0,0 +1,108 @@
...
+ */
+package org.jumpmind.symmetric.model;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.junit.jupiter.api.Test;
+
+class BatchIdTest {
+    @Test
+    void toStringFormatsAsNodeDashBatch() {
```
w3bbd3v-jm: We typically use a `test` prefix for every test name. Then followed by the method name. Then extra stuff you might want to add to convey the purpose of the test, like you've added, would follow and be separated by an underscore. 

So this would become `testToString_formatsAsNodeDashBatch`. I don't think it's a rule we explicitly follow, but it's a normal Java/JUnit standard that you would find in general. 

I recommend renaming the test names and fixing the sonar warnings. Otherwise looks good to me as well.

w3bbd3v-jm (approved):

evan-miller-jumpmind (approved):

PR #897, opened 2026-06-29
SYM-7740: Add unit tests for utility classes in symmetric-util 

ShivamJumpmind:
As part of the 3.18 testing focus, several pure-logic utility classes in the symmetric-util module have no unit-test coverage. These classes have deterministic, side-effect-free behavior (no database, network, file I/O, or external services), which makes them low-risk, high-value targets for backfilling coverage and exercising the SonarQube quality gate without touching production code.

This ticket covers the six classes: 

VersionUtil (org.jumpmind.util) — static version parsing/comparison; delegates to AbstractVersion. Cover parseVersion (non-numeric stripping, missing major/minor/patch defaulting to 0), isOlderThanVersion (major→minor→patch precedence, equality), isOlderMinorVersion (ignores patch), and blank/development/SNAPSHOT handling. Exercises AbstractVersion's comparison logic as a side effect.

Context (org.jumpmind.util) — Map wrapper. Cover put/get/remove/keySet/getContext, including missing-key behavior.

SortedProperties (org.jumpmind.properties) — verify keys() returns keys in sorted order. (Note: different package from the others, same module.)

RandomTimeSlot (org.jumpmind.util) — getRandomValueSeededByExternalId() is seeded by externalId, so it's deterministic: same externalId/maxValue yields the same value, and the result stays within the expected bound.

LinkedCaseInsensitiveMap (org.jumpmind.util) — case-insensitive key semantics: put/get/containsKey/remove across mixed-case keys resolve to the same entry.

Statistics (org.jumpmind.util) — counters (increment/get/set/contains, increment(category, n)), table stats (incrementTableStats/getTableStats), and timers (startTimer/stopTimer use System.currentTimeMillis(), so assert elapsed ≥ 0 rather than an exact value).

### DIFF
--- a/symmetric-util/src/test/java/org/jumpmind/properties/TypedPropertiesTest.java
+++ b/symmetric-util/src/test/java/org/jumpmind/properties/TypedPropertiesTest.java
@@ -5,7 +5,6 @@ import static org.junit.Assert.assertNull;
 
 import java.util.Properties;
 
-import org.jumpmind.properties.TypedProperties;
 import org.junit.jupiter.api.Test;
 
 class TypedPropertiesTest {
--- a/symmetric-util/src/test/java/org/jumpmind/properties/TypedPropertiesTest.java
+++ b/symmetric-util/src/test/java/org/jumpmind/properties/TypedPropertiesTest.java
@@ -1,4 +1,4 @@
-package org.jumpmind.util.properties;
+package org.jumpmind.properties;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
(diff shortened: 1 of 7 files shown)


symmetric-util/src/test/java/org/jumpmind/properties/TypedPropertiesTest.java
w3bbd3v-jm: Since you fixed the package name by changing the old name, you can remove this import. Sonar is also reporting this.

w3bbd3v-jm (approved):

symmetric-util/src/test/java/org/jumpmind/properties/TypedPropertiesTest.java
pavel-jm, line 8: Agreed - let's remove:
import org.jumpmind.properties.TypedProperties;

pavel-jm (changes requested):

pavel-jm (approved):

PR #898, opened 2026-06-30
SYM-7742: Added variable support to the isBlank, isEmpty, & isNull column transforms

evan-miller-jumpmind:


evan-miller-jumpmind:
I realized I forgot to add unit tests for `TransformVariableUtils`. I'm working on that now.

symmetric-core/src/main/java/org/jumpmind/symmetric/io/data/transform/TransformVariableUtils.java
```diff
@@ -0,0 +1,126 @@
...
+            return context.getData().getAttribute(CsvData.ATTRIBUTE_SOURCE_NODE_ID);
+        } else if (OPTION_OLD_VALUE.equalsIgnoreCase(varName)) {
+            return oldValue;
+        } else if (OPTION_NULL.equals(varName)) {
+            return null;
+        } else if (Strings.CS.equalsAny(varName, OPTION_SOURCE_TABLE_NAME, OPTION_SOURCE_CATALOG_NAME, OPTION_SOURCE_SCHEMA_NAME)) {
+            return resolveSourceNameVariable(varName, context);
+        } else if (OPTION_SOURCE_DML_TYPE.equals(varName)) {
+            return data.getSourceDmlType().toString();
+        } else if (OPTION_BATCH_ID.equals(varName)) {
```
w3bbd3v-jm: Some of the string comparisons use `equalsIgnoreCase` while some don't. Is there some reason for this? Should they all ignore case?

w3bbd3v-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/io/data/transform/TransformVariableUtils.java
evan-miller-jumpmind: I copied these string comparisons over from `VariableColumnTransform`. I don't know why some of them are case-sensitive. I'll update them to ignore case.

PR #901, opened 2026-07-01
SYM-7746: Fixed NPE from null triggerhist pk

mdrouard-jumpmind:


pavel-jm (approved):

symmetric-core/src/test/java/org/jumpmind/symmetric/service/impl/TriggerRouterServiceTest.java
```diff
@@ -579,4 +595,201 @@ public List<TriggerHistory> getActiveTriggerHistories() {
...
+
+    private Trigger newTrigger(String triggerId, String tableName) {
+        Trigger trigger = new Trigger();
+        trigger.setTriggerId(triggerId);
+        trigger.setSourceTableName(tableName);
+        return trigger;
+    }
+
+    @Test
+    void testGetTriggerToTableSupportingInfo_noPrimaryKey_synthesizesPrimaryKeys() throws Exception {
```
pavel-jm: Let's remove Exception declaration from method's signature, as it cannot be thrown from method's body. Repeat for other methods listed by Sonar.

mdrouard-jumpmind: Removed the Throws exception

w3bbd3v-jm (approved):

PR #903, opened 2026-07-02
SYM-7365: Aligned AWS Aurora Postgres endpoint with Pro edition

evan-miller-jumpmind:


symmetric-db/src/main/java/org/jumpmind/db/platform/DatabaseNamesConstants.java
```diff
@@ -50,6 +50,7 @@ private DatabaseNamesConstants() {
     public final static String DB2AS400 = "db2as400";
     public final static String POSTGRESQL = "postgres";
     public final static String POSTGRESQL95 = "postgres95";
+    public static final String POSTGRESQL_AURORA = "postgresaurora";
```
pavel-jm: If it's not too much trouble, let's put Aurora first in these names and labels and Postgres second. 

The AWS Aurora is a distinct type of database engine, which also "happens" to offer some Postgres API compatibility (not complete, hence the dialect).

pavel-jm (approved):

PR #913, opened 2026-07-13
SYM-7228: Moving File Sync Canvas to 3.18

mdrouard-jumpmind:
Moved File Sync Canvas Code From 3.17 to 3.18. 

PRO PR: https://github.com/jumpmindinc/symmetric-pro/pull/933


pavel-jm (approved):
Thank you!

evan-miller-jumpmind (commented):
Should we add a section to the user guide that explains how to configure file sync via the Configure Diagram?

symmetric-sqlexplorer/build.gradle
```diff
@@ -45,6 +45,7 @@ dependencies {
 	
 	implementation "com.vaadin:vaadin-core:$vaadinVersion"
 	implementation "org.glassfish.jaxb:jaxb-runtime:$jaxbVersion"
+	implementation "org.springframework.data:spring-data-commons:4.0.6"
```
w3bbd3v-jm: This should use the same springVersion used by the project. I think there is a `$springVersion` tag? It would be in common.gradle.

w3bbd3v-jm: Maybe not? It looks like it has a different versioning. There is a newer version though you could use. Maybe consider making a new version in common.gradle for it in case we use another other spring data dependencies in the future?

pavel-jm:
> Should we add a section to the user guide that explains how to configure file sync via the Configure Diagram?

Agreed! Let's add couple screenshot somewhere towards the end of the section 3. Set Up New Replication (unless there is a better spot)?

PR #920, opened 2026-07-15
SYM-7705: The first time a node restarts, all triggers are forcefully rebuilt

ShivamJumpmind:
Problem
On a node's first restart after installation, all triggers (system and application tables) are forcefully rebuilt, even though they were already installed correctly during Deployment/setup. This causes unnecessary DDL work.


symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
```diff
@@ -501,6 +501,15 @@ public void setup() {
         }
         isStartupDbParametersDifferentFromLastStart = detectStartupDbParametersDifferentFromLastStart();
         setupDatabase(isStartupDbParametersDifferentFromLastStart);
+        // Shivam's solution (new method): persistStartupDbParametersHashIfMissing();
```
ShivamJumpmind: I have added temporary comments throughout here for explanation and discussion purposes, we will remove them later.

symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
```diff
@@ -1619,15 +1628,39 @@ public ICacheManager getCacheManager() {
...
         boolean dbParamsDifferent = false;
         try {
-            int hashDbParams = parameterService.hashParameterValues(ParameterConstants.STARTUP_DB_OBJECTS_SETUP_PARAMS);
-            String currentHashDbParamsAsString = "0x" + Integer.toHexString(hashDbParams);
+            String currentHashDbParamsAsString = computeCurrentDbParamsHash();
             String priorHashDbParams = contextService.getString(ContextConstants.STARTUP_DB_OBJECTS_SETUP_HASH);
             if (currentHashDbParamsAsString.equals(priorHashDbParams)) {
                 log.debug("No change in SymmetricDS startup database parameters. Hash {} == {}", currentHashDbParamsAsString,
                         priorHashDbParams);
+            } else if (priorHashDbParams == null) {
```
ShivamJumpmind: I thought it might be a good idea to add a specific case for the null scenario here for clarity ...

symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
```diff
@@ -501,6 +501,15 @@ public void setup() {
         }
         isStartupDbParametersDifferentFromLastStart = detectStartupDbParametersDifferentFromLastStart();
         setupDatabase(isStartupDbParametersDifferentFromLastStart);
+        // Shivam's solution (new method): persistStartupDbParametersHashIfMissing();
+        // Pavel's solution (just simply call the existing detectStartupDbParametersDifferentFromLastStart() method again:
```
ShivamJumpmind: Code is currently using Pavel's solution .... this solution is tested and works

evan-miller-jumpmind (commented):
I'm okay with either approach. I think Shivam's approach is a little better because in most cases it doesn't call `computeCurrentDbParamsHash()` a second time.

symmetric-core/src/test/java/org/jumpmind/symmetric/AbstractSymmetricEngineTest.java
```diff
@@ -50,6 +57,55 @@ private void setPlatform(IDatabasePlatform platform) throws Exception {
...
+        assertDoesNotThrow(() -> engine.persistStartupDbParametersHashIfMissing());
+        verify(contextService, never()).save(Mockito.anyString(), Mockito.anyString());
+    }
+
+    @Test
+    public void testComputeCurrentDbParamsHash_returnsHexFormattedString() throws Exception {
+        ParameterService parameterService = mock(ParameterService.class);
+        when(parameterService.hashParameterValues(ParameterConstants.STARTUP_DB_OBJECTS_SETUP_PARAMS)).thenReturn(0xec461721);
+        setField("parameterService", parameterService);
+        assertTrue("0xec461721".equals(engine.computeCurrentDbParamsHash()));
```
pavel-jm: I agree with Sonar, let's use
assertEquals("0xec461721", ... )

symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
pavel-jm: There are two reasons priorHashDbParams==null
1. Table SYM_CONTEXT does not exist
2. Table exists, but has was not yet saved.

So my vote is not to add this IF branch...

symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
pavel-jm: Oh, nice! Thanks for double checking!

evan-miller-jumpmind (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
```diff
@@ -528,6 +528,7 @@ public void setupDatabase(boolean force) {
             } else {
                 log.info("Checking tables and objects. force={}", force);
                 symmetricDialect.initTablesAndDatabaseObjects();
+                detectStartupDbParametersDifferentFromLastStart(); // persist hash now that sym_context exists (SYM-7705)
```
w3bbd3v-jm: If we want to keep the comment, can we at least remove the JIRA? The JIRA is part of the commit message and kept in the commit history already.

symmetric-core/src/test/java/org/jumpmind/symmetric/AbstractSymmetricEngineTest.java
```diff
@@ -50,6 +59,69 @@ private void setPlatform(IDatabasePlatform platform) throws Exception {
...
     }
 
+    private void setField(String fieldName, Object value) throws Exception {
+        Field field = AbstractSymmetricEngine.class.getDeclaredField(fieldName);
+        field.setAccessible(true);
+        field.set(engine, value);
+    }
+
+    @Test
+    public void testDetectStartupDbParametersDifferentFromLastStart_hashMatchesReturnsFalseAndDoesNotSave() throws Exception {
```
w3bbd3v-jm: Sonar (sometimes) complains about `public` access on the unit tests. Can we remove them? No one should be calling these tests from other code outside the class.

evan-miller-jumpmind (approved):
I agree with Andy's comments. Otherwise, it looks good to me.

evan-miller-jumpmind (approved):

w3bbd3v-jm (approved):

pavel-jm (approved):
Thank you!

PR #923, opened 2026-07-15
SYM-7787: Resolve sonar warnings for KafkaWriter and add unit tests

alexandraboffo8:
Sonar flagged S4512 (security hotspot) on the BeanUtils.setProperty calls used to populate a POJO for Confluent/Avro Kafka messages, since it can't statically verify that the property name being set is constrained.

getColumnName() already implicitly guards this — it only ever returns a property name that matches an actual PropertyDescriptor on the target bean, so unrecognized columns are silently skipped before reaching BeanUtils. This change adds an explicit, local allowlist check (derived from the bean's own writable properties) right at each BeanUtils.setProperty call site, so the constraint is visible to the scanner and to future readers, rather than relying on the implicit behavior of a different method. No behavior change is expected, since the existing guard already prevents any non-bean property from reaching this code.

Applied to both writeKafka() and execute(), including the primary-key-only path.

Also added test coverage for the previously-untested Confluent/Avro POJO branches in both methods, and for getColumnName().

### DIFF
--- a/symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/writer/KafkaWriter.java
+++ b/symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/writer/KafkaWriter.java
@@ -768,13 +768,9 @@ public class KafkaWriter extends DynamicDefaultDatabaseWriter {
                                         } catch (Exception e) {
                                             log.debug(rowData[i] + " was not a recognized date format so treating it as a long.");
                                         }
-                                        if (allowedProperties.contains(colName)) {
-                                            BeanUtils.setProperty(pojo, colName, date != null ? date.getTime() : rowData[i]);
-                                        }
+                                        BeanUtils.setProperty(pojo, colName, date != null ? date.getTime() : rowData[i]);
                                     } else {
-                                        if (allowedProperties.contains(colName)) {
-                                            BeanUtils.setProperty(pojo, colName, rowData[i]);
-                                        }
+                                        BeanUtils.setProperty(pojo, colName, rowData[i]);
                                     }
                                 }
                             }
--- a/symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/writer/KafkaWriter.java
+++ b/symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/writer/KafkaWriter.java
@@ -30,8 +30,10 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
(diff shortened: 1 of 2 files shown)


symmetric-io/src/main/java/org/jumpmind/symmetric/io/data/writer/KafkaWriter.java
```diff
@@ -750,9 +768,13 @@ public int writeKafka(CsvData data, Table table) {
                                         } catch (Exception e) {
                                             log.debug(rowData[i] + " was not a recognized date format so treating it as a long.");
                                         }
-                                        BeanUtils.setProperty(pojo, colName, date != null ? date.getTime() : rowData[i]);
+                                        if (allowedProperties.contains(colName)) {
```
alexandraboffo8, line 771: Sonar is showing that there is duplication on new code, but I repeated lines to make it a more explicit check for each BeanUtils.setProperty call.

symmetric-io/src/test/java/org/jumpmind/symmetric/io/data/writer/KafkaWriterTest.java
```diff
@@ -586,4 +840,46 @@ public void setMyField(String myField) {
             this.myField = myField;
         }
     }
+
+    // Bean used to exercise the Confluent/Avro branch of writeKafka where a POJO is
+    // instantiated and populated from row data. Property types cover each population branch:
+    // String (CharSequence branch), Long (date/long branch), and Integer (generic else branch).
+    public static class PojoTestBean {
```
w3bbd3v-jm, line 847: Can we keep access to this local to the test class (i.e. private)? Or at least default access?

symmetric-io/src/test/java/org/jumpmind/symmetric/io/data/writer/KafkaWriterTest.java
```diff
@@ -167,6 +168,34 @@ public void testWriteKafkaJsonFormat() {
         assertTrue(recordValue.contains("\"test_name\""));
     }
 
+    @Test
+    public void testExecute_jsonFormat() {
```
w3bbd3v-jm: I think you're following the existing pattern established by this class by making the tests public. I think we can just drop the access modifier though. Sonar usually complains about this too.

w3bbd3v-jm (approved):

symmetric-io/src/test/java/org/jumpmind/symmetric/io/data/writer/KafkaWriterTest.java
alexandraboffo8: It looks like when I set it either private or default, the tests relying on BeanUtils.setProperty start to fail. I think it is because BeanUtils is trying to access the methods in that class from a different package.

PR #936, opened 2026-07-20
SYM-7702: Prevented a warning message from being logged whenever a node starts up with a platform that lacks a dedicated dialect

evan-miller-jumpmind:
I'm proposing that we remove this log message. Here's my reasoning:

1. There are some platforms that don't need a dedicated dialect, like Kafka and RabbitMQ. This message gets logged for platforms like these (in both open-source and Pro), even though no functionality is missing.
2. We added this message specifically for SQL Server and Oracle. We already have an `AbstractSymmetricEngine.checkForProOnlyDatabase()` method that throws an exception for these databases and prevents the engine from starting up. This exception's message makes the `JdbcSymmetricDialectFactory` warning logging redundant. Also, pointing out that the `GenericSymmetricDialect` has limited functionality doesn't make much sense when the engine isn't going to start anyway.

### DIFF
--- a/symmetric-client/src/main/java/org/jumpmind/symmetric/db/JdbcSymmetricDialectFactory.java
+++ b/symmetric-client/src/main/java/org/jumpmind/symmetric/db/JdbcSymmetricDialectFactory.java
@@ -153,10 +153,6 @@ public class JdbcSymmetricDialectFactory implements ISymmetricDialectFactory {
         } else if (platform instanceof IngresDatabasePlatform) {
             dialect = new IngresSymmetricDialect(parameterService, platform);
         } else {
-            log.warn("No dedicated SymmetricDS dialect found for platform '{}'. "
-                    + "Falling back to GenericSymmetricDialect, which has limited functionality. "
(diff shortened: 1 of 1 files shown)


symmetric-client/src/main/java/org/jumpmind/symmetric/db/JdbcSymmetricDialectFactory.java
```diff
@@ -153,10 +153,6 @@ public ISymmetricDialect create(IParameterService parameterService, IDatabasePla
         } else if (platform instanceof IngresDatabasePlatform) {
             dialect = new IngresSymmetricDialect(parameterService, platform);
         } else {
-            log.warn("No dedicated SymmetricDS dialect found for platform '{}'. "
```
w3bbd3v-jm, line 156: Are we sure we want to remove this? Can we detect non-JDBC platforms (i.e. like Kafka) and just not log it rather than removing it for everyone?

EDIT: I just read your description. That makes sense to me.

w3bbd3v-jm (approved):

PR #938, opened 2026-07-20
SYM-7703: Fixed startup error due to Jackson version mismatch after installing the Kafka module

evan-miller-jumpmind:


pavel-jm (approved):
Nice catch!!!

PR #943, opened 2026-07-23
SYM-7821: Change ClusteredCacheManager to degrade gracefully instead of throwing

alexandraboffo8:
ClusteredCacheManager (new in SYM-7674) hard-throws RuntimeException("Service was not yet initialized!") from broadcastEngineState() and the peer-query methods when !isInitialized(). SYM-7674 wired broadcastEngineState(...) into AbstractSymmetricEngine.setup()/start() — the universal startup path. But ClusteredCacheManager.initialize() is only ever called by SymmetricEngineHolder (the web/server layer). So any engine started without the web server — the service/embedded integration tests via ClientSymmetricEngine.start() — throws at startup, start()'s catch(Throwable) swallows it, and the sym schema is never created → every dependent test fails.   

This causes these integration test classes to fail:   

ProClusterServiceTest, JdbcTriggerRouterServiceTest, JdbcRouterServiceTest, JdbcDataLoaderServiceTest, JdbcDataExtractorServiceTest, DbExportImportTest, JdbcClusterServiceTest

The proposed fix is to have the 6 methods degrade gracefully when !isInitialized() instead of throwing: broadcastEngineState/rebroadcastCurrentState → no-op return; isAnyPeerOnline/isAnyPeerWithEngineInState/isAnyPeerInState → return false; getActiveServerIds → return empty set.

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ClusteredCacheManager.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ClusteredCacheManager.java
@@ -193,7 +193,7 @@ public class ClusteredCacheManager implements IClusteredCacheManager {
     @Override
     public Set<String> getActiveServerIds() {
         if (!isInitialized()) {
-            throw new RuntimeException("Service was not yet initialized!");
+            return new HashSet<>();
         }
         Set<String> active = new HashSet<>();
         if (peerNetworkCoordinator == null) {
@@ -327,7 +327,7 @@ public class ClusteredCacheManager implements IClusteredCacheManager {
     @Override
     public boolean isAnyPeerInState(String eventType) {
         if (!isInitialized()) {
-            throw new RuntimeException("Service was not yet initialized!");
+            return false;
         }
         if (peerNetworkCoordinator == null) {
             return false;
@@ -344,7 +344,7 @@ public class ClusteredCacheManager implements IClusteredCacheManager {
     @Override
     public boolean isAnyPeerOnline() {
         if (!isInitialized()) {
-            throw new RuntimeException("Service was not yet initialized!");
+            return false;
         }
         if (peerNetworkCoordinator == null) {
             return false;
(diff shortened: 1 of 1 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ClusteredCacheManager.java
```diff
@@ -193,7 +193,7 @@ public boolean isPeerOfflineLongEnough(String serverId, long staleThresholdMs) {
     @Override
     public Set<String> getActiveServerIds() {
         if (!isInitialized()) {
-            throw new RuntimeException("Service was not yet initialized!");
+            return new HashSet<>();
```
w3bbd3v-jm, line 196: Would this be better?  `Collections.emptySet();`

NOTE: If any client tries to alter the set it will fail since it's immutable.

Would need some additional testing if we use it...

symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ClusteredCacheManager.java
```diff
@@ -327,7 +327,7 @@ public long getStaleIntervalMs() {
     @Override
     public boolean isAnyPeerInState(String eventType) {
         if (!isInitialized()) {
-            throw new RuntimeException("Service was not yet initialized!");
+            return false;
```
w3bbd3v-jm, line 330: For any of these method that were throwing an exception, should we log something out instead, then return?

symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ClusteredCacheManager.java
alexandraboffo8: That does make sense to me, the reasoning for returning an empty hash set instead is because below this 'if' statement, there is another 'if' statement that returns an empty set, and it uses hash set. Although it is formatted a bit differently as it returns the empty hash set, set as a variable.

symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ClusteredCacheManager.java
alexandraboffo8: That sounds like a good idea, I could do log.debug with a more specific message to each method?

symmetric-core/src/main/java/org/jumpmind/symmetric/cache/ClusteredCacheManager.java
pavel-jm, line 196: It is my understanding that the engine holder is used in all production code?
But not in test code?

Should we consider adding EngineHolde .init() in test code as well?

In theory, such initialization would fix tests and avoid weaker guard statements in Production code (if possible)

pavel-jm (changes requested):

PR #947, opened 2026-07-27
SYM-7816: Added support for Google Cloud SQL for PostgreSQL endpoints (Pro edition)

evan-miller-jumpmind:


symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlVariantDetector.java
```diff
@@ -41,4 +41,13 @@ public static boolean isAuroraPostgres(Connection connection) {
             return false;
         }
     }
+
+    public static boolean isCloudSqlPostgres(Connection connection) {
+        try (Statement stmt = connection.createStatement();
+                ResultSet rs = stmt.executeQuery("show cloudsql.iam_authentication")) {
```
w3bbd3v-jm, line 47: We can drop `ResultSet rs = ` since it's not used?

evan-miller-jumpmind: If I remove that, it doesn't compile. I think it's necessary for the try-with-resources syntax.

symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
```diff
@@ -889,6 +889,14 @@ protected void checkForProOnlyDatabase() {
                             + "Contact the SymmetricDS sales team for more information.",
                     dbVersionName);
         }
+        if (DatabaseNamesConstants.CLOUDSQL_POSTGRESQL.equalsIgnoreCase(dbVersionName)
+                && !DatabaseNamesConstants.CLOUDSQL_POSTGRESQL.equalsIgnoreCase(platform.getName())) {
```
w3bbd3v-jm, line 893: I'm curious about this part of the condition: `!DatabaseNamesConstants.CLOUDSQL_POSTGRESQL.equalsIgnoreCase(platform.getName()`. 

Can the dbVersion and the platform name both match the constant and be GC SQL for PostgresSQL?

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlVariantDetector.java
w3bbd3v-jm: Okay, for some reason that does sound familiar.

symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
evan-miller-jumpmind: Yes, they will both match the constant when running the Pro version. Only the Pro version has a platform that is specific to Google Cloud SQL for PostgreSQL. The open-source version will use `PostgreSql95DatabasePlatform` which gets its name from `DatabaseNamesConstants.POSTGRESQL95`.

w3bbd3v-jm (approved):

PR #953, opened 2026-07-28
SYM-7824: Instance ownership collision - HOSTNAME replacement

ShivamJumpmind:
**Description**

Problem
The host_name recorded in sym_node_host is the engine's server id, which by default resolves at each startup as: [cluster.server.id](http://cluster.server.id/) parameter → $(HOSTNAME) 

The HOSTNAME template is not replaced. 

**Impact**
The sym_node_host runtime table is keyed on (node_id, host_name), and the update statement matches on those columns. When the resolved hostname changes, the engine can't match its existing row, so it inserts a new one and strands the old row. The stranded row stops heart-beating and still carries the installation's instance id, which interacts badly with the exclusive-ownership check.


ShivamJumpmind:
I believe that to solve this hostname indeterminism issue, more changes will need to be made. The change this PR contains is for sure a step in the right direction ... I am wondering if this PR gets merged, if the JIRA ticket for it will become void.

pavel-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/util/TypedPropertiesFactory.java
```diff
@@ -143,6 +143,9 @@ public static final void replaceSystemAndEnvironmentVariables(Properties propert
                 if (value.contains("hostName")) {
                     value = FormatUtils.replace("hostName", AppUtils.getHostName(), value);
                 }
+                if (value.contains("HOSTNAME")) {
```
w3bbd3v-jm: Can we consolidate this if statement with the one above it? It's almost exactly the same.

pavel-jm: I thought about it, but the text in the string is different (case sensitive).
Note: this is the crooks of the issue - if the text replacement was case insensitive this PR would not exist.

w3bbd3v-jm: Couple approaches: 
Is `FormatUtils.replace("hostName", AppUtils.getHostName(), value); ` a no-op when the key is not found? 
If that's the case, then you can remove the if check and end up with something like this:

`private static Map<String, String> hostNameReplacements() {
    String hostName = AppUtils.getHostName();
    Map<String, String> replacements = new HashMap<>(2);
    replacements.put("hostName", hostName);
    replacements.put("HOSTNAME", hostName);
    return replacements;
}`

If you want to keep the if check anyways, you could do something like this:
`private static final List<String> HOSTNAME_TOKENS = List.of("hostName", "HOSTNAME");`
`for (String token : HOSTNAME_TOKENS) {
    if (value.contains(token)) {
        value = FormatUtils.replace(token, AppUtils.getHostName(), value);
    }
}`

w3bbd3v-jm:
> I believe that to solve this hostname indeterminism issue, more changes will need to be made. The change this PR contains is for sure a step in the right direction ... I am wondering if this PR gets merged, if the JIRA ticket for it will become void.

Agree that this seems like a band-aid and does not actually fix the bug. Do we know what it would take to fix the bug or the level of effort? I'd rather we address the real issue if possible. 

If this is the approach we want for now, then I'd vote that we create a new bug ticket to tackle the root of the issue. And make it clear that this does not fully address the bug. 

Just my 2 cents...

symmetric-core/src/main/java/org/jumpmind/symmetric/util/TypedPropertiesFactory.java
pavel-jm: @ShivamJumpmind , Could you adopt this cool suggestion from Andy, please?

pavel-jm (approved):

pavel-jm:
> > I believe that to solve this hostname indeterminism issue, more changes will need to be made. The change this PR contains is for sure a step in the right direction ... I am wondering if this PR gets merged, if the JIRA ticket for it will become void.
> 
> Agree that this seems like a band-aid and does not actually fix the bug. Do we know what it would take to fix the bug or the level of effort? I'd rather we address the real issue if possible.
> 
> If this is the approach we want for now, then I'd vote that we create a new bug ticket to tackle the root of the issue. And make it clear that this does not fully address the bug.
> 
> Just my 2 cents...

We've got the root cause fixed now - thanks to your inquiry!

PR #955, opened 2026-07-28
SYM-7833: Updated log class name for all descendants of AbstractDatabasePlatform

evan-miller-jumpmind:
I could have avoided adding `getClassName()` to every open-source platform by having `AbstractDatabasePlatform` implement it and call `getClass().getName()`. However, I think it's better to force each concrete platform to have its own implementation of `getClassName()` so that when someone adds a new Pro platform, they don't forget to implement this method and handle obfuscation. Let me know if this needs changed.

### DIFF
--- a/symmetric-db/build.gradle
+++ b/symmetric-db/build.gradle
@@ -19,4 +19,10 @@ apply from: symAssembleDir + '/common.gradle'
         }
 
         testImplementation project(path: ':symmetric-util', configuration: 'testArtifacts')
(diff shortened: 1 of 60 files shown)


w3bbd3v-jm (approved):

evan-miller-jumpmind:
Merging this PR will break the build on the `release/3.18` branch, so I'll wait to merge it until the Pro PR is approved. That way they can both be merged at around the same time.

pavel-jm (approved):

PR #957, opened 2026-07-29
SYM-7840: Changed variable type from 'integer' to 'numeric(18,0)' in Interbase trigger template

riandg:
Changed variable type from 'integer' to 'numeric(18,0)' in InterbaseTriggerTemplate.java to support larger generator values and prevent arithmetic overflow exceptions during trigger execution.


philipmarzullo64:
Thanks for your contribution, we used another pull request to make the code change, but used your code as the solution.

PR #962, opened 2026-07-30
fix: guard sync-triggers reconciler against transiently empty trigger config

abell-ops:
See also Jira cards: 
https://jumpmind.atlassian.net/browse/SYM-7844
https://jumpmind.atlassian.net/browse/JMCH-2156

Adds a defensive guard so the sync-triggers reconciler cannot drop every capture trigger when the trigger configuration transiently reads back empty.

## The defect

`TriggerRouterService.syncTriggers(...)` reads the current node's trigger configuration into `triggersForCurrentNode`, then calls `inactivateTriggers(...)`, which looks up each active `TriggerHistory` by trigger id in that list. If the lookup misses (`trigger == null`), it logs "About to remove triggers for inactivated table" and drops the trigger.

This is the *intended* behavior when `createTriggersForTables` is `false` (initial load in progress — `triggersForCurrentNode.clear()` is called on purpose in that case). But there is no guard for the *unexpected* case: if `triggersForCurrentNode` comes back empty while trigger creation is otherwise enabled — e.g. because a non-atomic configuration reseed replicated a config delete before its corresponding re-insert was applied — every lookup misses and **every** active capture trigger gets inactivated/dropped fleet-wide, with nothing to force a rebuild until the config is complete again. This traces back to a downstream production outage (TPS ticket JMCH-2156) triggered by exactly this kind of transient config state.

## The fix

- Extracted the decision into a new protected, unit-testable helper, `shouldSkipInactivationForEmptyConfig(List<Trigger> triggersForCurrentNode, List<TriggerHistory> activeTriggerHistories, boolean createTriggersForTables)`, in `TriggerRouterService`.
- In `syncTriggers(...)`, just before the `inactivateTriggers(...)` call, if the helper returns `true`, the inactivation pass is skipped for this run and a `log.warn(...)` explains why (config read back empty while N active histories exist; likely transient/in-flight; will reconcile on the next sync-triggers run).
- The helper only returns `true` when `createTriggersForTables` is `true` — so it never touches the intentional initial-load clear path (`createTriggersForTables == false`), where dropping is still expected.
- Gated by a new parameter, `sync.triggers.skip.inactivation.when.config.empty` (`ParameterConstants.SYNC_TRIGGERS_SKIP_INACTIVATION_WHEN_CONFIG_EMPTY`), defaulted to `true` (safe/guard-on) in `symmetric-default.properties`, `DatabaseOverridable: true` so it can be tuned/disabled per node if ever needed.
- `updateOrCreateDatabaseTriggers(...)` is unaffected (already a no-op on empty config) and continues to run every time.

## Files changed

- `symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/TriggerRouterService.java` — added `shouldSkipInactivationForEmptyConfig(...)`; wired the skip check into `syncTriggers(StringBuilder, boolean)` immediately before the existing `inactivateTriggers(...)` call.
- `symmetric-core/src/main/java/org/jumpmind/symmetric/common/ParameterConstants.java` — added `SYNC_TRIGGERS_SKIP_INACTIVATION_WHEN_CONFIG_EMPTY`.
- `symmetric-core/src/main/resources/symmetric-default.properties` — documented default (`true`) for the new parameter, following the existing `sync.triggers.*` doc-comment convention.
- `symmetric-core/src/test/java/org/jumpmind/symmetric/service/impl/TriggerRouterServiceTest.java` — added unit tests for `shouldSkipInactivationForEmptyConfig(...)`.

## Unit test

Five new tests exercising `shouldSkipInactivationForEmptyConfig(...)` directly (no database):
- empty config + non-empty active histories + trigger creation enabled → `true` (the guarded/positive case)
- non-empty config → `false`
- empty active histories → `false`
- `createTriggersForTables == false` (the intentional initial-load clear) → `false`, and the parameter is never even consulted in that path
- new parameter explicitly disabled → `false`

Ran narrowly: `./gradlew :symmetric-core:test --tests '*TriggerRouterServiceTest*'` (from `symmetric-assemble`, the Gradle root, with JDK 17 per `common.gradle`'s `sourceCompatibility`). Result: **29 tests, 0 failures, 0 errors** (24 pre-existing + 5 new) — `BUILD SUCCESSFUL`.

## Known limitations / notes

- This guard prevents the destructive drop, but it does not itself force a follow-up rebuild — it relies on the *next* `syncTriggers` run reading a complete configuration and reconciling normally. If the config never becomes complete again, capture would remain stale (though no worse than before this change, and vastly safer than dropping everything).
- Complementary config-side fixes (e.g. making the config reseed atomic, or ordering delete/re-insert replication) are out of scope here and tracked separately downstream.
- **This PR is a draft.** It is unit-tested only (`symmetric-core` module, no database) — it has not been run against a live SymmetricDS engine or exercised via integration tests, per the scope of this change.

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/common/ParameterConstants.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/common/ParameterConstants.java
@@ -377,6 +377,7 @@ final public class ParameterConstants {
     public static final String SYNC_TRIGGERS_TIMEOUT_IN_SECONDS = "sync.triggers.timeout.in.seconds";
     public static final String SYNC_TRIGGERS_REG_SVR_INSTALL_WITHOUT_CONFIG = "sync.triggers.reg.svr.install.without.config";
     public static final String SYNC_TRIGGERS_FIX_DUPLICATE_ACTIVE_TRIGGER_HISTORIES = "sync.triggers.fix.duplicate.active.trigger.histories";
+    public static final String SYNC_TRIGGERS_SKIP_INACTIVATION_WHEN_CONFIG_EMPTY = "sync.triggers.skip.inactivation.when.config.empty";
     public static final String MONITOR_EVENTS_CAPTURE_ENABLED = "monitor.events.capture.enabled";
     public static final String HYBRID_PUSH_PULL_ENABLED = "hybrid.push.pull.enabled";
     public static final String HYBRID_PUSH_PULL_TIMEOUT = "hybrid.push.pull.timeout.ms";
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/TriggerRouterService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/TriggerRouterService.java
@@ -1669,7 +1669,16 @@ public class TriggerRouterService extends AbstractService implements ITriggerRou
                         ts = System.currentTimeMillis();
                         List<TriggerHistory> activeTriggerHistories = getActiveTriggerHistories();
                         context.incrementActiveTriggerHistoriesTime(System.currentTimeMillis() - ts);
-                        inactivateTriggers(triggersForCurrentNode, sqlBuffer, activeTriggerHistories, context);
+                        if (shouldSkipInactivationForEmptyConfig(triggersForCurrentNode, activeTriggerHistories, createTriggersForTables)) {
+                            log.warn("Trigger configuration for this node read back empty while {} active trigger "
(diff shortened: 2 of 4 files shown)


abell-ops:
Closing — per the JMCH-2156 review (Aug 4, 2026), SymmetricDS is working-as-designed: the config batch is expected to be applied as a single atomic transaction, and the fix belongs on the JMC/Commerce loader side. Leaving this empty-config guard idea available if the SymmetricDS team revisits self-healing (≈ RCA item P6). Ref JMCH-2156.

PR #964, opened 2026-07-31
SYM-7748: Log4j2 to Logback Migration

ShivamJumpmind:
### **This is one of the fixes in the Log4j2 to Logback migration.**

This PR is for making sure the open-source repo also aligns with Logback, logging files are added to the snapshot, and that there is a warn logging message should the migration fail.


symmetric-util/src/main/java/org/jumpmind/util/LogbackHelper.java
```diff
@@ -74,6 +74,8 @@ public void initialize(boolean isDebug) {
             }
             context.putProperty("HOSTNAME", AppUtils.getHostName());
             enforceProtectedLoggers();
+        } else {
+            log.warn("No conf/logback.xml file exists! Possible failures: Symmetric Installation OR Log4j2_Logback migration.");
```
evan-miller-jumpmind: On line 64 above, the name of the file could be `logback-debug.xml` instead of `logback.xml`. This log message should take that into account.

evan-miller-jumpmind (approved):

pavel-jm (approved):

PR #965, opened 2026-08-01
SYM-7848: Database Health Tracker helps jobs and connection manager perform a preflight check of DB connection

pavel-jm:
See card for details: https://jumpmind.atlassian.net/browse/SYM-7848


symmetric-server/src/main/java/org/jumpmind/symmetric/web/NodeConcurrencyInterceptor.java
```diff
@@ -143,6 +143,9 @@ protected void sendError(HttpServletResponse resp, ReservationStatus status, Str
         if (status == ReservationStatus.DUPLICATE) {
             log.debug("Node {} is already connected", nodeId);
             ServletUtils.sendError(resp, WebConstants.SC_ALREADY_CONNECTED);
+        } else if (status == ReservationStatus.NOT_READY) {
```
ShivamJumpmind, line 146: WebConstants.getHttpMessage() has a case for every other status this method sends (SC_ALREADY_CONNECTED, SC_NO_RESERVATION,SC_SERVICE_BUSY), but not for SC_SERVICE_NOT_READY. It's consumed in AbstractService.java:391, which only appends httpMessage= when non-null — so the remote node logs a bare httpCode=671 with no reason. Could we add else if (httpCode == SC_SERVICE_NOT_READY) { httpMessage = "Service is not ready"; } to that chain?

symmetric-client/src/main/java/org/jumpmind/symmetric/job/AbstractJob.java
```diff
@@ -269,6 +269,10 @@ protected boolean checkPrerequsites(boolean force) {
             log.info("The engine is not currently started, will not run job '{}'", getName());
             return false;
         }
+        if (!force && !engine.isRuntimeDbHealthy()) {
```
ShivamJumpmind: This is the only prerequisite that does real I/O, but it's placed ahead of the cheap in-memory guards (running.get(), paused.get(), the registration check, the node-group check). A paused job — or one scoped to a different node group — will now open a database connection every cycle just to be told it shouldn't run. Moving this block down to just before the final return true gives identical behavior for jobs that actually run and skips the connection test for jobs that were never going to.

symmetric-core/src/main/resources/symmetric-default.properties
```diff
@@ -415,6 +415,30 @@ sync.table.prefix=sym
...
+# Enables a pre-flight health check of the runtime database connection before each job run
+# and before accepting an incoming push or pull.  The check only opens and closes a connection,
+# so a failure always means the connection could not be established.
+#
+# Tags: database
+# Type: boolean
+db.health.check.enabled=true
+
+# Number of consecutive connection test failures before the runtime database is declared unhealthy.
+# While unhealthy, jobs are skipped and incoming syncs are rejected as busy for a wait period of
```
ShivamJumpmind: the code returns ReservationStatus.NOT_READY → SC_SERVICE_NOT_READY, not busy. whats-new.ad correctly says "service-not-ready". Should this say "rejected as not ready"?

symmetric-core/src/main/resources/symmetric-default.properties
```diff
@@ -415,6 +415,30 @@ sync.table.prefix=sym
...
+# Tags: database
+# Type: integer
+db.health.check.failure.threshold=5
+
+# Number of seconds used to calculate the wait period between connection tests while the runtime
+# database is unhealthy.  The wait period is db.health.check.failure.threshold multiplied by this value.
+#
+# Tags: database
+# Type: integer
+db.health.check.timeout.seconds=60
```
ShivamJumpmind: Do we think this could be renamed to something like db.health.check.retry.interval.seconds? I think a name like this might be better since nothing is actually "timing out"?

symmetric-core/src/main/java/org/jumpmind/symmetric/util/DatabaseHealthTracker.java
```diff
@@ -0,0 +1,168 @@
...
+        long waitMs = getUnhealthyWaitMs();
+        unhealthyUntilMs = clock.getAsLong() + waitMs;
+        updateEngineReadiness(false);
+        log.warn("Runtime database is unhealthy after {} consecutive connection failures, pausing jobs and incoming syncs for {} ms: {}",
+                failures, waitMs, lastResult.result());
+    }
+
+    private boolean testConnection() {
+        try {
+            sqlTemplateSupplier.get().testConnection();
```
ShivamJumpmind: is there a case for a DB that hangs? if a DB outright refuses a connection, that seems to be handled well ... but what if a DB hangs or takes long to accept a connection call it sees? is there a timeout set up for that?

symmetric-core/src/main/java/org/jumpmind/symmetric/model/DbHealthCheckResult.java
```diff
@@ -0,0 +1,29 @@
...
+ * under the License.
+ */
+package org.jumpmind.symmetric.model;
+
+import java.util.Date;
+
+/**
+ * Outcome of a single runtime database connection test: when it ran, whether a connection could be established, and either "OK" or the failure message.
+ */
+public record DbHealthCheckResult(Date recorded, boolean isHealthy, String result) {
```
ShivamJumpmind: Cosmetic - do we think it might be a good idea to just change the result field ... so we don't run into seeing _result.result()_?

ShivamJumpmind (commented):
With the help of Claude, I tried my best to leave some feedback. I am new to this topic and these classes however, so please take my comments with a grain of salt. I am happy to discuss my comments, or talk more about this improvement.

symmetric-core/src/main/java/org/jumpmind/symmetric/model/DbHealthCheckResult.java
pavel-jm: Sure,  Updated variable name

symmetric-core/src/main/resources/symmetric-default.properties
pavel-jm: Sure, removing dashes makes it more readable. Thanks!

symmetric-core/src/main/resources/symmetric-default.properties
pavel-jm: Thanks! Retry is a simpler, cleaner moniker.

symmetric-core/src/main/java/org/jumpmind/symmetric/util/DatabaseHealthTracker.java
pavel-jm: Good catch! Added check (timeout) = db.pool.max.wait.millis

evan-miller-jumpmind:
Is this feature meant to be open-source or Pro only? The Jira ticket has the Pro component, but all of the changes are open-source.

symmetric-server/src/main/java/org/jumpmind/symmetric/web/NodeConcurrencyInterceptor.java
pavel-jm: Thanks! Added

symmetric-client/src/main/java/org/jumpmind/symmetric/job/AbstractJob.java
pavel-jm: Agreed to the first two points (and moved down). But there is still a risk that method isRegisteredWithServer will have a cache-miss scenario and call out to (unavailable) database.

symmetric-core/src/main/java/org/jumpmind/symmetric/common/ParameterConstants.java
```diff
@@ -484,7 +487,8 @@ private ParameterConstants() {
             LOAD_ONLY_PROPERTY_PREFIX + DB_POOL_URL };
     public static final String[] ALL_JDBC_PARAMS = new String[] { DB_FETCH_SIZE, DB_QUERY_TIMEOUT_SECS, JDBC_EXECUTE_BATCH_SIZE, JDBC_ISOLATION_LEVEL,
             JDBC_READ_STRINGS_AS_BYTES, TREAT_BINARY_AS_LOB_ENABLED, LOG_SLOW_SQL_THRESHOLD_MILLIS, LOG_SQL_PARAMETERS_INLINE,
-            INCLUDE_ROWIDENTIFIER_AS_COLUMN };
+            INCLUDE_ROWIDENTIFIER_AS_COLUMN, "mssql.use.ntypes.for.sync"
+    };
```
evan-miller-jumpmind, line 489: What is the purpose of this change? I don't see how it's related to the Database Health Tracker.

pavel-jm: Yep! This was pulled in from a bug fix where database setup is missed on start up.

evan-miller-jumpmind: Ok. It seems to me like it would fit better under a separate ticket and PR, but I'll leave that decision up to you.

evan-miller-jumpmind (approved):
Don't forget to remove the Pro component from the Jira ticket. Otherwise, it looks good to me.

philipmarzullo64 (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/common/ParameterConstants.java
pavel-jm: Agreed. New PR is https://github.com/jumpmindinc/symmetric-ds/pull/967

pavel-jm:
> Don't forget to remove the Pro component from the Jira ticket. Otherwise, it looks good to me.

Done. Thank you!

PR #969, opened 2026-08-03
SYM-7867: Upgrade DB2 Test Implementation driver version for integration tests

alexandraboffo8:
The DB2 test driver org.jumpmind.symmetric.module:db2jcc:9.7 predates JDBC 4 and doesn't implement Connection.isValid(int). DBCP2 calls it to validate every pooled connection, so the pool fails with AbstractMethodError before any SQL runs and the engine can't start, no DB2 integration test is possible. 
 
Fix is a one-line swap in symmetric-ds/symmetric-jdbc/build.gradle to com.ibm.db2:jcc:11.5.9.0; the driver class name is unchanged, so no db-test.properties driver edit is needed.


w3bbd3v-jm (approved):

symmetric-jdbc/build.gradle
```diff
@@ -24,7 +24,7 @@ apply from: symAssembleDir + '/common.gradle'
         testImplementation "org.apache.derby:derbytools:$derbyVersion"
         testImplementation "org.hsqldb:hsqldb:$hsqldbVersion"
         testImplementation "org.xerial:sqlite-jdbc:$sqliteVersion"
-        testImplementation "org.jumpmind.symmetric.module:db2jcc:9.7"
+        testImplementation "com.ibm.db2:jcc:11.5.9.0"
```
w3bbd3v-jm: Any chance the latest version `12.1.5.0` works?

alexandraboffo8: Yes, I tested it and I can switch it!

PR #973, opened 2026-08-04
SYM-2983: Resume transfer of large batches using HTTP Range headers

evan-miller-jumpmind:


evan-miller-jumpmind:
I'm taking a look at the Sonar alerts.

w3bbd3v-jm (changes requested):
Are we sure we want to include this in 3.18? The scope for 3.18 is already rather large and we still have a lot of testing to go. Including another new feature seems to bloat the release even more than it is. Maybe it's time we open release/3.19?

evan-miller-jumpmind:
> Are we sure we want to include this in 3.18? The scope for 3.18 is already rather large and we still have a lot of testing to go. Including another new feature seems to bloat the release even more than it is. Maybe it's time we open release/3.19?

We're leaning towards including it in 3.18, but we still need to make a final decision. I created a lab environment in symmetric-labs to make it easier to test this change.

evan-miller-jumpmind:
I fixed most of the Sonar alerts and now only 8 remain.

The Sonar warning in `FileSyncService.handleExtractionError()` is inaccurate. The `currentBatch` variable can't use `log.error()`'s built-in formatting or else `e` will be treated as a variable to be substituted into the string rather than an exception, which will cause an exception to be thrown.

The 3 Sonar failures related to closing streams are false positives. The `HttpOutgoingTransport` and `HttpIncomingTransport` own these streams and handle closing them.

The Sonar failure on `FileSyncService.processUnzippedBatches()` is on pre-existing logic that I extracted into a renamed method. It would be high-risk to resolve it, so I left it as it is.

Three of the alerts in `StagedResource` are on code that I didn't touch at all. The root cause of the failure in `getWriter()` is actually in `buildFile()`, which is untouched.

pavel-jm:
> new feature seems to bloat

We have reviewed this feature in the team meeting. 
   Let me know if you'd like us to schedule another meeting on this topic?

symmetric-core/src/main/java/org/jumpmind/symmetric/io/stage/SimpleStagingDataWriter.java
```diff
@@ -367,6 +385,88 @@ protected IStagedResource getStagedResource() {
...
+     *         resumed retry: resume is enabled, the batch's staged content is file-backed, and its ETag was captured before the interruption
+     */
+    protected boolean isResumableInterruption(Exception ex, IStagedResource resource) {
+        return ex instanceof IOException && batch != null && currentBatchEtag != null && resource.isFileResource()
+                && engine.getParameterService().is(ParameterConstants.TRANSPORT_HTTP_RESUME_ENABLED)
+                && getResumeCache() != null;
+    }
+
+    protected void registerForResume(IStagedResource resource) {
+        resource.close();
```
pavel-jm, line 442: Claude's concern 1
Client and server measure the offset against different streams. 

The client's Range offset is its partial staged file size
  (SimpleStagingDataWriter.registerForResume → resource.getSize()). But the client never stages the stats_columns, stats, and  etag lines it consumed from the wire (process() captures them without writeLine). 

The server, meanwhile, applies skipCount to the composed wire stream — extractSingleBatchForResume wraps the destination in CountingSkippingWriter and then
  transferFromStaging writes the injected stats/ETAG preamble extras through that counting writer. So the server's skip coordinate includes ~100–200 characters the client's coordinate excludes. 

Every honored resume re-sends that many characters  the client already staged, starting mid-line: the client appends a partial-line fragment plus duplicated rows to its staged
  file → corrupted batch (load error at best; silently duplicated rows if the cut lands on a line boundary). The file-sync path  does not have this problem — it skips raw bytes of the staged zip with no injected content — which highlights the fix: on a  resume response, apply skipCount to the staged file content only and suppress the stats/ETAG injection, so the ETag-validated
  representation and the ranged representation are the same thing. 

(This is also the RFC-level principle being violated: a 206 range must be computed against the exact representation the validator covers, and the injected per-request stats lines make
  the wire representation non-invariant between attempts.)
?

symmetric-core/src/main/java/org/jumpmind/symmetric/extract/CountingSkippingWriter.java
```diff
@@ -0,0 +1,75 @@
...
+ * this class must agree on that same unit on both the client and server side. Since both sides read the exact same staged, UTF-8 file deterministically, this
+ * is internally consistent even though it is not a literal byte offset per RFC 9110 Range semantics.
+ */
+public class CountingSkippingWriter extends Writer {
+    private final Writer delegate;
+    private final long skipCount;
+    private long totalCount;
+
+    public CountingSkippingWriter(Writer delegate, long skipCount) {
+        this.delegate = delegate;
```
pavel-jm, line 42: Claude's concern 2:
Bytes vs. characters. The client sends bytes= + UTF-8 byte length of its staged file (File.length() via
  StagedResource.getSize()); the server skips decoded characters. 

CountingSkippingWriter's javadoc claims both sides "agree on  that same unit" — they might not always agree. 
Any multi-byte character in the already-received portion makes the server skip more characters  than the client staged, producing a gap (lost data) on resume. 

Improve: 
Count the same unit on both sides (e.g., client  tracks characters written to staging, or server counts encoded UTF-8 bytes).

symmetric-core/src/main/java/org/jumpmind/symmetric/io/stage/SimpleStagingDataWriter.java
```diff
@@ -95,6 +100,9 @@ public void process() throws IOException {
             String batchStatsColumnsLine = null;
             String batchStatsLine = null;
             Statistics batchStats = null;
+            if (resumeEntry != null) {
+                resource = beginResumedBatch();
+            }
```
pavel-jm, line 106: Claude's concern 3:
Stale resume entry is never cleared when the local partial is gone. 

If beginResumedBatch() returns null (partial  staged resource missing or finalized), the entire 206 body is discarded as invalid lines, batch stays null, so the  commit-branch clearResumeCacheEntry never runs — and DataLoaderService removes the entry only when the server did not honor  the resume. 

The server's state is intact, so it will honor the same resume on every subsequent pull: that queue loops (forever) serving one discarded batch and never pulls anything else. 

Improvement:
beginResumedBatch should clear the cache entry when it  bails out.

symmetric-core/src/test/java/org/jumpmind/symmetric/extract/CountingSkippingWriterTest.java
```diff
@@ -0,0 +1,103 @@
...
+ * <http://www.gnu.org/licenses/>.
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.jumpmind.symmetric.extract;
```
pavel-jm, line 21: Having a new unit test is great, but we likely need a new one:

1. Extract a stream, 
2. truncate at an arbitrary offset (ideally after a multi-byte character), 
3. run the client writer, 
4. resume with the registered entry against a real   extractSingleBatchForResume, 
5. and assert the reassembled staged file is identical to an uninterrupted transfer. 

The existing tests each mock the opposite side, so the coordinate mismatch is invisible to them.

symmetric-server/src/main/java/org/jumpmind/symmetric/web/FileSyncPullUriHandler.java
```diff
@@ -50,25 +51,41 @@ public void handle(HttpServletRequest req, HttpServletResponse res) throws IOExc
...
                     "Node must be specified");
             return;
-        } else {
-            log.debug("File sync pull request received from {}", nodeId);
         }
+        String batchIdParam = ServletUtils.getParameter(req, WebConstants.BATCH_ID);
+        String ifETagHeader = req.getHeader(WebConstants.HEADER_IF_ETAG);
+        String rangeHeader = req.getHeader(WebConstants.HEADER_RANGE);
+        log.debug("File sync pull request received from {}: batchId={}, {}={}, {}={}", nodeId, batchIdParam,
+                WebConstants.HEADER_IF_ETAG, ifETagHeader, WebConstants.HEADER_RANGE, rangeHeader);
```
pavel-jm, line 62: Double-check if this class in missing the Content-Range header (to set it)?

RFC 9110 §14.4 requires bytes 200-499/500.
 The client only checks header presence, also network intermediaries (reverse proxies, WAFs) may validate or mangle malformed 206 responses.

symmetric-server/src/main/java/org/jumpmind/symmetric/web/FileSyncPullUriHandler.java
```diff
@@ -50,25 +51,41 @@ public void handle(HttpServletRequest req, HttpServletResponse res) throws IOExc
             ServletUtils.sendError(res, WebConstants.SC_BAD_REQUEST,
                     "Node must be specified");
             return;
-        } else {
-            log.debug("File sync pull request received from {}", nodeId);
         }
+        String batchIdParam = ServletUtils.getParameter(req, WebConstants.BATCH_ID);
+        String ifETagHeader = req.getHeader(WebConstants.HEADER_IF_ETAG);
```
pavel-jm, line 59: ETAG value should be enclosed in double-quotes.
Probably quoted + base64 encoded, too?

pavel-jm (changes requested):
Thanks!   I'll share additional details over chat

w3bbd3v-jm:
> > new feature seems to bloat
> 
> We have reviewed this feature in the team meeting. Let me know if you'd like us to schedule another meeting on this topic?

My vote is for 3.19.

evan-miller-jumpmind:
I'm taking a look at the 7 new Sonar alerts.

pavel-jm:
It appears there are two SonarQube findings worth a follow up. 
   Both are for the same resource-leak prevention practice:
  > Use try-with-resources or close this "OutputStream" in a "finally" clause.

evan-miller-jumpmind:
> It appears there are two SonarQube findings worth a follow up. Both are for the same resource-leak prevention practice:
> 
> > Use try-with-resources or close this "OutputStream" in a "finally" clause.

When I looked at this before, I found that the 3 Sonar failures related to closing streams are false positives. The `HttpOutgoingTransport` and `HttpIncomingTransport` own these streams and handle closing them.

evan-miller-jumpmind:
We decided to move this feature to 3.19. I'll mark this PR as a draft for now and change the target branch to `release/3.19` once it's available.

PR #977, opened 2026-08-05
SYM-7829: Add an Insight to reduce max batch to send on multi-homed nodes

Sai-Gumpalli:


symmetric-core/src/main/java/org/jumpmind/symmetric/common/MonitorConstants.java
```diff
@@ -91,6 +92,7 @@ public static Map<String, String> getMonitorTypesByVersion() {
         for (String name : new String[] { DB_PERFORMANCE, CONNECTION_POOL_METRICS }) {
             map.put(name, "3.18.0");
         }
+        map.put(MULTI_HOMED_MAX_BATCH_TO_SEND, "3.18.0");
```
evan-miller-jumpmind: Can you add the new insight name to the `String[]` on line 92 instead?

pavel-jm (approved):

symmetric-assemble/src/asciidoc/configuration/monitors.ad
```diff
@@ -79,6 +79,8 @@ be compared to a threshold value.
 
 |maxBatchToSend|Smallest Max Batch to Send for a channel.|✔
 
+|multiHomedMaxBatchToSend|Number of channels with a Max Batch to Send greater than the configured threshold on a multi-homed node.|✔
```
evan-miller-jumpmind: This description will need updated to something like "Largest Max Batch to Send for a channel used only for multi-homed nodes."

I think the `maxBatchToSend` description on line 80 above is inaccurate and should say "Largest" instead of "Smallest", so you could update that as well.

evan-miller-jumpmind (approved):

PR #978, opened 2026-08-05
SYM-7733: Fixed db.client.connection.count metric not getting updated

evan-miller-jumpmind:
The problem was that the Connection Pool Metrics monitor calls `IEngineMetricsService.getLongGauge()`, but the `db.client.connection.count` metric was defined with a double gauge.


pavel-jm (approved):
Nice find! Thanks

PR #982, opened 2026-08-07
SYM-7880: Faster implementation of getFullyQualifiedName

evan-miller-jumpmind:
This implementation is 3-10x faster when `quoteString` is empty, which is the case for sync triggers, log miner, and compare & repair.


pavel-jm (approved):
Wow!  3-10x faster is a very nice acceleration, especially for single-file Log Miner jobs!

PR #983, opened 2026-08-07
SYM-7885: Change column names to not match Sybase ASE reserved words

alexandraboffo8:
There is a matching symmetric-pro PR for this fix.


pavel-jm (approved):
Thank you!

PR #987, opened 2026-08-11
SYM-7883: Prevent clearing cluster-peer locks when local engine is shutting down

pavel-jm:
Prevent clearing peer locks when local engine is shutting down.
Detect own shutdown and ignore peer messages


evan-miller-jumpmind (commented):
There are 5 Sonar warnings that would be very easy to fix.

Were you able to run the 3i02_jcs Docker lab test 1 and confirm that this fixes the symptoms from the ticket description?

evan-miller-jumpmind (approved):

PR #988, opened 2026-08-11
SYM-7891: Fix slow Routing job when node cache misses, nodes list is empty

pavel-jm:
Routing job runs table scan on sym_node for each data record, when there are no nodes in list (node has syn_enabled=false, node is in another tier, node no longer exists)

Also:
* NodeCache.getNodesByGroup() reloads all enabled nodes on any miss, and flushNodeGroupCache() invalidated EVERY group rather than the one that missed.

* A node that is enabled and in the target group but excluded by a channel ignore or grouplet would refresh/run query per row being routed.

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RouterService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RouterService.java
@@ -781,6 +781,40 @@ public class RouterService extends AbstractService implements IRouterService, IN
         return nodes;
     }
 
+    protected Collection<String> findNodeIdsFromNodeList(Data data, TriggerRouter triggerRouter, ChannelRouterContext context) {
+        List<String> targetNodeIds = Arrays.asList(data.getNodeList().split(","));
+        Collection<String> nodeIds = CollectionUtils.intersection(targetNodeIds, toNodeIds(findAvailableNodes(triggerRouter, context)));
+        Collection<String> missingNodeIds = CollectionUtils.subtract(targetNodeIds, nodeIds);
+        if (isNodeCacheRefreshNeeded(missingNodeIds, triggerRouter)) {
+            engine.getNodeService().flushNodeGroupCache();
+            context.getAvailableNodes().remove(triggerRouter);
+            nodeIds = CollectionUtils.intersection(targetNodeIds, toNodeIds(findAvailableNodes(triggerRouter, context)));
+        }
+        if (nodeIds.isEmpty() && log.isDebugEnabled()) {
+            log.debug(
+                    "None of the target nodes specified in the data.node_list field ({}) were qualified nodes. Data id {} for table '{}' will not be routed using the {} router",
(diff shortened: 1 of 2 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RouterService.java
```diff
@@ -781,6 +781,40 @@ protected Set<Node> findAvailableNodes(TriggerRouter triggerRouter, ChannelRoute
         return nodes;
     }
 
+    protected Collection<String> findNodeIdsFromNodeList(Data data, TriggerRouter triggerRouter, ChannelRouterContext context) {
+        List<String> targetNodeIds = Arrays.asList(data.getNodeList().split(","));
```
pavel-jm, line 785: Refactor: moved this existing  code from createReader

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RouterService.java
```diff
@@ -781,6 +781,40 @@ protected Set<Node> findAvailableNodes(TriggerRouter triggerRouter, ChannelRoute
...
+        return nodeIds;
+    }
+
+    /**
+     * Tells whether refreshing the node cache can resolve any of the missing nodes, which is only true for a node the cache has not seen yet as an enabled
+     * member of the router's target node group. Any other missing node stays unresolved after a refresh, so refreshing would re-read every enabled node on
+     * EVERY data row (expensive!) for nothing.
+     */
+    protected boolean isNodeCacheRefreshNeeded(Collection<String> missingNodeIds, TriggerRouter triggerRouter) {
+        String targetNodeGroupId = triggerRouter.getRouter().getNodeGroupLink().getTargetNodeGroupId();
```
pavel-jm: New code - consolidate decision about "missing" nodes to avoid per-data-row cache-miss:
* Newly registered
* sync_enabled=false
* Belongs to lower (in-store) tier

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RouterService.java
```diff
@@ -976,22 +1010,8 @@ protected int routeData(ProcessInfo processInfo, Data data, ChannelRouterContext
...
-                            engine.getNodeService().flushNodeGroupCache();
-                            context.getAvailableNodes().remove(triggerRouter);
-                            nodeIds = CollectionUtils.intersection(targetNodeIdsList, toNodeIds(findAvailableNodes(triggerRouter, context)));
-                        }
-                        if (nodeIds.size() == 0 && log.isDebugEnabled()) {
-                            log.debug(
-                                    "None of the target nodes specified in the data.node_list field ({}) were qualified nodes. Data id {} for table '{}' will not be routed using the {} router",
-                                    new Object[] { targetNodeIds, data.getDataId(), data.getTableName(), router.getRouterId() });
-                        }
+                    if (StringUtils.isNotBlank(data.getNodeList())) {
```
pavel-jm, line 1023: Code above just moved to findNodeIdsFromNodeList

philipmarzullo64 (approved):

PR #994, opened 2026-08-13
SYM-7915: Prevent initial loads deadlocks due to a JVM restart mid-extract

gp510:
Fixes [SYM-7915](https://jumpmind.atlassian.net/browse/SYM-7915).

## The problem

An initial load whose extract is interrupted by a JVM restart comes back permanently stuck. `sym_extract_request` reads `OK` with `extracted_rows=0` while its batches are still `RQ`. Every extraction selector filters on `NE` so nothing re-extracts it, `extractBatch` refuses to deliver an `RQ` batch, nothing reconciles the two, and nothing surfaces it.

The only symptom in the product is a pair of INFO log lines. The console shows an ACTIVE pipeline making no progress, indistinguishable from a slow load. At the reporting site it ran three days, was never recovered, and the load had to be discarded.

## Root cause

`extractOutgoingBatch` marked the **entire** request — all of its batches — `OK` from a single batch's `extract_row_count` and `extract_millis`, with no check that anything was extracted and no check that the rest of the range had finished. A batch that skipped extraction because it was already extracted contributes a zero row count and a near-zero duration, which is exactly the row observed.

## Three layers

- **Guard** — `isExtractRequestComplete` gates the status write. When incomplete, statistics accumulate through a new incremental statement instead, so progress stays visible without claiming completion.
- **Reconciler** — `recoverStuckExtractRequests` returns such requests to `NE`, called from `queueWork`, which `InitialLoadExtractorJob` already drives. That covers startup for free. A startup-only check would not have helped the reporting site, whose node ran three days in this state.
- **`restartExtractRequest`** now also zeroes `extracted_rows` / `extracted_millis`, which it previously left behind.

## Why the detection has no false positives

`MultiBatchStagingWriter.close()` advances every remaining batch in the range, so a finished extract leaves none at `RQ`. **A request at `OK` with an `RQ` batch in its range is an impossible state, not a healthy in-flight one.** That is what makes both the guard and the reconciler safe against a running extract.

## The constraint that shapes the recovery

`restartExtractRequest` flips the whole range back to `RQ` through `updateOutgoingBatchStatusSql`, which carries **no status predicate** — so it re-sends batches that already landed. Requests whose range contains delivered batches are therefore **not** restarted automatically; they are reported at ERROR every pass with the recovery choices and require `force`. The product cannot know whether the target tolerates a re-apply, and the field evidence (500,000 rows committed with no `sym_incoming_batch` row) shows the target's own record can be untrustworthy.

## Review notes

- No schema change. Everything needed is derivable from `sym_outgoing_batch`.
- `DataExtractorService` is the only implementer of the interface and `FileSyncExtractorService` inherits it; there is no PRO implementer, so the interface addition breaks nothing.
- `FileSyncExtractorService` returns 0 from the recovery, matching how `updateExtractRequestsForThreading` is already handled there — otherwise it would scan and restart ordinary data extract requests and resolve their staging through the file-sync override.
- The requested/delivered lookups deliberately select one row rather than `count(*)`: the answers are booleans, and a count visits the whole batch range, so across a load split into N batches the cost would grow with N².
- The recovery scan is throttled to its own staleness threshold. `queueWork` runs every 10s but a request cannot qualify for 5 minutes, so 29 of every 30 scans could not have a new answer.

**781 tests pass in `:symmetric-core`, 0 failures.**

## Not covered here

The end-to-end kill-and-recover path needs a real database — it is on the ticket as manual QA.

**Found while reading and deliberately not folded in:** `sym_extract_request` rows are inserted as `LS` and flipped `LS→NE` by `DataService`, and nothing recovers an `LS` request either, so a kill during load setup strands a load the same way. The reconciler is the natural home, but it is a separate defect and should be its own ticket.

[SYM-7915]: https://jumpmind.atlassian.net/browse/SYM-7915?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/common/ParameterConstants.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/common/ParameterConstants.java
@@ -159,6 +159,8 @@ final public class ParameterConstants {
     public static final String INITIAL_LOAD_CONCAT_CSV_IN_SQL_ENABLED = "initial.load.concat.csv.in.sql.enabled";
     public static final String INITIAL_LOAD_USE_COLUMN_TEMPLATES_ENABLED = "initial.load.use.column.templates.enabled";
     public static final String INITIAL_LOAD_EXTRACT_THREAD_COUNT_PER_SERVER = "initial.load.extract.thread.per.server.count";
+    public static final String INITIAL_LOAD_EXTRACT_REQUEST_RECOVERY_ENABLED = "initial.load.extract.request.recovery.enabled";
+    public static final String INITIAL_LOAD_EXTRACT_REQUEST_RECOVERY_THRESHOLD_MS = "initial.load.extract.request.recovery.threshold.ms";
     public static final String INITIAL_LOAD_EXTRACT_MAX_PROCESS_TIME_MS = "initial.load.extract.max.process.time.ms";
     public static final String INITIAL_LOAD_EXTRACT_TIMEOUT_MS = "initial.load.extract.timeout.ms";
     public static final String INITIAL_LOAD_EXTRACT_USE_TWO_PASS_LOB = "initial.load.extract.use.two.pass.lob";
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/common/ParameterConstants.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/common/ParameterConstants.java
@@ -159,8 +159,6 @@ final public class ParameterConstants {
     public static final String INITIAL_LOAD_CONCAT_CSV_IN_SQL_ENABLED = "initial.load.concat.csv.in.sql.enabled";
     public static final String INITIAL_LOAD_USE_COLUMN_TEMPLATES_ENABLED = "initial.load.use.column.templates.enabled";
     public static final String INITIAL_LOAD_EXTRACT_THREAD_COUNT_PER_SERVER = "initial.load.extract.thread.per.server.count";
-    public static final String INITIAL_LOAD_EXTRACT_REQUEST_RECOVERY_ENABLED = "initial.load.extract.request.recovery.enabled";
-    public static final String INITIAL_LOAD_EXTRACT_REQUEST_RECOVERY_THRESHOLD_MS = "initial.load.extract.request.recovery.threshold.ms";
     public static final String INITIAL_LOAD_EXTRACT_MAX_PROCESS_TIME_MS = "initial.load.extract.max.process.time.ms";
     public static final String INITIAL_LOAD_EXTRACT_TIMEOUT_MS = "initial.load.extract.timeout.ms";
     public static final String INITIAL_LOAD_EXTRACT_USE_TWO_PASS_LOB = "initial.load.extract.use.two.pass.lob";
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
@@ -1733,6 +1733,9 @@ public class DataExtractorService extends AbstractService implements IDataExtrac
         return transformExtractWriter;
     }
 
+    /** When recovery last scanned for stuck extract requests, so it is not re-run faster than a request can become stale. */
+    protected long lastStuckExtractRequestCheckMs;
+
     @Override
     public RemoteNodeStatuses queueWork(boolean force) {
         final RemoteNodeStatuses statuses = new RemoteNodeStatuses(configurationService.getChannels(false));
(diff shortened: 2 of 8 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
```diff
@@ -2087,6 +2106,106 @@ public void execute(NodeCommunication nodeCommunication, RemoteNodeStatus status
...
+     * after a completed extract-job run none of them are left at {@code RQ}. A request at {@code OK} with any batch in its range still {@code RQ} is therefore
+     * an impossible state rather than a healthy in-flight one, which is what makes both this guard and {@link #recoverStuckExtractRequests(boolean)} safe from
+     * false positives. The zero-row check is a heuristic and is called out as such where it appears.
+     */
+    protected boolean isExtractRequestComplete(ExtractRequest request, OutgoingBatch currentBatch, ExtractMode mode) {
+        if (mode == ExtractMode.EXTRACT_ONLY) {
+            // changeBatchStatus does not persist in this mode, so the batch's own row would still read RQ.
+            return false;
+        }
+        if (request.getEndBatchId() > request.getStartBatchId() && currentBatch.getExtractRowCount() == 0 && request.getRows() > 0) {
```
pavel-jm, line 2122: Unfortunately this condition is hard to read... Making it hard to support.
There are plenty of variables inplay:
* Clustered servers overtaking each other's load requests (without prior crash).
* Own multi-threaded processing streams racing to re-do potentially crashed reqiuest.

I wonder if a more productive way to move forward is a demo with this issue discussed in a medium-sized gathering?..

Also, voluminous comments are a "code smell". It alerts reviewer that logic is so complex/convoluted/unrefined that code alone is not no longer self-documenting...

pavel-jm (changes requested):

symmetric-core/src/main/java/org/jumpmind/symmetric/common/ParameterConstants.java
```diff
@@ -159,6 +159,8 @@ private ParameterConstants() {
     public static final String INITIAL_LOAD_CONCAT_CSV_IN_SQL_ENABLED = "initial.load.concat.csv.in.sql.enabled";
     public static final String INITIAL_LOAD_USE_COLUMN_TEMPLATES_ENABLED = "initial.load.use.column.templates.enabled";
     public static final String INITIAL_LOAD_EXTRACT_THREAD_COUNT_PER_SERVER = "initial.load.extract.thread.per.server.count";
+    public static final String INITIAL_LOAD_EXTRACT_REQUEST_RECOVERY_ENABLED = "initial.load.extract.request.recovery.enabled";
```
pavel-jm, line 162: Not sure why this is new parameter is needed?
  I was hoping the issue is fixed without user needing to analyse it and set parameters?

pavel-jm (changes requested):
Let's schedule a code review/debate meeting to settle on the best idea to solve this?

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
gp510: Addressed, and split into the two variables you named plus the readability point.

**Complexity / comment smell.** The clause you quoted (the zero-row heuristic) turned out to be a pure fast-path optimisation sitting on top of an already-exact query, buying nothing but the readability cost you're pointing at, a case it misjudges, and most of the comment volume. Deleted it. The method is now:

```java
protected boolean isExtractRequestComplete(ExtractRequest request, ExtractMode mode) {
    if (mode == ExtractMode.EXTRACT_ONLY) {
        return false; // changeBatchStatus does not persist here, so the batch would still read RQ
    }
    return !hasRequestedBatchesInRange(request);
}
```

`currentBatch` dropped from the signature since it was only ever used by the deleted clause.

**"Own multi-threaded processing streams racing to re-do a potentially crashed request."** Checked this one and don't think it can happen. Thread ids are sticky (`ReloadQueueThreadAssigner` only assigns when null, never reassigns an in-flight request), each request maps to exactly one `(node, reload!N)` queue, that queue is one `sym_node_communication` row guarded by both the in-JVM `executing` set and the row lock, and `execute()` processes requests sequentially within a server. Happy to be shown a path I'm missing, but traced it and didn't find one.

**"Clustered servers overtaking each other's load requests (without prior crash)."** This one is real, and it's the important finding. `ClusterConstants.INITIAL_LOAD_EXTRACT` only covers queueing, not extraction: `queueWork` submits to a thread pool and returns, so the unlock fires before extraction even starts. The lock that actually matters is the `sym_node_communication` row for the request's queue, broken on staleness by `initial.load.extract.timeout.ms`, and nothing refreshes it while a server is extracting. So a server can steal the queue and restart a request that a first server is still genuinely working, which then commits `OK` over a range the second server just reset to `RQ`.

Added a guard for it: before restarting, `recoverStuckExtractRequests` now checks whether the request's queue is still actively locked by another server, and skips it if so, without needing `force`. That is not a stuck request, it's one still being worked.

symmetric-core/src/main/java/org/jumpmind/symmetric/common/ParameterConstants.java
gp510: Both parameters removed.

`initial.load.extract.request.recovery.enabled` is gone. Recovery is now unconditional. A kill switch on a bug fix concedes the detection isn't trustworthy, which is the thing to actually fix if it's true.

`initial.load.extract.request.recovery.threshold.ms` is gone too, replaced by the existing `initial.load.extract.timeout.ms` (2h default). That parameter already means exactly this in the product: the point at which an extract's lock is considered broken (`NodeCommunicationService.getLockTimeoutDate()`). Reusing it is more conservative than the 5 minute default it replaces, and it's one less thing to explain since it's already documented and `DatabaseOverridable`.

It also turned out to matter for the concurrency question raised on the other thread: reusing the same timeout that governs when another server is allowed to steal the queue lock means the two checks agree on what "stale" means, rather than racing on two independent clocks.

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
```diff
@@ -2087,6 +2091,130 @@ public void execute(NodeCommunication nodeCommunication, RemoteNodeStatus status
...
+            return 0;
+        }
+        lastStuckExtractRequestCheckMs = now;
+        List<ExtractRequest> stuck = sqlTemplateDirty.query(getSql("selectStuckExtractRequestsSql"), new ExtractRequestMapper(),
+                engine.getNodeId(), ExtractStatus.OK.name(), new Date(now - thresholdMs));
+        int restarted = 0;
+        for (ExtractRequest request : stuck) {
+            if (!force && isExtractQueueLocked(request)) {
+                log.info("{} Its queue is still locked by another server, so it is still being extracted rather than stuck.",
+                        describeStuckRequest(request));
```
w3bbd3v-jm: There are 6 new sonar warnings here to address. Please fix before merging.

gp510: Pushed a fix for all 6. Two loop continues collapsed into a single if/else-if/else chain (S135), and the three describeStuckRequest log calls now guard on the matching isXxxEnabled() before building the string (S2629 x3), plus an unused import and an unnecessary public modifier in the test (S1128, S5786). Same 20 tests still pass.

PR #995, opened 2026-08-13
SYM-7916: continue past DDL that fails only because the object already exists

gp510:
Fixes [SYM-7916](https://jumpmind.atlassian.net/browse/SYM-7916).

## The problem

Re-sending a table definition to a target where the load already created the objects raises "already exists" (SQL Server 1913, Oracle ORA-00955, MySQL duplicate key name) and fails the batch carrying it. Table definitions ride the **same channel as change data**, so one failed batch holds every batch queued behind it and all replication for that table stops — while the only possible effect of the statement succeeding would have been to create something that was already there.

Observed six times across five dates and three engines at one site. Recovery also required non-obvious operator knowledge: retrying re-issues the same DDL and fails identically, so the batch has to be **ignored** rather than have its error cleared, and nothing said so.

## Why loader tolerance rather than guarded CREATE INDEX

The ticket listed both options. The deciding reason is concrete: **`MsSqlDdlBuilder` and `OracleDdlBuilder` are not in this repository at all.** `symmetric-db/platform` has ase, cassandra, db2, derby, firebird, greenplum, h2, hbase, hsqldb, informix, ingres, interbase, kafka, mysql, nuodb, postgresql, raima, redshift, sqlanywhere, sqlite, voltdb. No mssql, no oracle.

The reported failure is SQL Server 1913, so guarded DDL here would fix PostgreSQL — a platform that is not failing — and leave the affected customer waiting on a separate change.

Supporting: MySQL has no `CREATE INDEX IF NOT EXISTS` at all; Oracle needs a PL/SQL block that then has to survive `SqlScriptReader` delimiter splitting; and a name-based guard silently skips an index that exists with a *different definition*, which is precisely the situation that triggers this — so option 1's masking risk is not actually smaller.

## It needed no new plumbing

`doesObjectAlreadyExist` is **already** on `ISqlTemplate` with `objectAlreadyExistsCodes` / `objectAlreadyExistsStates` behind it, and `JdbcSqlTemplate` already holds the settings object that `ClientSymmetricEngine` populates. So:

- `objectAlreadyExistsMessageParts` adds cross-vendor message matching, which is what gives platforms outside this repository the fix with no per-platform change. Error codes stay per-platform, since the same number means different things on different vendors.
- The DDL statement loop tolerates create/alter failures that classify as already-exists, alongside the existing `isDrop` / `isSequenceCreate` handling, logging the skipped statement and the driver message at WARN.
- When tolerance is off and such a statement fails, the log now says the batch will fail identically on retry and should be ignored rather than cleared.
- `db.tolerate.object.already.exists.on.ddl`, default true, is the revert switch.

## Blast radius

**No interface signature changed and no emitted DDL changed.** The ~20 PRO platform subclasses of `AbstractDdlBuilder` that are invisible from this repo are unaffected, and both `PostgreSqlDdlBuilderTest` exact-string assertions pass untouched.

**A full PRO compile is still required before merge** — the OSS build passing proves nothing about the other platforms.

**1117 tests pass** across `symmetric-jdbc`, `symmetric-db`, `symmetric-io` and `symmetric-core`, 0 failures.

## Two things reviewers should weigh

**Cannot be verified without the vendors.** SQL Server 1913 and Oracle ORA-00955 message matching is what the OSS-only reach rests on, and server-localized messages defeat it. PRO should follow up by populating the per-platform error-code arrays; that is the real fix and this is the fallback. Worth adding both to the PRO integration matrix.

**The parameter is not database-overridable**, and the comment says so explicitly. It is read through `SqlTemplateSettings`, built once at startup from the properties file before `sym_parameter` is available, so it needs a restart. Its neighbours read the same way declare no `DatabaseOverridable` either. If we want it console-settable, the flag belongs on `DatabaseWriterSettings` threaded through `SqlScript` as a peer of `failOnDrops` — a bigger change that would touch `ISqlTemplate`/`IDatabasePlatform` signatures, which is exactly the blast radius this approach avoids. Happy to go that way if reviewers prefer it.

## Worth recording

The deeper cause is a **false-positive `AddIndexChange`**: `ModelComparator.findCorrespondingIndex` matches on full `IIndex.equals`, not name, while 1913 is a name collision. So the diff fires when the read-back model differs in definition from an index whose name already exists at the target. This change suppresses the symptom, not the model mismatch.

[SYM-7916]: https://jumpmind.atlassian.net/browse/SYM-7916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/symmetric-core/src/main/resources/symmetric-default.properties
+++ b/symmetric-core/src/main/resources/symmetric-default.properties
@@ -423,6 +423,19 @@ db.sql.query.timeout.seconds=3600
 # Tags: database
 db.jdbc.isolation.level=
 
+# Continue past a DDL statement that failed only because the object it creates already exists, instead of failing
+# the batch carrying it.  Table definitions travel on the same channel as change data, so a redundant CREATE INDEX
+# or CREATE TABLE stops every batch queued behind it and halts replication for that table -- even though the only
+# possible effect of the statement succeeding would have been to create something that was already there.  The
+# skipped statement is logged at WARN with the driver's message.  Set to false to restore the previous behavior of
+# failing the batch, in which case note that retrying re-issues the same DDL and fails identically, so the batch
+# has to be ignored rather than have its error cleared.
+#
+# DatabaseOverridable: true
+# Tags: database
+# Type: boolean
+db.tolerate.object.already.exists.on.ddl=true
+
 # If set to true forces database columns that contain character data to be read as bytes (bypassing JDBC driver character encoding) 
 # so the raw values be encoded using the system default character set (usually UTF8).  This property was added to bypass MySQL character
 # encoding so the raw data can be converted to utf8 directly.
--- a/symmetric-core/src/main/resources/symmetric-default.properties
+++ b/symmetric-core/src/main/resources/symmetric-default.properties
@@ -431,7 +431,10 @@ db.jdbc.isolation.level=
 # failing the batch, in which case note that retrying re-issues the same DDL and fails identically, so the batch
 # has to be ignored rather than have its error cleared.
 #
-# DatabaseOverridable: true
+# Read through SqlTemplateSettings, which is built once at startup from the properties file, so unlike most parameters
+# this one is not overridable from sym_parameter and a change needs a restart.  Its neighbours in this block that are
+# read the same way (db.jdbc.isolation.level, db.read.strings.as.bytes) likewise declare no DatabaseOverridable.
+#
 # Tags: database
 # Type: boolean
 db.tolerate.object.already.exists.on.ddl=true
--- a/symmetric-core/src/main/resources/symmetric-default.properties
+++ b/symmetric-core/src/main/resources/symmetric-default.properties
@@ -431,6 +431,11 @@ db.jdbc.isolation.level=
 # failing the batch, in which case note that retrying re-issues the same DDL and fails identically, so the batch
 # has to be ignored rather than have its error cleared.
 #
+# Defaulted on deliberately, after narrowing which messages qualify (see objectAlreadyExistsMessageParts in
+# JdbcSqlTemplate). The two message fragments that could plausibly mean something other than "this exact object
+# is already there" (MySQL's duplicate key/column name messages) were removed during review; everything left is
+# an unambiguous statement of existence, so tolerating it does not risk masking real schema drift.
+#
 # Read through SqlTemplateSettings, which is built once at startup from the properties file, so unlike most parameters
 # this one is not overridable from sym_parameter and a change needs a restart.  Its neighbours in this block that are
 # read the same way (db.jdbc.isolation.level, db.read.strings.as.bytes) likewise declare no DatabaseOverridable.
--- a/symmetric-jdbc/src/main/java/org/jumpmind/db/sql/JdbcSqlTemplate.java
+++ b/symmetric-jdbc/src/main/java/org/jumpmind/db/sql/JdbcSqlTemplate.java
@@ -90,6 +90,15 @@ public class JdbcSqlTemplate extends AbstractSqlTemplate implements ISqlTemplate
     protected String[] dataTruncationStates;
     protected int[] objectAlreadyExistsCodes;
     protected String[] objectAlreadyExistsStates;
+    /**
+     * Cross-vendor message fragments identifying an "object already exists" failure, matched case-insensitively. Error codes cannot be shared defaults, since
+     * the same number means different things on different vendors, and only PostgreSQL populates the SQLSTATE list today. These cover platforms whose DDL
+     * builders live outside this repository -- SQL Server 1913 ("an index or statistics with name 'x' already exists"), Oracle ORA-00955 ("name is already used
+     * by an existing object"), MySQL ("Duplicate key name") -- so they get the tolerance with no per-platform change. Server-localized messages defeat message
+     * matching, so a platform that depends on this should still populate {@link #objectAlreadyExistsCodes}.
+     */
+    protected String[] objectAlreadyExistsMessageParts = { "already exists", "already an object named",
+            "is already used by an existing object", "duplicate key name", "duplicate column name" };
     protected int[] objectDoesNotExistCodes;
     protected String[] objectDoesNotExistStates;
     protected int isolationLevel;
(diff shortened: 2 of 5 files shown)


symmetric-core/src/main/resources/symmetric-default.properties
```diff
@@ -423,6 +423,22 @@ db.sql.query.timeout.seconds=3600
...
+# failing the batch, in which case note that retrying re-issues the same DDL and fails identically, so the batch
+# has to be ignored rather than have its error cleared.
+#
+# Read through SqlTemplateSettings, which is built once at startup from the properties file, so unlike most parameters
+# this one is not overridable from sym_parameter and a change needs a restart.  Its neighbours in this block that are
+# read the same way (db.jdbc.isolation.level, db.read.strings.as.bytes) likewise declare no DatabaseOverridable.
+#
+# Tags: database
+# Type: boolean
+db.tolerate.object.already.exists.on.ddl=true
```
pavel-jm, line 440: I wonder if this parameter is unnecessary IF the new code checks for the version of database, to ensure it supports syntax like " ... IF NOT EXISTS ..."
or a wrapper SQL block which prevents creating index with the same name.

symmetric-jdbc/src/main/java/org/jumpmind/db/sql/JdbcSqlTemplate.java
```diff
@@ -90,6 +90,15 @@ public class JdbcSqlTemplate extends AbstractSqlTemplate implements ISqlTemplate
...
     protected String[] objectAlreadyExistsStates;
+    /**
+     * Cross-vendor message fragments identifying an "object already exists" failure, matched case-insensitively. Error codes cannot be shared defaults, since
+     * the same number means different things on different vendors, and only PostgreSQL populates the SQLSTATE list today. These cover platforms whose DDL
+     * builders live outside this repository -- SQL Server 1913 ("an index or statistics with name 'x' already exists"), Oracle ORA-00955 ("name is already used
+     * by an existing object"), MySQL ("Duplicate key name") -- so they get the tolerance with no per-platform change. Server-localized messages defeat message
+     * matching, so a platform that depends on this should still populate {@link #objectAlreadyExistsCodes}.
+     */
+    protected String[] objectAlreadyExistsMessageParts = { "already exists", "already an object named",
+            "is already used by an existing object", "duplicate key name", "duplicate column name" };
```
pavel-jm, line 101: It is a great move to consolidate messages, which indicate that the target object already exists! This allows us to add messages in languages other than English, too. 

I am not yet convinced that these messages simply flag existance vs. out-of-sync target database which does need manually attention to get replication flowing successfully? (We do not want to paint over structural problems only to troubleshoot data import failures 1 batch later)
* duplicate key name
* duplicate column name

pavel-jm (changes requested):

symmetric-jdbc/src/main/java/org/jumpmind/db/sql/JdbcSqlTemplate.java
gp510: Taking the two you flagged separately, because I think you are right about one of them.

duplicate key name — I have direct evidence it is a false positive, not drift. PostgreSQL 16 hub to MySQL 8 spoke, ordinary three-table initial load with create-tables, no deliberate trigger. The control build failed exactly the way the customer reports:

WARN  [JdbcSqlTemplate] Duplicate key name 'item_category_fk'.
  Failed to execute: CREATE INDEX `item_category_fk` ON `item` (`category_id`)
ERROR [ManageIncomingBatchListener] Failed to load batch hub-139
sym_incoming_batch 139 went to ER, sql_code 1061, and stayed there through 188 identical retries with CDC head-of-line blocked behind it.

The mechanism is in our own reader, not in a drifted target. MySqlDdlReader.isInternalForeignKeyIndex returns true when the FK name equals the index name, and AbstractJdbcDdlReader then calls table.removeIndex(indexIdx). So MySQL backs the FK with that index and hides it from the model we just read back. The next create event diffs as "index missing" and emits DDL for something that is demonstrably present:

KEY `item_category_fk` (`category_id`),
CONSTRAINT `item_category_fk` FOREIGN KEY (`category_id`) REFERENCES `category` (`category_id`)
The target is not out of sync. The reader cannot see what is there. That is the case this entry is for.

duplicate column name — you are right and I should not have included it. I only ever exercised 1061. I never tested 1060, and a duplicate column is a much better candidate for genuine drift than a duplicate index name is. I will either drop it or come back with a test that shows a benign case. Dropping it is my inclination.

Worth adding on "painting over": the tolerated path is not silent. It logs at WARN with the failing statement, and when the parameter is off we now emit operator guidance naming the batch as permanently unretryable, which the current code does not do at all. So the change makes a currently-invisible failure visible in both directions.

symmetric-core/src/main/resources/symmetric-default.properties
gp510: Taking the two you flagged separately, because I think you are right about one of them.

duplicate key name — I have direct evidence it is a false positive, not drift. PostgreSQL 16 hub to MySQL 8 spoke, ordinary three-table initial load with create-tables, no deliberate trigger. The control build failed exactly the way the customer reports:

WARN  [JdbcSqlTemplate] Duplicate key name 'item_category_fk'.
  Failed to execute: CREATE INDEX `item_category_fk` ON `item` (`category_id`)
ERROR [ManageIncomingBatchListener] Failed to load batch hub-139
sym_incoming_batch 139 went to ER, sql_code 1061, and stayed there through 188 identical retries with CDC head-of-line blocked behind it.

The mechanism is in our own reader, not in a drifted target. MySqlDdlReader.isInternalForeignKeyIndex returns true when the FK name equals the index name, and AbstractJdbcDdlReader then calls table.removeIndex(indexIdx). So MySQL backs the FK with that index and hides it from the model we just read back. The next create event diffs as "index missing" and emits DDL for something that is demonstrably present:

KEY `item_category_fk` (`category_id`),
CONSTRAINT `item_category_fk` FOREIGN KEY (`category_id`) REFERENCES `category` (`category_id`)
The target is not out of sync. The reader cannot see what is there. That is the case this entry is for.

duplicate column name — you are right and I should not have included it. I only ever exercised 1061. I never tested 1060, and a duplicate column is a much better candidate for genuine drift than a duplicate index name is. I will either drop it or come back with a test that shows a benign case. Dropping it is my inclination.

Worth adding on "painting over": the tolerated path is not silent. It logs at WARN with the failing statement, and when the parameter is off we now emit operator guidance naming the batch as permanently unretryable, which the current code does not do at all. So the change makes a currently-invisible failure visible in both directions.

symmetric-jdbc/src/main/java/org/jumpmind/db/sql/JdbcSqlTemplate.java
gp510: Coming back to this after actually testing it, because my 08-17 reply here does not hold up and I want to correct it before anything else.

**My error first.** I defended `duplicate key name` with `MySqlDdlReader.isInternalForeignKeyIndex` stripping an FK-backing index from the read-back model. That mechanism is real, I confirmed it again just now, but it is MySQL-only. The customer this ticket traces back to targets SQL Server. I answered a SQL Server question with MySQL evidence, and it read as a rebuttal of Evan's redirect when it was not one. Sorry for the noise that caused.

**On the redirect itself: I built a direct test rather than re-arguing the code, and the case-sensitivity mechanism does not reproduce.**

I extracted the actual shipped classes from the customer's own build (`jumpmind/symmetricds-pro:3.17.8.RC4`, unmodified, straight out of the container's `WEB-INF/lib`), read a real table's indexes live over jt400 from the same public IBM i this whole engagement mines, created the matching indexes on a live SQL Server target by hand to represent a target that already went through one successful sync, and called the actual `ModelComparator.detectIndexChanges` with `caseSensitive=false` (which is what `db.metadata.ignore.case=true` produces, the shipped default and this customer's verified value on all four of their engines).

```
=== AS/400 source indexes, as read via jt400 ===
  F42119_2 [SDLNID]
  F42119_3 [SDDRQJ]
  F42119_4 [SDLNID, SDDRQJ]
=== SQL Server target indexes, as read live ===
  f42119_2 [SDLNID]
  f42119_3 [sddrqj]
  f42119_4 [SDLNID, sddrqj]

changes detected: 0
```

Zero changes, with the AS/400's uppercase names run through the same lowercasing
`alterCaseToMatchDatabaseDefaultCase` would apply. As a sanity check that this harness actually detects
real differences and isn't just always returning zero, I added one more index on the AS/400 side only and
reran with no other change: `changes detected: 1`, correctly identifying that exact index. So the first
result is a real finding, not a harness bug.

@evan-miller-jumpmind, I think this is consistent with something else: `findCorrespondingIndex` already
branches to `equalsIgnoreCase` when `db.metadata.ignore.case` is true (`ModelComparator.java:537-546`), and
I believe `816d6b4510` already removed the genuinely case-sensitive re-check that used to sit right after
it in `detectIndexChanges`. If you do have a live repro, the thing I'd ask about is whether
`db.metadata.ignore.case` was set to `false` in that environment — because with it false,
`findCorrespondingIndex` falls through to `IIndex.equals()`, which also compares `platformIndexes`, and
that is a different, real bug I have not tested here.

Full writeup and the test harness: `Demo Central/labs/mssql-runtime-journal-user-poc/sym7916-repro/` (internal lab, referencing it for anyone who wants to rerun it).

**Where this leaves the PR.** I don't think this settles what *did* cause the customer's failures, only that this specific mechanism didn't. The strongest lead left is the customer's own log: the failing index name advances over several days (`f0101_10`, then `_11`, then `_12`/`_13` on later dates), while the earlier-numbered indexes on the same table never fail. That is more consistent with a stored DDL script being reapplied against a target that has moved on than with a live comparison bug, but I have not reproduced that specific sequence and am not asserting it as proven.

Given that, I'd rather narrow this PR than keep defending the full signature list. Dropping `duplicate column name` (already agreed) and now `duplicate key name` as well, since it only ever covered the MySQL mechanism and that deserves its own fix on the emission side rather than living in the tolerance list. Will push that update along with closing the Sonar gap.

PR #998, opened 2026-08-17
SYM-7712: Fixed triggers.removed.count metric not getting incremented

evan-miller-jumpmind:


w3bbd3v-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/TriggerRouterService.java
```diff
@@ -1883,6 +1883,7 @@ protected void dropTrigger(StringBuilder sqlBuffer, String catalog, String schem
         if (StringUtils.isNotBlank(triggerName)) {
             try {
                 symmetricDialect.removeTrigger(sqlBuffer, catalog, schema, triggerName, tableName);
+                statisticManager.incrementTriggersRemovedCount(1);
```
w3bbd3v-jm, line 1886: Is this a new pattern we need to do to capture and maintain accurate stats? 
If so, this is somewhat concerning to me, b/c for every remove trigger call, we have to remember to increment the count using the statistics manager. Maybe I'm overthinking it? Instead, we could use a wrapper class that would `removeAndIncrement` for us.

evan-miller-jumpmind: Yes, `StatisticManager` methods like `incrementTriggersRemovedCount()` are a new pattern that we're adding in 3.18. Many of the metrics are updated using this kind of method. This is the only place where Sync Triggers calls the `removeTrigger()` method so it doesn't seem to me like a wrapper class is necessary.

PR #1000, opened 2026-08-18
SYM-7933: Fix decimal values using wrong scale on Sybase ASE by removing SybaseJdbcSqlTemplate's jConnect-specific setDecimalValue override

alexandraboffo8:


symmetric-jdbc/src/main/java/org/jumpmind/db/platform/sybase/SybaseJdbcSqlTemplate.java
```diff
@@ -37,13 +32,9 @@
 import org.jumpmind.db.sql.JdbcSqlTemplate;
 import org.jumpmind.db.sql.SqlTemplateSettings;
 import org.jumpmind.db.sql.SymmetricLobHandler;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.springframework.jdbc.core.SqlTypeValue;
 
 public class SybaseJdbcSqlTemplate extends JdbcSqlTemplate implements ISqlTemplate {
-    private static final Logger log = LoggerFactory.getLogger(SybaseJdbcSqlTemplate.class);
-    protected static final String NATIVE_PREPARED_STATEMENT_NAME = "com.sybase.jdbc4.jdbc.SybPreparedStatement";
```
pavel-jm, line 46: It is my understanding that this override was put in place because of truncation (loss of precision) for floating point column types (NUMERIC, DECIMAL).
The DEV card does not specify why this solution is no longer needed?!

Just want to understand what steps were taken to ensure floating point continues to work correctly on Sybase ?

w3bbd3v-jm (approved):

PR #1018, opened 2026-08-21
SYM-7817: Added support for Azure Database for PostgreSQL endpoints (Pro edition)

evan-miller-jumpmind:


symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
```diff
@@ -886,38 +886,11 @@ private void startNodeAndJobs(Node node, boolean startJobs) {
     protected void checkForProOnlyDatabase() {
```
w3bbd3v-jm: Really like the simplification in how this method is implemented! 
Can we move it to a new helper class, like `ProOnlyDatabaseValidator` and the `start` method would just call `validate` , something like `new ProOnlyDatabaseValidator(platform).validate();`?

symmetric-core/src/main/java/org/jumpmind/symmetric/AbstractSymmetricEngine.java
```diff
@@ -886,38 +886,11 @@ private void startNodeAndJobs(Node node, boolean startJobs) {
...
-        }
-        if (DatabaseNamesConstants.CLOUDSQL_MYSQL.equalsIgnoreCase(dbVersionName)
-                && !DatabaseNamesConstants.CLOUDSQL_MYSQL.equalsIgnoreCase(platform.getName())) {
-            throw new SymmetricException(
-                    "The detected database platform '%s' is not supported in SymmetricDS open source. "
-                            + "Google Cloud SQL for MySQL requires SymmetricDS Pro. "
-                            + "Contact the SymmetricDS sales team for more information.",
-                    dbVersionName);
-        }
+        checkCloudDatabaseRequiresPro(DatabaseNamesConstants.AURORA_POSTGRESQL, dbVersionName, "AWS Aurora PostgreSQL");
```
w3bbd3v-jm: Instead of adding a new call to  `checkCloudDatabaseRequiresPro` can we add these DB names to a static list instead and just loop through it?

w3bbd3v-jm (approved):

PR #1019, opened 2026-08-24
SYM-7802: Dedicated Startup Parameter service (for consolidating parameter import and updates)

evan-miller-jumpmind:
The purpose of this PR is to add a startup parameter service that consolidates all parameters that are available without database access. This allows for easier debugging, including the ability to log a dump of all startup parameters, including their values and where they're set. Here are some design decisions that we've made:

- A new `StartupParameter` model object should be created that has a name, type, value, default value, source, engine name, and `isSensitive` flag
- The `StartupParameterService` should be a singleton, similar to `MetricsManager`
- The `StartupParameterService` must keep track of each engine as well as global startup parameter values (represented by an engine name of "*")
- The `IStartupParameterService` should have 2 APIs: one for retrieving engine-level startup parameters and another for retrieving global startup parameters
- The `ParameterService` should get the values of startup parameters from the `StartupParameterService`


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/StartupParameterService.java
```diff
@@ -0,0 +1,344 @@
...
+            StartupParameter parameter = resolve(key);
+            if (isTrackable(parameter)) {
+                parameters.put(key, parameter);
+            }
+        }
+    }
+
+    private boolean isTrackable(StartupParameter parameter) {
+        ParameterMetaData metaData = parameterMetaData.get(parameter.name());
+        boolean isRuntimeParameter = metaData != null && metaData.isDatabaseOverridable();
```
pavel-jm: I wonder if there is a logical discrepancy between isTrackable & isDatabaseOverridable?

A true "Startup parameter" cannot be also "isDatabaseOverridable", because database is not yet initialized at startup?

Let's schedule a brainstorming meeting for review (not urgent).

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/StartupParameterService.java
```diff
@@ -0,0 +1,344 @@
...
+        try {
+            return Double.parseDouble(value.trim());
+        } catch (NumberFormatException ex) {
+            TypedProperties.logPropertiesException(log, key, value);
+            return defaultValue;
+        }
+    }
+
+    @Override
+    public StartupParameter getParameter(String key) {
```
pavel-jm: Initial concerns for us to review in the next collaboration meeting:
   
  1. Cache-defeat perf bug — StartupParameterService.getParameter() only caches a resolved value into its parameters map when isTrackable() is true. But isTrackable()
     deliberately returns false for any database-overridable parameter still at its default — the common case for most ParameterConstants entries. So for those keys, every
     call re-runs resolve() from scratch, including a full TypedProperties copy in substituteTokensInDefault() whenever the default contains a $(...) token (e.g.
     cluster.server.id's default is literally $(hostName)). The class is trying to be both "the lean, reportable snapshot" (getAllParameters()/dumpAsText()) and "a
     general-purpose cached lookup" with the same gate — and the caching half loses for the majority of keys.

  2. Two divergent instances, no rule for which to use — AbstractCommandLauncher keeps one static, server.properties-scoped IStartupParameterService; each
     AbstractSymmetricEngine builds its own separate, engine.properties-scoped one. refresh()/refreshSystemProperty() are only ever called on the static one
     (SymmetricWebServer.setSystemProperty); a repo-wide grep shows refresh() has zero production callers anywhere, and no per-engine instance is ever refreshed after
     construction. So a system property changed after an engine boots can silently disagree between the two "authoritative" instances — exactly the divergence the service's
     own Javadoc says it exists to prevent.

symmetric-client/src/main/java/org/jumpmind/symmetric/AbstractCommandLauncher.java
```diff
@@ -108,15 +114,22 @@ protected static void initFromServerProperties() {
             File serverPropertiesFile = new File(DEFAULT_SERVER_PROPERTIES);
             if (!serverPropertiesFile.exists()) {
                 log.debug("Failed to load " + DEFAULT_SERVER_PROPERTIES + ". File does not exist.");
+                StartupParameterService.getInstance().registerGlobal(new TypedProperties(), new HashMap<String, Source>());
```
w3bbd3v-jm: Normally we should be using the interface not the concrete class. It is possible for us to register the concrete class in one place on startup? Then the code can safely use the interface at runtime and preserve encapsulation in the code.

evan-miller-jumpmind: Yes, I'll work on this change

pavel-jm (approved):

PR #1024, opened 2026-08-25
SYM-7950: Upgraded Jetty version to 12.1.12

evan-miller-jumpmind:
I also upgraded Spring Boot because it was pulling in an older version of Jetty.


symmetric-assemble/common.gradle
```diff
@@ -339,12 +339,12 @@ subprojects { subproject ->
         servletVersion = '4.0.1'
         springVersion = '7.0.8'
```
w3bbd3v-jm: Spring Boot 4.1.1 brings in Spring 7.0.9. Can we keep that in sync, please?
Reference: https://docs.spring.io/spring-boot/appendix/dependency-versions/coordinates.html

w3bbd3v-jm (approved):

PR #1033, opened 2026-08-26
SYM-7938: Avoid blocking registration batch ack for long running SyncTriggers for triggers with templete variables

Sai-Gumpalli:


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/TriggerRouterService.java
```diff
@@ -2038,6 +2044,7 @@ public boolean syncTriggers(String targetExternalId, boolean force) {
                 return syncTriggers(tables, force);
```
evan-miller-jumpmind: Shouldn't we should add another call to `activeSyncTriggersNodes.remove(targetExternalId)` before this early `return`?

evan-miller-jumpmind: Maybe this isn't the correct approach because it would remove the external ID from the map before sync triggers is complete. The problem is that the external ID never gets removed from the map once sync triggers completes.

evan-miller-jumpmind (approved):

w3bbd3v-jm (approved):

PR #1036, opened 2026-08-27
SYM-7724: Fixed rows.create.time.min and rows.create.time.max metrics not tracking custom channels

evan-miller-jumpmind:


symmetric-core/src/main/java/org/jumpmind/symmetric/statistic/StatisticManager.java
```diff
@@ -950,7 +950,7 @@ private void setChannelLongGauge(String metricId, String channelId, long value)
             IEngineMetricsService svc = engine.getMetricsService();
             if (svc == null)
                 return;
-            ISymLongGauge g = svc.getLongGauge(metricId, MetricAttributeList.of(new MetricAttribute(CHANNEL, channelId)));
+            ISymLongGauge g = svc.registerLongGauge(metricId, MetricAttributeList.of(new MetricAttribute(CHANNEL, channelId)));
```
w3bbd3v-jm, line 953: Do we need to apply a like-fix for `setChannelDoubleGauge` or `setEngineDoubleGauge`, etc.?

evan-miller-jumpmind: A fix for `setChannelDoubleGauge` will be necessary, but it's out of scope for this ticket. This method is used for the `rows.unrouted.channel.count` metric, which is covered by [SYM-7722](https://jumpmind.atlassian.net/browse/SYM-7722). This ticket was marked as low priority.

The `setEngineDoubleGauge` method doesn't need this fix. It's only used for metrics that aren't channel-scoped.

w3bbd3v-jm (approved):

PR #1040, opened 2026-08-27
SYM-7938: Avoid blocking registration batch ack for long running SyncTriggers for triggers with templete variables

Sai-Gumpalli:


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/TriggerRouterService.java
```diff
@@ -2022,23 +2024,35 @@ private boolean containsExactMatchForSourceTableName(Table table, List<Trigger>
...
+                if (!tables.isEmpty()) {
+                    return syncTriggers(tables, force);
+                }
             }
-            if (tables.size() > 0) {
-                return syncTriggers(tables, force);
-            }
+            return true;
+        } catch (Exception e) {
+            log.error("Error while Syncing Tringgers for node {}", targetExternalId);
```
pavel-jm: Let's combine these two lines into one, like this?
`log.error("Error while Syncing Triggers for node "+ targetExternalId, e);`

pavel-jm: Let's change "Tringgers" to "Triggers"

pavel-jm (changes requested):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/TriggerRouterService.java
```diff
@@ -2022,23 +2024,34 @@ private boolean containsExactMatchForSourceTableName(Table table, List<Trigger>
...
-            List<Table> tables = new ArrayList<Table>();
-            for (Trigger trigger : triggers) {
-                if (trigger.getSourceTableName().contains("targetExternalId")) {
-                    Table table = platform.readTableFromDatabase(trigger.getSourceCatalogName(), trigger.getSourceSchemaName(),
-                            FormatUtils.replace("targetExternalId", targetExternalId, trigger.getSourceTableName()));
-                    if (table != null) {
-                        tables.add(table);
+        if (!activeSyncTriggersNodes.add(targetExternalId)) {
+            log.info("Sync Triggers is already running for node {}", targetExternalId);
+            return true;
```
w3bbd3v-jm: Should this really return `false` since `true` would mean sync triggers ran here? Might be good to know if we keep attempting to run the job while it's already running.

pavel-jm (commented):
Thank you!

pavel-jm (approved):

PR #1041, opened 2026-08-27
SYM-7723: Fixed rows.create.time.min and rows.create.time.max metrics not tracking unrouted data as expected

evan-miller-jumpmind:
The problem was that the `rows.create.time.min` and `rows.create.time.max` metrics are supposed to track the create times of unrouted data (according to their documentation), but instead they were tracking routed data. This is because their statistics were populated the same way as the `data_min_create_time` and `data_max_create_time` columns in `sym_node_host_channel_stats`, which track routed data.

To make this distinction clear, I added new `setDataUnroutedMinCreateTime()` and `setDataUnroutedMaxCreateTime()` methods to `IStatisticManager`, renamed the existing methods for routed data, and added comments to clarify what each method does.

It can be expensive to query for unrouted data, so having the Routing job run these queries would be impractical. Instead, I added a new `RefreshDataCreateTimeMetricsJob` to query the unrouted data and record the correct statistics. I followed the example of the existing `RefreshBacklogReportJob`, which also runs potentially expensive queries for metrics every 15 minutes.

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RouterService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RouterService.java
@@ -33,6 +33,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
@@ -45,7 +46,6 @@ import org.apache.commons.lang3.StringUtils;
 import org.jumpmind.db.model.Column;
 import org.jumpmind.db.model.Relation;
 import org.jumpmind.db.model.Table;
-import org.jumpmind.db.sql.ISqlRowMapper;
 import org.jumpmind.db.sql.ISqlTransaction;
 import org.jumpmind.db.sql.Row;
 import org.jumpmind.symmetric.ISymmetricEngine;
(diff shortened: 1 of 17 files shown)


symmetric-core/src/test/java/org/jumpmind/symmetric/service/impl/RouterServiceTest.java
```diff
@@ -353,4 +362,103 @@ private ChannelRouterContext newContext() {
...
+        IExtensionService extensionService = mock(IExtensionService.class);
+        when(engine.getExtensionService()).thenReturn(extensionService);
+        RouterService testRouterService = new RouterService(engine);
+        testRouterService.setSqlMap(new RouterServiceSqlMap(databasePlatform,
+                Collections.singletonMap("data", TableConstants.getTableName("sym", TableConstants.SYM_DATA))));
+        return testRouterService;
+    }
+
+    @Test
+    public void testBuildGapQualifiedQueryUsesGapRangeSqlWhenGapCountAtOrBelowThreshold() {
```
w3bbd3v-jm: We can drop the `public` access modifier from these tests.

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RouterService.java
```diff
@@ -1385,4 +1400,28 @@ protected boolean isAllDataRead() {
         }
         return true;
     }
+
+    protected record GapQualifiedQuery(String sql, Object[] args, int[] types) {
```
w3bbd3v-jm, line 1404: Do you think this record has potential outside of the `RouterService`? Wondering if we move it to its own class. I'd like to start moving stuff out of this service too ideally,

w3bbd3v-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RouterService.java
evan-miller-jumpmind: If we keep the name of `GapQualifiedQuery`, then I think it refers to functionality that is pretty specific to `RouterService`. We could move it into its own class and rename it to something more generic.

evan-miller-jumpmind: Would you like me to move it into its own class and rename it to just `Query`? Or if you have a better name, let me know.

w3bbd3v-jm: I'm rethinking this...unless you wanted to spend more time on it. It would be getting outside the scope of this ticket though. I could see it using a builder as well.

mdrouard-jumpmind (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RouterService.java
evan-miller-jumpmind: Ok, I'll leave it as it is for this ticket.

PR #1043, opened 2026-08-27
SYM-7929: Deferred foreign keys and indexes silently dropped when an initial load uses multiple extract threads

ShivamJumpmind:
Please refer to JIRA ticket for details: [Deferred foreign keys and indexes silently dropped when an initial load uses multiple extract threads](https://jumpmind.atlassian.net/browse/SYM-7929)


symmetric-core/src/test/java/org/jumpmind/symmetric/service/impl/DataExtractorServiceDeferredConstraintsTest.java
evan-miller-jumpmind, line 1: Could these new tests be included in the existing `DataExtractorServiceTest` instead of creating a new test file? Typically we don't have more than one test file per class.

Also, there's a Sonar alert in this class that is very easy to fix.

evan-miller-jumpmind (approved):
Looks good to me. There is one more Sonar alert that would be easy to resolve though.

symmetric-core/src/main/java/org/jumpmind/symmetric/extract/SelectFromSymDataSource.java
```diff
@@ -319,7 +319,9 @@ protected boolean processCreateEvent(TriggerHistory triggerHistory, String route
         if (outgoingBatch.getLoadId() > 0) {
             TableReloadStatus tableReloadStatus = dataService.getTableReloadStatusByLoadIdAndSourceNodeId(outgoingBatch.getLoadId(), engine.getNodeId());
             if (tableReloadStatus != null && tableReloadStatus.isCompleted()) {
-                // Ignore create table (indexes and foreign keys) at end of load if it was cancelled
+                // Ignore create table (indexes and foreign keys) at end of load if it was cancelled or load is already complete
+                log.warn("Discarding create event (FKs/indexes) for batch {} table {} because load {} is already complete or load was cancelled",
```
w3bbd3v-jm: I think by adding your log statement, you no longer need the code comment.

Some of the log statement could be marginally improved. Can you change to: 
"Discarding create event (FKs/indexes) for batch {} and table {}, because load {} is already complete or cancelled."

w3bbd3v-jm (approved):

symmetric-core/src/test/java/org/jumpmind/symmetric/service/impl/DataExtractorServiceTest.java
```diff
@@ -104,24 +161,125 @@ public void selectFromSymDataSource_csvValuesAreExtracted_triggerRouterIsNotMark
...
+        assertEquals(2, service.sendPasses.get(),
+                "a load completed through the batch-loaded path must release its deferred-constraints claim");
+    }
+
     @Test
     void extract_failureDuringSendPhase_incrementsDataSentErrorsOnly() throws Exception {
         IStatisticManager statisticManager = mock(IStatisticManager.class);
         IOutgoingBatchService outgoingBatchService = mock(IOutgoingBatchService.class);
-        DataExtractorService service = spy(buildExtractService(statisticManager, outgoingBatchService));
+        DataExtractorService extractService = spy(buildExtractService(statisticManager, outgoingBatchService));
```
w3bbd3v-jm, line 269: Adding the instance variables to the unit test class brought up new sonar warnings. You could try to consolidate these so that the test class is a little cleaner.

PR #1044, opened 2026-08-28
SYM-7726: Fixed rows.sent.errors metric not getting updated

evan-miller-jumpmind:


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
w3bbd3v-jm, line 602: By removing this are we somehow dropping any of the existing batch statistics functionality? Or is it being replaced somewhere I'm not seeing? I ask because this is one of the scarier methods to have to change.

evan-miller-jumpmind: What removal are you referring to? Your comment is on a line of code that I didn't change.

w3bbd3v-jm: lol, that's odd -- when I looked at it, it had the red-ish line where it was removed.  Please disregard!

w3bbd3v-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/DataExtractorService.java
evan-miller-jumpmind: When you click a line of code it highlights it in yellow, so that might have been it. Otherwise, I don't know what would have caused that either.

PR #1047, opened 2026-08-28
SYM-7722: Fixed rows.unrouted.channel.count metric not getting updated

evan-miller-jumpmind:
There were 2 problems to fix:

1. The `rows.unrouted.channel.count` metric has no stats for custom channels
2. The `rows.unrouted.channel.count` doesn't track unrouted data for disabled channels

To fix problem 1, I took a similar approach to what I did for SYM-7725 and added an `IEngineMetricsService.registerDoubleGauge()` method which adds a new channel rather than skipping it because it doesn't exist.

To fix problem 2, I added a new `IRouterService.findUnroutedDataCountByChannel()` method to retrieve the unrouted data count per channel, including disabled channels. I had `RefreshDataCreateTimeMetricsJob` run this query and renamed it to `RefreshUnroutedDataMetricsJob` because it now covers more than just the create time metrics. I made sure to avoid making the same mistakes from SYM-7968 through SYM-7970.

The code that used to populate these metrics was in `RouterService.routeDataForChannel()`. It also populated the `sym_node_host_channel_stats.data_unrouted` column, which was affected by the same issue. Rather than leaving this code in place, I removed it and had `RefreshUnroutedDataMetricsJob` populate both places. I kept the `routing.collect.stats.unrouted` parameter gate but changed its default value to `true` because the query will no longer slow down the Routing job.

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RouterServiceSqlMap.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RouterServiceSqlMap.java
@@ -35,6 +35,10 @@ public class RouterServiceSqlMap extends AbstractSqlMap {
         putSql("selectChannelDataCreateTimeRangeUsingStartDataId",
                 "select channel_id, min(create_time) as min_create_time, max(create_time) as max_create_time "
                         + "from $(data) where data_id >= ? group by channel_id");
+        putSql("selectChannelDataUnroutedCountSql",
+                "select d.channel_id, count(*) as unrouted_count from $(data) d "
+                        + "where exists (select 1 from $(data_gap) g where g.is_expired = 0 and d.data_id between g.start_id and g.end_id) "
+                        + "group by d.channel_id");
         putSql("selectDataUsingGapsSql",
                 "select $(selectDataUsingGapsSqlHint) d.data_id, d.table_name, d.event_type, d.row_data as row_data, d.pk_data as pk_data, d.old_data as old_data, "
                         + " d.create_time, d.trigger_hist_id, d.channel_id, d.transaction_id, d.source_node_id, d.external_data, d.node_list, d.is_prerouted "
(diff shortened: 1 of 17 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RouterServiceSqlMap.java
```diff
@@ -35,6 +35,10 @@ public RouterServiceSqlMap(IDatabasePlatform platform, Map<String, String> repla
         putSql("selectChannelDataCreateTimeRangeUsingStartDataId",
                 "select channel_id, min(create_time) as min_create_time, max(create_time) as max_create_time "
                         + "from $(data) where data_id >= ? group by channel_id");
+        putSql("selectChannelDataUnroutedCountSql",
+                "select d.channel_id, count(*) as unrouted_count from $(data) d "
```
pavel-jm: On a very large system (with millions of records in SYM_DATA, this query) will be expensive, because each DATA row initiates a query against (potentially large) DATA_EVENT table.

evan-miller-jumpmind: Sure, does this give you the information you're looking for, or do I need to run it on a larger database?

Image

Image

pavel-jm (approved):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RouterServiceSqlMap.java
pavel-jm: I just ran a test on a 100M table and 10M gaps. Let's use a JOIN in a reversed order instead: 

  SELECT d.channel_id, count(*) AS unrouted_count
  FROM sym_data_gap g
  JOIN sym_data d
      ON d.data_id BETWEEN g.start_id AND g.end_id
  WHERE g.is_expired = 0
  GROUP BY d.channel_id

In my testing new join had improved query runtime from 9 minutes to 5 seconds (100M rec table).

pavel-jm (changes requested):

symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/RouterServiceSqlMap.java
evan-miller-jumpmind: Thanks for testing that out! I will make the change.

pavel-jm (approved):
Thank you!

PR #1050, opened 2026-08-31
SYM-7972: Prevented container instance ID from regenerating on a restart due to unreliable MAC-address fingerprint validation

evan-miller-jumpmind:
The try-with-resources changes were necessary to get the unit tests working and to prevent potential Sonar warnings. The actual fix is on lines 182-183, where I made it skip the `ClusterHardwareInstanceGenerator.isValid()` check for containerized deployments. This check's sole purpose is to verify that the MAC address hasn't changed, which doesn't make sense for containers.

### DIFF
--- a/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/ClusterService.java
+++ b/symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/ClusterService.java
@@ -52,6 +52,7 @@ import static org.jumpmind.symmetric.service.ClusterConstants.WATCHDOG;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
+import java.io.InputStream;
 import java.net.URL;
 import java.nio.charset.Charset;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


symmetric-core/src/main/java/org/jumpmind/symmetric/service/impl/ClusterService.java
```diff
@@ -195,7 +197,9 @@ public static String initInstanceId(IClusterInstanceGenerator generator) {
             if (instanceIdFile != null) {
                 try {
                     instanceIdFile.getParentFile().mkdirs();
-                    IOUtils.write(newInstanceId, new FileOutputStream(instanceIdFile), Charset.defaultCharset());
+                    try (FileOutputStream out = new FileOutputStream(instanceIdFile)) {
+                        IOUtils.write(newInstanceId, out, Charset.defaultCharset());
+                    }
```
pavel-jm: Any failure inside this new TRY block would be a mystery to both user and support team. 
Let's separate this into two non-overlapping TRY blocks.

w3bbd3v-jm (approved):

PR #1052, opened 2026-08-31
SYM-7894: Spring 7+ update in docs

pavel-jm:


symmetric-assemble/gradle.properties
```diff
@@ -17,7 +17,7 @@ vaadinVersion=25.2.6
 vaadinPluginVersion=25.2.6
 org.gradle.daemon=false
 org.gradle.jvmargs=-Xmx2g
-latestPreviousVersion=3.17.0
+latestPreviousVersion=3.17.8
```
w3bbd3v-jm: The last release is still 3.17.7. Not sure if that makes a difference here.

w3bbd3v-jm (approved):

symmetric-assemble/gradle.properties
pavel-jm: Cool! Thanks

PR #1060, opened 2026-09-01
SYM-7737: Added support for distinguishing between STORED and VIRTUAL generated columns on PostgreSQL

evan-miller-jumpmind:
Unlike other databases, before this change, Postgres actually only supported STORED generated columns. Support for VIRTUAL generated columns was added in Postgres 18.

The new `nonPersistedGeneratedColumnsSupported` flag in `DatabaseInfo` defaults to `true` because databases other than Postgres added support for non-persisted generated columns before (or at the same time as) they added support for persisted generated columns. The `generatedColumnsSupported` flag takes precedence, so if it's `false`, then this new flag won't be checked.

### DIFF
--- a/symmetric-client/src/test/java/org/jumpmind/symmetric/db/postgresql/PostgreSqlSymmetricDialectTest.java
+++ b/symmetric-client/src/test/java/org/jumpmind/symmetric/db/postgresql/PostgreSqlSymmetricDialectTest.java
@@ -67,6 +67,7 @@ class PostgreSqlSymmetricDialectTest {
 
     @Test
     void constructor_enablesPersistedButNotNonPersistedGeneratedColumnSupport_atVersion14() {
+        setupWithMajorVersion(14);
         assertTrue(platform.getDatabaseInfo().isPersistedGeneratedColumnsSupported());
         assertFalse(platform.getDatabaseInfo().isNonPersistedGeneratedColumnsSupported());
     }
--- a/symmetric-client/src/test/java/org/jumpmind/symmetric/db/postgresql/PostgreSqlSymmetricDialectTest.java
+++ b/symmetric-client/src/test/java/org/jumpmind/symmetric/db/postgresql/PostgreSqlSymmetricDialectTest.java
@@ -20,8 +20,10 @@
  */
 package org.jumpmind.symmetric.db.postgresql;
 
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertSame;
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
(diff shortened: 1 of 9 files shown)


symmetric-db/src/main/java/org/jumpmind/db/model/Table.java
```diff
@@ -378,15 +378,22 @@ public boolean doesIndexContainNonPersistedGeneratedColumn(IIndex index) {
      * Determines whether an index can be created on this table for the given platform. Generated columns that the platform can't index (either because it
      * physically stores the value differently than the source, or because it doesn't materialize the value at all) are not indexable there, unless the platform
      * doesn't support generated columns in the first place, in which case the column is replicated as a plain column and any index is fine.
+     * <p>
+     * A platform that doesn't support non-persisted generated columns at all falls back to creating any such column as persisted instead, so a non-persisted
```
pavel-jm: Oh, these double-negatives make my head spin! :-)
  Let's focus on what this method does, please?

Better yet, use longer variable names (hasPersistenGeneratedColumns) to explain rules as they are applied?

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDdlReader.java
```diff
@@ -108,10 +108,46 @@ protected Relation readRelation(Connection connection, DatabaseMetaDataWrapper m
...
+                if (column.isGenerated()) {
+                    column.setPersisted(true);
+                }
+            }
+            return;
+        }
+        String sql = "SELECT a.attname, a.attgenerated FROM pg_attribute a "
+                + "JOIN pg_class c ON a.attrelid = c.oid "
+                + "JOIN pg_namespace n ON n.oid = c.relnamespace "
+                + "WHERE c.relname = ? AND n.nspname = ? AND a.attgenerated <> ''";
```
pavel-jm, line 136: There is a potential for slowness here, might be critical for customers with several hundred tables!?

Let's brainstorm together with additional team members?

pavel-jm (changes requested):

symmetric-client/src/test/java/org/jumpmind/symmetric/db/postgresql/PostgreSqlSymmetricDialectTest.java
```diff
@@ -56,6 +58,38 @@ void setup() {
...
+    @Test
+    void constructor_disablesPersistedAndNonPersistedGeneratedColumnSupport_belowVersion12() {
+        setupWithMajorVersion(11);
+        assertFalse(platform.getDatabaseInfo().isPersistedGeneratedColumnsSupported());
+        assertFalse(platform.getDatabaseInfo().isNonPersistedGeneratedColumnsSupported());
+    }
+
+    @Test
+    void constructor_enablesPersistedButNotNonPersistedGeneratedColumnSupport_atVersion14() {
+        assertTrue(platform.getDatabaseInfo().isPersistedGeneratedColumnsSupported());
```
w3bbd3v-jm, line 71: should we add `setupWithMajorVersion(14);` just in case the default changes in the setup? Then the test name would lie to us.

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDdlReader.java
evan-miller-jumpmind: One idea is to have `readRelations()` (the method that handles reading all tables in a schema) run one query for the whole schema and cache the results, then have `markGeneratedColumnStorageKind()` check that cache. If anyone has any other ideas, please let me know. Or I could schedule a meeting to discuss this.

symmetric-db/src/main/java/org/jumpmind/db/model/Table.java
evan-miller-jumpmind: I tried to make it easier to read. Let me know if looks good or if you have any changes in mind.

symmetric-db/src/main/java/org/jumpmind/db/model/Table.java
```diff
@@ -375,18 +375,31 @@ public boolean doesIndexContainNonPersistedGeneratedColumn(IIndex index) {
...
+     * </ol>
      */
     public boolean canCreateIndex(IIndex index, DatabaseInfo databaseInfo) {
         if (!databaseInfo.isGeneratedColumnsSupported()) {
             return true;
         }
-        if (doesIndexContainPersistedGeneratedColumn(index) && !databaseInfo.isPersistedGeneratedColumnsSupported()) {
+        boolean indexHasPersistedGeneratedColumn = doesIndexContainPersistedGeneratedColumn(index);
+        boolean indexHasNonPersistedGeneratedColumn = doesIndexContainNonPersistedGeneratedColumn(index);
+        boolean targetSupportsNonPersistedGeneratedColumns = databaseInfo.isNonPersistedGeneratedColumnsSupported();
```
w3bbd3v-jm: Could we inline this in the `if` statement? It seems simpilier IMO since there are other 2 boolean variables.

symmetric-db/src/main/java/org/jumpmind/db/model/Table.java
pavel-jm: Thank you!

pavel-jm (approved):

symmetric-jdbc/src/main/java/org/jumpmind/db/platform/postgresql/PostgreSqlDdlReader.java
pavel-jm: Good idea, let's invite all experienced engineers to discuss this!

I would love for us to have some meta-data regarding on the Relation object regarding how it was sources individual read/bulk read... There are still scenarios where an individual read is needed - so this table-level query is still needed!

PR #1061, opened 2026-09-02
SYM-7739: Added support for STORED GENERATED columns on MySQL

evan-miller-jumpmind:


w3bbd3v-jm (approved):
Looks good to me! Just a few minor Sonar things to address.

PR #1063, opened 2026-09-02
SYM-7961: Document SQL Server Change Tracking Limitation for Intermediate Row Values

Sai-Gumpalli:


evan-miller-jumpmind (commented):
Is it true that if a row is inserted and then updated before the Log Miner runs, SQL Server Change Tracking will only report only one event (either an insert or an update, I'm not sure which but maybe you know)? If so, it may be worth mentioning this and how it can cause unexpected behavior when the trigger is only configured to capture certain event types.

Also, the ticket number in the PR title has a typo.

symmetric-assemble/src/asciidoc/appendix/mssqlserver.ad
```diff
@@ -115,6 +115,9 @@ if it's not already enabled, which requires ALTER TABLE permission.
 [source, SQL]
 alter table mytable enable change_tracking;
 
+==== Known Limitations
+* When using Log Miner, SQL Server Change Tracking does not store row values in its change table; it only records the row's primary key and the type of change (insert, update, or delete). As a result, only the current value of a row is captured, not the value it held at the time of the change. If a row is updated multiple times before SymmetricDS reads the changes, or is later updated again before the change is captured, the intermediate values are lost and only the most recent value is synchronized. This is a limitation of SQL Server Change Tracking.
```
pavel-jm: Could you put some distance between SymDS and database, please? 

Example:
When using Log Miner mode of SymmetricDS with Microsoft SQL Server the Change Tracking feature is used to facilitate transfer about change events from the database to SymmetricDS. This SQL Server  feature has limitations to be aware of ...

w3bbd3v-jm (approved):
Seems better to me. Thanks!

evan-miller-jumpmind (approved):

symmetric-assemble/src/asciidoc/appendix/mssqlserver.ad
pavel-jm: Thank you!

pavel-jm (approved):

PR #1066, opened 2026-09-03
SYM-7978: Drop sym_node_channel_ctl before upgrading from pre-3.17 version

pavel-jm:


symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -72,229 +73,47 @@ public class DatabaseUpgradeListener implements IDatabaseUpgradeListener, ISymme
     protected boolean isUpgradeFromPre316;
     protected boolean isUpgradeFromPre317;
 
+    @Override
```
pavel-jm, line 76: Refactor - Just moved existing method to the top

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -72,229 +73,47 @@ public class DatabaseUpgradeListener implements IDatabaseUpgradeListener, ISymme
...
+    }
+
     @Override
     public String beforeUpgrade(ISymmetricDialect symmetricDialect, String tablePrefix, Database currentModel, Database desiredModel)
             throws IOException {
         StringBuilder sb = new StringBuilder();
+        ISqlTemplate sqlTemplate = engine.getSqlTemplate();
         isUpgradeFromPre38 = isUpgradeFromPre38(tablePrefix, currentModel, desiredModel);
         if (isUpgradeFromPre38) {
-            Table transformTable = currentModel.findTable(tablePrefix + "_" + TableConstants.SYM_TRANSFORM_TABLE);
```
pavel-jm, line 81: Refactor - Just moved existing code to new helper method

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -72,229 +73,47 @@ public class DatabaseUpgradeListener implements IDatabaseUpgradeListener, ISymme
...
-                engine.getSqlTemplate().update("delete from " + dataGapTableName);
-            }
-            String nodeCommunicationTable = tablePrefix + "_" + TableConstants.SYM_NODE_COMMUNICATION;
-            if (currentModel.findTable(nodeCommunicationTable) != null) {
-                engine.getSqlTemplate().update("delete from " + tablePrefix + "_" + TableConstants.SYM_NODE_COMMUNICATION);
-            }
+            beforeUpgradeFromPre3_8(tablePrefix, currentModel, sqlTemplate);
         }
         if (isUpgradeFromPre310(tablePrefix, currentModel, desiredModel)) {
-            String name = engine.getDatabasePlatform().getName();
```
pavel-jm, line 96: Refactor - Just moved existing code to new helper method

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -72,229 +73,47 @@ public class DatabaseUpgradeListener implements IDatabaseUpgradeListener, ISymme
...
-                    engine.getSqlTemplate().update("alter table " + tablePrefix + "_" + TableConstants.SYM_NODE_SECURITY
-                            + " drop constraint " + tablePrefix + "_fk_sec_2_node");
-                } catch (Exception e) {
-                    log.info("Unable to drop FK constraint " + tablePrefix + "_fk_sec_2_node to node table", e);
-                }
-            }
+            beforeUpgradeFromPre3_10(tablePrefix, currentModel, sqlTemplate);
         }
-        if (isUpgradeFromPre311(tablePrefix, currentModel, desiredModel) && shouldFixDataEvent311(tablePrefix)) {
-            fixDataEvent311(tablePrefix);
```
pavel-jm, line 114: Refactor - Just moved existing code to new helper method

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -72,229 +73,47 @@ public class DatabaseUpgradeListener implements IDatabaseUpgradeListener, ISymme
...
+        if (isUpgradeFromPre311(tablePrefix, currentModel, desiredModel)) {
+            beforeUpgradeFromPre3_11(tablePrefix);
         }
         if (isUpgradeFromPre312(tablePrefix, currentModel, desiredModel)) {
-            if (engine.getParameterService().isRegistrationServer()) {
```
pavel-jm, line 117: Refactor - Just moved existing code to new helper method

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -72,229 +73,47 @@ public class DatabaseUpgradeListener implements IDatabaseUpgradeListener, ISymme
...
-                    log.info("Unable to drop FK constraint to router table: {}", e.getMessage());
-                }
-            }
+            beforeUpgradeFromPre3_12(tablePrefix, currentModel, sqlTemplate);
         }
         if (isUpgradeFromPre3125(tablePrefix, currentModel, desiredModel)) {
             isUpgradeFromPre3125 = true;
         }
         isUpgradeFromPre314 = isUpgradeFromPre314(tablePrefix, currentModel, desiredModel);
-        if (engine.getDatabasePlatform().getName().equals(DatabaseNamesConstants.INFORMIX)) {
```
pavel-jm, line 162: Refactor - Just moved existing code to new helper method for Informix

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -72,229 +73,47 @@ public class DatabaseUpgradeListener implements IDatabaseUpgradeListener, ISymme
...
-                    }
-                }
-            }
-        }
+        fixInformixTriggerLongVarcharColumns(tablePrefix, desiredModel);
         if (isUpgradeFromPre315(tablePrefix, currentModel)) {
             isUpgradeFromPre315 = true;
         }
         if (isUpgradeFromPre315) {
-            String name = engine.getDatabasePlatform().getName();
```
pavel-jm, line 179: Refactor - Just moved existing code to new helper method

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -72,229 +73,47 @@ public class DatabaseUpgradeListener implements IDatabaseUpgradeListener, ISymme
...
-                    engine.getSqlTemplate().update("truncate table " + tablePrefix + "_" + TableConstants.SYM_REGISTRATION_REQUEST);
-                } catch (Exception e) {
-                    log.info("Unable to truncate registration request table: {}", e.getMessage());
-                }
-            }
+            beforeUpgradeFromPre3_15(tablePrefix, currentModel, sqlTemplate);
         }
         isUpgradeFromPre316 = isUpgradeFromPre316(tablePrefix, currentModel);
         if (isUpgradeFromPre316) {
-            String[] tableNames = { tablePrefix + "_design_diagram", tablePrefix + "_diagram_group" };
```
pavel-jm, line 277: Refactor - Just moved existing code to new helper method

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -72,229 +73,47 @@ public class DatabaseUpgradeListener implements IDatabaseUpgradeListener, ISymme
...
-                    } catch (Exception e) {
-                        log.info("Unable to drop table {} because: {}", tableName, e.getMessage());
-                    }
-                }
-            }
+            beforeUpgradeFromPre3_16(tablePrefix, currentModel, sqlTemplate);
         }
         isUpgradeFromPre317 = isUpgradeFromPre317(tablePrefix, currentModel);
         if (isUpgradeFromPre317) {
-            String nodeChannelControlTableName = TableConstants.getTableName(tablePrefix, TableConstants.SYM_NODE_CHANNEL_CTL);
```
pavel-jm, line 291: Not just a refactor - The new helper method performs a DROP, not DELETE as before

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -304,6 +123,160 @@ public String beforeUpgrade(ISymmetricDialect symmetricDialect, String tablePref
         return sb.toString();
     }
 
+    protected boolean dropTableDueToUpgrade(Table table, Database currentModel, ISqlTemplate sqlTemplate) {
```
pavel-jm: Refactor - Moved (and chopped to smaller bites) existing code to a few new helper methods

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -606,8 +579,98 @@ protected void createH2SequenceIfMissing(ISymmetricDialect symmetricDialect, Str
         }
     }
 
-    @Override
```
pavel-jm, line 609: Refactor - Just moved existing code to the top of class

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -606,8 +579,98 @@ protected void createH2SequenceIfMissing(ISymmetricDialect symmetricDialect, Str
...
+
+    protected void beforeUpgradeFromPre3_16(String tablePrefix, Database currentModel, ISqlTemplate sqlTemplate) {
+        String[] tableNames = { TableConstants.getTableName(tablePrefix, TableConstants.SYM_DESIGN_DIAGRAM),
+                TableConstants.getTableName(tablePrefix, TableConstants.SYM_DIAGRAM_GROUP) };
+        dropTables(tableNames, currentModel, sqlTemplate);
+    }
+
+    protected void beforeUpgradeFromPre3_17(String tablePrefix, Database currentModel, ISqlTemplate sqlTemplate) {
+        String[] tableNames = { TableConstants.getTableName(tablePrefix, TableConstants.SYM_NODE_CHANNEL_CTL) };
+        dropTables(tableNames, currentModel, sqlTemplate);
```
pavel-jm: Actual change is here: DROP instead of DELETE

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -304,6 +123,160 @@ public String beforeUpgrade(ISymmetricDialect symmetricDialect, String tablePref
...
+            String constraintName = sqlTemplate.queryForString(
+                    "select name from sysobjects where xtype = 'F' and parent_obj = object_id('" + table.getName() + "')");
+            return dropConstraintFromTable(table, constraintName, sqlTemplate);
+        } catch (Exception e) {
+            log.warn("Unable to find foreign key constraint for table {} during upgrade process because: {}", table.getName(), e.getMessage());
+        }
+        return false;
+    }
+
+    protected boolean dropFkFromTables(String[] tableNames, Database currentModel, ISqlTemplate sqlTemplate) {
```
w3bbd3v-jm: Is this called by anyone, other than the unit test?

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -304,6 +123,160 @@ public String beforeUpgrade(ISymmetricDialect symmetricDialect, String tablePref
...
+        try {
+            sqlTemplate.update("drop index " + table.getName() + "." + indexName);
+            return true;
+        } catch (Exception e) {
+            log.warn("Unable to drop index {} from table {} during upgrade process because: {}", indexName, table.getName(), e.getMessage());
+        }
+        return false;
+    }
+
+    protected boolean dropForeignKeyConstraintDueToUpgrade(Table table, ISqlTemplate sqlTemplate) {
```
w3bbd3v-jm: Is this called by anyone, other than the unit test and `dropFkFromTables`?

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -72,229 +73,47 @@ public class DatabaseUpgradeListener implements IDatabaseUpgradeListener, ISymme
...
-            }
-            String dataGapTableName = tablePrefix + "_" + TableConstants.SYM_DATA_GAP;
-            if (currentModel.findTable(dataGapTableName) != null) {
-                engine.getSqlTemplate().update("delete from " + dataGapTableName);
-            }
-            String nodeCommunicationTable = tablePrefix + "_" + TableConstants.SYM_NODE_COMMUNICATION;
-            if (currentModel.findTable(nodeCommunicationTable) != null) {
-                engine.getSqlTemplate().update("delete from " + tablePrefix + "_" + TableConstants.SYM_NODE_COMMUNICATION);
-            }
+            beforeUpgradeFromPre3_8(tablePrefix, currentModel, sqlTemplate);
```
w3bbd3v-jm: Is it simpler to just call this `upgradeFromPre3_8`?  Same applies to other new methods with `before` prefix. With "before" I feel like I'm overthinking what it means.

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -304,6 +123,160 @@ public String beforeUpgrade(ISymmetricDialect symmetricDialect, String tablePref
...
+            log.info("Per upgrade process, dropping table: {}", table.getName());
+            sqlTemplate.update("drop table " + table.getName());
+            return true;
+        } catch (Exception e) {
+            log.warn("Unable to drop table {} during upgrade process because: {}", table.getName(), e.getMessage());
+        }
+        return false;
+    }
+
+    protected boolean dropTables(String[] tableNames, Database currentModel, ISqlTemplate sqlTemplate) {
```
w3bbd3v-jm: This returns a boolean type, but the return value is never used. We could make this `void` instead.

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -304,6 +123,160 @@ public String beforeUpgrade(ISymmetricDialect symmetricDialect, String tablePref
...
+            log.info("Per upgrade process, truncating table: {}", table.getName());
+            sqlTemplate.update("truncate table " + table.getName());
+            return true;
+        } catch (Exception e) {
+            log.warn("Unable to truncate table {} during upgrade process because: {}", table.getName(), e.getMessage());
+        }
+        return false;
+    }
+
+    protected boolean truncateTables(String[] tableNames, Database currentModel, ISqlTemplate sqlTemplate) {
```
w3bbd3v-jm: This returns a boolean type, but the return value is never used. We could make this void instead.

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
pavel-jm: Actually, there are two "families" of methods in the IDatabaseUpgradeListener interface: 
* beforeUpdate and 
* afterUpgrade

The beforeUpgradeFromPre3_8 name refers to two things:
1. It's called "before" the actual software upgrade is attempted
2. It only applies to upgrades from versions prior to 3.8

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
pavel-jm: Nice catch! Code was duplicated. Removed!

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
pavel-jm: Nope, there is code duplication. Removed!

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
w3bbd3v-jm: Okay, thanks for clarifying. That makes a bit more sense.

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -304,6 +123,135 @@ public String beforeUpgrade(ISymmetricDialect symmetricDialect, String tablePref
...
+            log.info("Per upgrade process, dropping table: {}", table.getName());
+            sqlTemplate.update("drop table " + table.getName());
+            return true;
+        } catch (Exception e) {
+            log.warn("Unable to drop table {} during upgrade process because: {}", table.getName(), e.getMessage());
+        }
+        return false;
+    }
+
+    protected boolean dropTables(String[] tableNames, Database currentModel, ISqlTemplate sqlTemplate) {
```
w3bbd3v-jm: Could we move the DB operations like `dropTables` to a new class? Something like `DatabaseUpgradeOperations`? Then `dropTableDueToUpgrade` could be internal to the new class. 

Same with `truncateTables`. Then `truncateTableDueToUpgrade` is encapsulated in `DatabaseUpgradeOperations`.

And `dropPkFromTables`. Then `dropPrimaryKeyConstraintDueToUpgrade` is encapsulated in `DatabaseUpgradeOperations`.

And so on...This would shorten the `DatabaseUpgradeListener` class quite a bit and keep the DB exposure more isolated in the new `DatabaseUpgradeOperations` class.

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
pavel-jm: True, added capture of return value - to log results better.

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
pavel-jm: True, added capture of return value - to log results better.

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
pavel-jm: That is an excellent idea for a larger refactoring ticket, but is out of scope for this particular fix.
New ticket = https://jumpmind.atlassian.net/browse/SYM-7979

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -606,8 +571,121 @@ protected void createH2SequenceIfMissing(ISymmetricDialect symmetricDialect, Str
...
+            fixDataEvent311(tablePrefix);
+        }
+        return true;
+    }
+
+    protected boolean beforeUpgradeFromPre3_12(String tablePrefix, Database currentModel, ISqlTemplate sqlTemplate, StringBuilder sqlScript) {
+        boolean success = true;
+        if (engine.getParameterService().isRegistrationServer()) {
+            log.info("Before upgrade, fixing router_type");
+            String sql = "update " + TableConstants.getTableName(tablePrefix, TableConstants.SYM_ROUTER)
```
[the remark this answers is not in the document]
pavel-jm: False positive. Input is a fixed constant in this code

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
pavel-jm, line 294: False - positive: Define a constant instead of duplicating this literal "update " 6 times.
Consolidating also creates risk. 
A new ticket will (potentially) revamp this approach anyway https://jumpmind.atlassian.net/browse/SYM-7979

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -606,8 +571,121 @@ protected void createH2SequenceIfMissing(ISymmetricDialect symmetricDialect, Str
...
+            success &= dropConstraintFromTable(currentModel.findTable(TableConstants.getTableName(tablePrefix, TableConstants.SYM_NODE_IDENTITY)),
+                    tablePrefix + "_fk_ident_2_node", sqlTemplate, sqlScript);
+            success &= dropConstraintFromTable(currentModel.findTable(TableConstants.getTableName(tablePrefix, TableConstants.SYM_NODE_SECURITY)),
+                    tablePrefix + "_fk_sec_2_node", sqlTemplate, sqlScript);
+        }
+        return success;
+    }
+
+    protected boolean beforeUpgradeFromPre3_11(String tablePrefix) {
+        if (shouldFixDataEvent311(tablePrefix)) {
```
ShivamJumpmind: maybe we could address these sonar warnings ... and in the process add consistency between the naming 311 and 3_11

pavel-jm: 311 is harder to read than 3_11.  I'm open to new ideas on this one!

ShivamJumpmind: Would it then be a good idea for consistency to name methods such as `shouldFixDataEvent311` to `shouldFixDataEvent3_11`?

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -606,8 +571,121 @@ protected void createH2SequenceIfMissing(ISymmetricDialect symmetricDialect, Str
...
+        return success;
+    }
+
+    protected boolean beforeUpgradeFromPre3_16(String tablePrefix, Database currentModel, ISqlTemplate sqlTemplate, StringBuilder sqlScript) {
+        String[] tableNames = { TableConstants.getTableName(tablePrefix, TableConstants.SYM_DESIGN_DIAGRAM),
+                TableConstants.getTableName(tablePrefix, TableConstants.SYM_DIAGRAM_GROUP) };
+        return dropTables(tableNames, currentModel, sqlTemplate, sqlScript);
+    }
+
+    protected boolean beforeUpgradeFromPre3_17(String tablePrefix, Database currentModel, ISqlTemplate sqlTemplate, StringBuilder sqlScript) {
```
w3bbd3v-jm, line 690: Can you add unit tests for this? Consider the other methods too since they are now pulled out into individual methods and are easier to test?

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -72,236 +73,200 @@ public class DatabaseUpgradeListener implements IDatabaseUpgradeListener, ISymme
...
+        try {
+            sqlTemplate.update(sql);
+            return true;
+        } catch (Exception e) {
+            log.warn("Unable to drop index {} from table {} during upgrade process because: {}", indexName, table.getName(), e.getMessage());
+        }
+        return false;
+    }
+
+    protected boolean dropConstraintFromTable(Table table, String constraintName, ISqlTemplate sqlTemplate, StringBuilder sqlScript) {
```
w3bbd3v-jm, line 227: What happens if `constraintName` is `null`?

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -72,236 +73,200 @@ public class DatabaseUpgradeListener implements IDatabaseUpgradeListener, ISymme
...
-                } catch (Exception e) {
-                    log.info("Unable to drop FK constraint " + tablePrefix + "_fk_sec_2_node to node table", e);
-                }
-            }
+            success &= beforeUpgradeFromPre3_10(tablePrefix, currentModel, sqlTemplate, sqlScript);
         }
-        if (isUpgradeFromPre311(tablePrefix, currentModel, desiredModel) && shouldFixDataEvent311(tablePrefix)) {
-            fixDataEvent311(tablePrefix);
+        if (isUpgradeFromPre311(tablePrefix, currentModel, desiredModel)) {
+            success &= beforeUpgradeFromPre3_11(tablePrefix);
```
w3bbd3v-jm, line 95: `beforeUpgradeFromPre3_11` always returns true here.

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -72,236 +73,200 @@ public class DatabaseUpgradeListener implements IDatabaseUpgradeListener, ISymme
...
-                        column.setJdbcTypeCode(Types.VARCHAR);
-                        column.setMappedType("VARCHAR");
-                        column.setMappedTypeCode(Types.VARCHAR);
-                        column.setSize("255");
-                    }
-                }
-            }
-        }
+        fixInformixTriggerLongVarcharColumns(tablePrefix, desiredModel);
         if (isUpgradeFromPre315(tablePrefix, currentModel)) {
```
w3bbd3v-jm: This block looks off. Was this intentional?

```
 if (isUpgradeFromPre315(tablePrefix, currentModel)) {
            isUpgradeFromPre315 = true;
 }
if (isUpgradeFromPre315) {
            success &= beforeUpgradeFromPre3_15(tablePrefix, currentModel, sqlTemplate, sqlScript);
 }
```

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
pavel-jm: Thanks! Added return type to inner helper for beforeUpgradeFromPre3_11

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
pavel-jm: Indeed! I've renamed methods, but kept member variables the same - to avoid confusion.

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
pavel-jm: Yep. Added 39 new unit tests to
DatabaseUpgradeListenerTest.java

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
pavel-jm: Thanks. Cannot drop unspecified constraint from table

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -606,8 +572,121 @@ protected void createH2SequenceIfMissing(ISymmetricDialect symmetricDialect, Str
...
+            return fixDataEvent3_11(tablePrefix);
+        }
+        return true;
+    }
+
+    protected boolean beforeUpgradeFromPre3_12(String tablePrefix, Database currentModel, ISqlTemplate sqlTemplate, StringBuilder sqlScript) {
+        boolean success = true;
+        if (engine.getParameterService().isRegistrationServer()) {
+            log.info("Before upgrade, fixing router_type");
+            String sql = "update " + TableConstants.getTableName(tablePrefix, TableConstants.SYM_ROUTER)
```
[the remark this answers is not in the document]
pavel-jm: False positive x2! Input is a fixed constant in this code

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
pavel-jm: Yep. Added

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
pavel-jm: Yes, I've actually renamed methods to no longer match member variables. That was confusing!

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -434,7 +401,7 @@ protected boolean isUpgradeFromPre38(String tablePrefix, Database currentModel,
         }
     }
 
-    protected boolean isUpgradeFromPre310(String tablePrefix, Database currentModel, Database desiredModel) {
+    protected boolean isUpgradeFromPre3_10(String tablePrefix, Database currentModel, Database desiredModel) {
```
w3bbd3v-jm: Sonar's finding is good here I think. It would simplify the method. I think I see 4 instances of this warning.
Remove unused "desiredModel" parameter.

pavel-jm: OK. Removed desiredModel

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -606,8 +574,121 @@ protected void createH2SequenceIfMissing(ISymmetricDialect symmetricDialect, Str
...
+    }
+
+    protected boolean beforeUpgradeFromPre3_16(String tablePrefix, Database currentModel, ISqlTemplate sqlTemplate, StringBuilder sqlScript) {
+        String[] tableNames = { TableConstants.getTableName(tablePrefix, TableConstants.SYM_DESIGN_DIAGRAM),
+                TableConstants.getTableName(tablePrefix, TableConstants.SYM_DIAGRAM_GROUP) };
+        return dropTables(tableNames, currentModel, sqlTemplate, sqlScript);
+    }
+
+    protected boolean beforeUpgradeFromPre3_17(String tablePrefix, Database currentModel, ISqlTemplate sqlTemplate, StringBuilder sqlScript) {
+        String[] tableNames = { TableConstants.getTableName(tablePrefix, TableConstants.SYM_NODE_CHANNEL_CTL) };
```
w3bbd3v-jm, line 691: Is the solution of dropping the node channel control table on upgrade decided on? I thought we didn't want to drop the table entirely?

pavel-jm: Yes, we've settled on that Wednesday. The reason is:
* Existing code already deletes every row from this table.
* Error related to PK issue comes up even for an empty backup table.
 
In the future we'll add more MariaDB-specific tweaks per ticket https://jumpmind.atlassian.net/browse/SYM-8048, but it will require a bit more research.

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
evan-miller-jumpmind, line 1: Should we rename all of the `success` variables in this file to `isSuccess` or `isSuccessful` to align with our usual naming convention?

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -72,236 +73,202 @@ public class DatabaseUpgradeListener implements IDatabaseUpgradeListener, ISymme
...
+        for (String tableName : tableNames) {
+            Table table = currentModel.findTable(tableName);
+            if (table != null) {
+                success &= truncateTableDueToUpgrade(table, sqlTemplate, sqlScript);
+            }
+        }
+        return success;
+    }
+
+    protected boolean dropPrimaryKeyConstraintDueToUpgrade(Table table, ISqlTemplate sqlTemplate, StringBuilder sqlScript) {
```
evan-miller-jumpmind: I think the name of this method should clarify that it is compatible with SQL Server only.

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -72,236 +73,202 @@ public class DatabaseUpgradeListener implements IDatabaseUpgradeListener, ISymme
...
+        try {
+            String constraintName = sqlTemplate.queryForString(sql);
+            return dropConstraintFromTable(table, constraintName, sqlTemplate, sqlScript);
+        } catch (Exception e) {
+            log.warn("Unable to find primary key constraint for table {} during upgrade process because: {}", table.getName(), e.getMessage());
+        }
+        return false;
+    }
+
+    protected boolean dropPkFromTables(String[] tableNames, Database currentModel, ISqlTemplate sqlTemplate, StringBuilder sqlScript) {
```
evan-miller-jumpmind: I think the name of this method should clarify that it is compatible with SQL Server only.

evan-miller-jumpmind (commented):
I left a few comments. Also, there are several methods in `DatabaseUpgradeListener` that write their own SQL to perform DDL operations when it would probably be better for them to delegate to methods such as `IDatabasePlatform.dropTables()` or `ISymmetricDialect.truncateTable()`. However, it seems like refactoring this would be out of scope for this ticket, since the listener was running SQL this way even before your changes.

pavel-jm:
> I left a few comments. Also, there are several methods in `DatabaseUpgradeListener` that write their own SQL to perform DDL operations when it would probably be better for them to delegate to methods such as `IDatabasePlatform.dropTables()` or `ISymmetricDialect.truncateTable()`. However, it seems like refactoring this would be out of scope for this ticket, since the listener was running SQL this way even before your changes.

Great suggestion. I've opened a follow up ticket https://jumpmind.atlassian.net/browse/SYM-8089, to avoid users waiting for the fix longer than necessary.

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
pavel-jm: Great suggestion. I've renamed it

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
pavel-jm: Great suggestion. I've renamed it

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
pavel-jm: Agreed, and renamed.

symmetric-core/src/main/java/org/jumpmind/symmetric/db/DatabaseUpgradeListener.java
```diff
@@ -456,9 +423,9 @@ protected boolean isUpgradeFromPre311(String tablePrefix, Database currentModel,
...
 
-    protected boolean shouldFixDataEvent311(String tablePrefix) {
-        boolean shouldFix = engine.getParameterService().is("upgrade.force.fix.data.event");
-        if (!shouldFix && !engine.getParameterService().is("upgrade.skip.fix.data.event")) {
+    protected boolean shouldFixDataEvent3_11(String tablePrefix) {
+        boolean isFixNeeded = engine.getParameterService().is("upgrade.force.fix.data.event");
+        if (!isFixNeeded && !engine.getParameterService().is("upgrade.skip.fix.data.event")) {
             HashSet<String> set = new HashSet<String>();
             String sql = "select t.trigger_id, r.target_node_group_id from " + tablePrefix + "_trigger t inner join " + tablePrefix +
                     "_trigger_router tr on tr.trigger_id = t.trigger_id inner join " + tablePrefix +
```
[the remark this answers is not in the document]
pavel-jm: Nope, tablePrefix is set by admin once during installation of SymmetricDS (remains static).
The source_node_group_id undergoes a separate identifier check, before being saved.

evan-miller-jumpmind (approved):

PR #1074, opened 2026-09-08
SYM-7981: centralize singleton construction into a single ServiceRegistry.

w3bbd3v-jm:


pavel-jm (approved):

evan-miller-jumpmind (commented):
Should we look at moving the `MetricsManager` and `ModuleManager` instances into `ServiceRegistry` as well, or is that outside of the scope of this ticket?

Also, did you check symmetric-pro for any remaining references to `IStartupParameterService.getInstance()` or `ClusteredCacheManager.getInstance()`?

evan-miller-jumpmind (approved):

PR #1083, opened 2026-09-13
SYM-8083: Thread-safe trigger template hash compute

pavel-jm:
Thread-safe trigger template hash compute to prevent multi-threaded SyncTriggers job from corrupting it, leading to unnecessary rebuilds.

AbstractTriggerTemplate.toHashedValue() accumulated into the shared hashedValue field inside an unsynchronized lazy block. With sync.triggers.thread.count.per.server > 1 the workers of updateOrCreateDatabaseTriggers added their terms into the same field, the cached value differed from run to run and every following sync-triggers rebuilt all triggers with reason TRIGGER_TEMPLATE_CHANGED. The hash is now calculated into a local and published once through a volatile field; the arithmetic is unchanged so stored trigger_template_hash values stay valid.

Adds a 16-thread barrier test that fails against the previous implementation, a test that the hash covers the SQL templates, and moves AbstractTriggerTemplateTest to JUnit Jupiter assertions.

Claude-Session: https://claude.ai/code/session_01XS4UKrACeXbihUGaSYwVXv


symmetric-core/src/main/java/org/jumpmind/symmetric/db/AbstractTriggerTemplate.java
```diff
@@ -92,7 +92,7 @@ abstract public class AbstractTriggerTemplate {
     protected String newColumnPrefix = "";
     protected String otherColumnTemplate;
     protected ISymmetricDialect symmetricDialect;
-    protected int hashedValue = 0;
+    protected volatile int hashedValue = 0;
```
pavel-jm, line 95: Significant change - guard has from being overwritten by other threads

symmetric-core/src/main/java/org/jumpmind/symmetric/db/AbstractTriggerTemplate.java
```diff
@@ -1190,25 +1190,32 @@ public String toString() {
...
-                hashedValue += sqlTemplates.get(key).hashCode();
-            }
-            Field[] fields = getClass().getSuperclass().getDeclaredFields();
-            for (Field field : fields) {
-                field.setAccessible(true);
-                if (field.getType().equals(String.class)) {
-                    try {
-                        String value = (String) field.get(this);
-                        if (value != null) {
-                            hashedValue += value.hashCode();
```
pavel-jm, line 1203: Stop using class member variable hashedValue as accumulator of incremental values!

symmetric-core/src/main/java/org/jumpmind/symmetric/db/AbstractTriggerTemplate.java
```diff
@@ -1190,25 +1190,32 @@ public String toString() {
...
-                        }
-                    } catch (Exception e) {
-                        log.warn("Failed to get hash code for field " + field.getName());
+            hashedValue = calculateHashedValue();
```
pavel-jm, line 1193: One-and-done update to this member

evan-miller-jumpmind (approved):

PR #1085, opened 2026-09-14
SYM-7998: Add Unit Tests for package org.jumpmind.symmetric.load

Sai-Gumpalli:


symmetric-core/src/test/java/org/jumpmind/symmetric/load/JavaDatabaseWriterFilterTest.java
```diff
@@ -0,0 +1,175 @@
...
+        when(extensionService.getCompiledClass(anyString())).thenThrow(new RuntimeException("compile failed"));
+        Set<String> scripts = new HashSet<String>(Arrays.asList("bad script"));
+        assertThrows(SymmetricException.class, () -> filter.executeScripts(context, "key", scripts, true));
+    }
+
+    @Test
+    void testExecuteScripts_swallowsErrorWhenNotFailOnError() throws Exception {
+        when(extensionService.getCompiledClass(anyString())).thenThrow(new RuntimeException("compile failed"));
+        Set<String> scripts = new HashSet<String>(Arrays.asList("bad script"));
+        filter.executeScripts(context, "key", scripts, false);
```
w3bbd3v-jm: On a lot of these, you may be able to use: `assertDoesNotThrow`.

symmetric-core/src/test/java/org/jumpmind/symmetric/load/DefaultDataLoaderFactoryTest.java
```diff
@@ -0,0 +1,138 @@
...
+    void testSetSymmetricEngine_wiresEngineAndParameterService() {
+        DefaultDataLoaderFactory f = new DefaultDataLoaderFactory();
+        f.setSymmetricEngine(engine);
+        DatabaseWriterSettings settings = f.buildDatabaseWriterSettings(null, null, null, null);
+        assertNotNull(settings);
+    }
+    // The Cassandra- and Kafka-platform branches of getDataWriter() are not covered here: CassandraPlatform
+    // can't even be mocked (its bytecode references the Cassandra driver, which isn't on this module's test
+    // classpath) and constructing a real KafkaWriter needs org.apache.avro, which is likewise absent. Exercising
+    // those branches would require adding those platform driver dependencies to symmetric-core's test classpath.
```
evan-miller-jumpmind: Usually we'd add the the necessary dependencies to the test classpath when we run into an issue like this.

symmetric-core/src/test/java/org/jumpmind/symmetric/load/DefaultDataLoaderFactoryTest.java
evan-miller-jumpmind, line 1: Should this test class also be responsible for testing the `DefaultTransformWriterConflictResolver` and `DynamicDefaultDatabaseWriter` anonymous classes in `getDataWriter()`?

evan-miller-jumpmind (changes requested):
I left a couple of comments. Here are some other issues:

1. The test classes are missing the AGPL license header
2. `ConfigurationChangedDatabaseWriterFilter` is in this same package and it could use some more thorough unit tests. Currently, `ConfigurationChangedDatabaseWriterFilterTest` only tests the `afterWrite()` method.
3. There are a lot of Sonar issues that need addressed

w3bbd3v-jm (approved):

evan-miller-jumpmind (approved):

PR #1094, opened 2026-09-14
SYM-8065: Add emptyTargetNodeGroup monitor type constant

pavel-jm:
## Summary

Open-source half of [SYM-8065](https://jumpmind.atlassian.net/browse/SYM-8065). Adds the monitor type name for a new pro monitor that reports `sym_node_group_link` rows whose target node group has no enabled nodes — a misconfiguration known to degrade routing performance.

The monitor implementation itself lives in the pro repo: **jumpmindinc/symmetric-pro#1173**. That PR needs this one first, since it references `MonitorConstants.EMPTY_TARGET_NODE_GROUP`.

## Changes

- `MonitorConstants` — new `EMPTY_TARGET_NODE_GROUP = "emptyTargetNodeGroup"` constant, plus a `3.18.1` entry in `getMonitorTypesByVersion()` so `ConfigurationVersionHelper.isNodeOlderThanMonitor` filters the `sym_monitor` row out of config sync to clients running an older engine that would not recognize the type.
- `monitors.ad` — reference row in the monitor-types table.

## Background

A node group link survives after its target group is emptied or renamed, and the routers configured against it survive too. In one deployment a device group was renamed, the old link and six routers were left behind, and the configuration was harmless for years — until the routing path addressed in SYM-7891 turned every cache miss into a full node reload at roughly 363 ms per data row.

## Testing

Covered by the pro PR's unit tests and lab verification; there is no behavior to test on this side beyond the constant and the version mapping. Config sync of the new `sym_monitor` row to two same-version clients was confirmed in lab83h2 on 3.18.1.RC2, so the version mapping does not wrongly block current nodes.

[SYM-8065]: https://jumpmind.atlassian.net/browse/SYM-8065?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


symmetric-assemble/src/asciidoc/configuration/monitors.ad
```diff
@@ -55,6 +55,8 @@ be compared to a threshold value.
 
 |postgresWal|Size in MB of the backlog of PostgreSQL Write-Ahead Logging (WAL) changes that haven't yet been mined by the PostgreSQL log miner.|
 
+|emptyTargetNodeGroup|Number of node group links whose target node group has no enabled nodes, which is a misconfiguration that degrades routing performance.|
```
evan-miller-jumpmind: This description is inaccurate compared to the current implementation. It doesn't mention that the monitor only checks group links whose source node group contains the current node.

pavel-jm: Thanks! Updated

evan-miller-jumpmind (approved):

PR #1108, opened 2026-09-15
SYM-7985: Add Unit Tests for package org.jumpmind.symmetric.web

ShivamJumpmind:
Please refer to JIRA ticket for details: [Add Unit Tests for package org.jumpmind.symmetric.web](https://jumpmind.atlassian.net/browse/SYM-7985)


symmetric-server/src/test/java/org/jumpmind/symmetric/web/AuthenticationSessionTest.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Test
+    void testEquals_withSameId() {
+        assertEquals(new AuthenticationSession("session-1"), session);
+    }
+
+    @Test
+    void testEquals_withDifferentId() {
+        assertFalse(session.equals(new AuthenticationSession("session-2")));
```
pavel-jm: I hope there is a better way to check session Id? 
 If there is no good helper / accessor method today, that might be a real gap and we should close it?..

symmetric-server/src/test/java/org/jumpmind/symmetric/web/AuthenticationSessionTest.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Test
+    void testEquals_withOtherType() {
+        assertFalse(session.equals("session-1"));
```
pavel-jm: Check if assertNotEquals is a better , plus more readable option?

symmetric-server/src/test/java/org/jumpmind/symmetric/web/HttpMethodFilterTest.java
```diff
@@ -0,0 +1,140 @@
...
+    void setUp() {
+        filter = new HttpMethodFilter();
+        filterConfig = mock(FilterConfig.class);
+        request = mock(HttpServletRequest.class);
+        response = mock(HttpServletResponse.class);
+        filterChain = mock(FilterChain.class);
+    }
+
+    @Test
+    void testDoFilter_allowsMethodWhenNothingIsConfigured() throws IOException, ServletException {
```
pavel-jm, line 59: Not sure there's much value in parameterized tests here? They might obscure test's intent

pavel-jm (changes requested):

symmetric-server/src/test/java/org/jumpmind/symmetric/web/AuthenticationSessionTest.java
ShivamJumpmind: I am not sure I understand the concern here. Could you please elaborate on why this is a sub-par way of checking a session-id?

ShivamJumpmind: It actually seems like there a `getID()` method ... would this be a better test:

@Test
void testEquals_withDifferentId() {
    assertNotEquals("session-2", session.getId());
}

pavel-jm (approved):

PR #1113, opened 2026-09-16
SYM-8065: Clarify emptyTargetNodeGroup monitor description

pavel-jm:
Follow-up to #1094 (merged) and jumpmindinc/symmetric-pro#1173.

Review on the pro PR raised that the monitor does not appear to cover a target node group that becomes empty because its nodes were disabled or unregistered, as opposed to a group that was renamed or emptied outright.

It does. `MonitorTypeEmptyTargetNodeGroup` re-evaluates `nodeService.findEnabledNodesFromNodeGroup(...)` on every run, which resolves to `NodeCache.getNodesByGroup` — built from `getEnabledNodesFromDatabase()`, i.e. `where sync_enabled = 1`. A node that is disabled drops out of that list, and a node that is deleted from `sym_node` never appears in it, so both show up as an empty target group on the next run at no extra query cost.

The description did not say so, which is what made the coverage look narrower than it is. This spells it out, and fixes the word order in the closing clause ("degrade performance the of the Routing job").

No code change.


evan-miller-jumpmind (approved):
Looks good. My comment was referring to the logging that was added to `ConfigurationChangedListener` rather than the monitor itself. It checks for changes to `sym_node_group_link`, but not changes to `sym_node`. It makes sense that the monitor will cover this case that the log message misses.

PR #1125, opened 2026-09-17
SYM-8018: Add Unit Tests for package org.jumpmind.properties

Sai-Gumpalli:


evan-miller-jumpmind:
Can you add some tests to `TypedPropertiesTest`? It only contains 4 tests and it looks like `TypedProperties` is missing some test coverage.

symmetric-util/src/test/java/org/jumpmind/properties/DefaultParameterParserTest.java
evan-miller-jumpmind, line 1: Did you notice that a `DefaultParameterParserTest` already exists in the `symmetric-core` project? I'm thinking we should only have one test file for `DefaultParameterParser`.

symmetric-util/src/test/java/org/jumpmind/properties/EnvironmentSpecificPropertiesTest.java
evan-miller-jumpmind, line 1: Did you notice that an `EnvironmentSpecificPropertiesTest` already exists in the `org.jumpmind.util` package? I'm thinking we should only have one test file for `EnvironmentSpecificProperties`.

w3bbd3v-jm (approved):

symmetric-util/src/test/java/org/jumpmind/properties/TypedPropertiesTest.java
evan-miller-jumpmind, line 1: Can you add the AGPL header to the top of this file?

evan-miller-jumpmind (approved):

PR #1128, opened 2026-09-17
SYM-8032: Added unit tests for org.jumpmind.symmetric.extract package

evan-miller-jumpmind:


evan-miller-jumpmind:
The sonar-scan check is repeatedly failing with "An unexpected error occurred. Please try again later.", but the issues that the initial scan found should be fixed now.

w3bbd3v-jm (approved):

PR #1131, opened 2026-09-21
SYM-8067: Log effective DB connection pool size at startup and warn when concurrent workers exceed it

Sai-Gumpalli:


evan-miller-jumpmind (commented):
The ticket description states that SymmetricDS should log the effective pool size at INFO level during engine startup. Is this missing, or did we decide not to do this?

symmetric-client/src/main/java/org/jumpmind/symmetric/ClientSymmetricEngine.java
```diff
@@ -553,4 +557,13 @@ protected boolean checkImportContainsCurrentGroup(URL fileUrl) {
...
         }
     }
+    
+	private static void logConnectionPoolSize(DataSource dataSource, TypedProperties properties) {
+		int concurrentWorkersMax = properties.getInt(ParameterConstants.CONCURRENT_WORKERS, 20);
+		int effectivePoolSize = IPooledDataSource.of(dataSource).getMaxTotal();
+		log.info("DB connection pool size = {}", effectivePoolSize);
+		if((concurrentWorkersMax * 2) > effectivePoolSize) {
+			log.warn("Configured http.concurrent.workers.max is {}, the effective connection pool size of the database is {}", concurrentWorkersMax, effectivePoolSize);
+		}
```
evan-miller-jumpmind: Can you put the INFO log message in an else block, so it doesn't log redundant information when the WARN log message fires? Also, in the WARN message, can you use `ParameterConstants.CONCURRENT_WORKERS` instead of `"http.concurrent.workers.max"`?

evan-miller-jumpmind (approved):

w3bbd3v-jm (approved):

